PR #1, opened 2020-03-03
Bail if original osc is enabled

Thann:
I think this makes it easier to toggle UOSC on and off. Perhaps for full Interoperability with other OSC's this should also require `uosc=yes` =]


darsain:
Thx!

PR #22, opened 2020-05-17
Change timeline_flash to respect "no-osc" preferences in input.conf

gabriele-tasca:
First of all thank you for the very nice and useful script.

I found the `timeline_flash` option in `uosc.conf` a little limiting compared to mpv's default options.  In `input.conf`, it is possible to add the `no-osd` prefix to each individual seek command to not flash the timeline.

Since the script already looks into `input.conf`, I added a couple lines to detect the command with the **highest** absolute value seek that is marked with `no-osd`. Then I set the timeline flash to activate only for seeks longer than that value.

In general this doesn't fully respect the user's preferences, unless their input.conf looks like this:
```
RIGHT no-osd seek  5
LEFT  no-osd seek -5

CTRL+RIGHT seek 10
CTRL+LEFT seek -10
```
where all the "short" seeks are marked with `no-osd` and the longer ones aren't, and there is no difference between positive and negative seeks. However I'd say this is a fairly normal approach and I'd expect the vast majority of people to be fine with this.

Feel free to change anything you don't like, or even to ignore the request if you find a better solution.

### DIFF
--- a/uosc.lua
+++ b/uosc.lua
@@ -303,6 +303,8 @@ local state = {
 }
 local forced_key_bindings -- defined at the bottom next to events
 
+local no_osd_l_max = 0 -- maximum seek length for which timeline_flash is suppressed
+
 -- HELPERS
 
 function round(number)
@@ -2167,7 +2169,7 @@ elements:add('timeline', Element.new({
 				if position and state.position then
 					local seek_length = math.abs(position - state.position)
 					-- Don't flash on video looping (seek to 0) or tiny seeks (frame-step)
-					if position > 0.5 and seek_length > 0.5 then this:flash() end
+					if position > 0.5 and seek_length > (no_osd_l_max+0.5) then this:flash() end
 				end
 			end)
 		end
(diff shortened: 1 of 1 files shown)


darsain:
Sorry, but there is no way I'd want to include and maintain something this hacky.

Unfortunately there is no API to tap into `no-osd` binding flag, so uosc just can't use it, or try to predict its usage in a reasonable and reliable way.

I do have a solution to this, I just haven't gotten to it yet.

gabriele-tasca:
Alright, no problem.
Another temporary solution while the final one isn't there would be simply to have an option for the minimum seek length for flashing. Consider that if anyone else complains.

darsain:
You can do this now:

```
right        seek  5
left         seek -5
shift+right  seek  10; script-binding uosc/flash-timeline
shift+left   seek -10; script-binding uosc/flash-timeline
```

FichteFoll:
That's not a bad solution, but unfortunately it doesn't cover seeks from external sources such as syncplay. I don't have an easy solution for that, however, and definitely consider this change to be an improvement.

PR #31, opened 2020-06-19
Support for sponsorblock categories

po5:
Chapter naming has changed to accommodate for categories.
This is the bare minimum to support new & old versions if you don't want all the new categories:
`chapter_ranges=^op| op$|opening<968638:0.5>.*, ^ed| ed$|^end|ending$<968638:0.5>.*|{eof}, sponsor start<3535a5:.5>sponsor end, segment start<3535a5:0.5>segment end`


darsain:
Thx. I'll switch it to bare minimum though. Maintaining all segment types would be too much.

PR #35, opened 2020-08-01
Don't add minus sign to end time when using total_time

po5:
With `total_time=yes`, end time is absolute. There shouldn't be a minus sign.


darsain:
Thx! Missed that one.

PR #38, opened 2020-09-29
Added optional features and some fixes

oe-d:
Check commits for details.

### DIFF
--- a/uosc.conf
+++ b/uosc.conf
@@ -1,3 +1,6 @@
+# show play/pause button
+play_pause=no
+
 # timeline size when fully retracted, 0 will hide it completely
 timeline_size_min=2
 # timeline size when fully expanded, in pixels, 0 to disable
--- a/uosc.conf
+++ b/uosc.conf
@@ -55,6 +55,9 @@ top_bar_size_fullscreen=46
 top_bar_controls=yes
 top_bar_title=yes
 
+# UI rendering delay in seconds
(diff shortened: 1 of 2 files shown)


darsain:
Sorry for wasting your time, but I don't think I'll be merging this. At least not all of it.

### Play/Pause button

One of the main points of uosc is to not bloat the UI and codebase with any unnecessary UI elements. Play pause button is the **prime** example of that. For 99% of people, the whole window is a play/pause button. Also having timeline be the whole width of the video is quite important to me, since it enables comfortable seeking of even tiny videos. So if I ever add buttons, they are going to be above the timeline. But play/pause is not going to be one of them.

### timeline_fast_seek and moving render_delay to options

Sure, this seems useful.

### render_fix_osd for windows

Could you describe this bug a bit more, and how does this fix it/work around it? I'm on windows most of the time personally, and never noticed any such issues.

oe-d:
A play button was really the only thing I was missing from your UI, but I can see why you wouldn't want it. It can be disabled though. Not everyone has always access to mouse/keyboard, like when you're remotely controlling your pc, so it is sometimes handy, and I think it's an essential feature in a media player.

On windows, if you pause the player without moving the mouse cursor, sometimes nothing renders until you move the cursor, or there's an osd message. Maybe this occurs only with specific mpv settings / on certain systems?

Anyways, you're free to add what you want from this.

PR #54, opened 2021-06-21
changed chapter_ranges config + sponscrub support

microraptor:
Changed displayed sponsorblock segments to only include the sponsor segments. However, also included option to display all segments that sponsorblock marks in the default colors of the browser plug-in.

Added support for sponscrub markers.

I think it makes more sense to only display the actual sponsor segments by default and then give the user the option to display all other segments, but in different colors.


darsain:
This is a userland config example. I don't want to bloat chapter_ranges, merge PRs, and release new version of uosc every time some new use case for it pops up. That's why it's designed to be a configuration property, and not built into uosc.

So please, put it rather as a snippet into discussions under show and tell category.

PR #61, opened 2021-12-25
Made subtitle loading possible when watching remote videos

allmeta:
Title.

When not watching a local video I still want to load locally downloaded subtitles.

This just makes the load menu default to $HOME when the video path is a protocol.


tomasklaen:
Thx!

PR #66, opened 2022-02-14
Custom Menu Font (Defaults to old behavior if not set)

DrPleaseRespect:
Allows to set a custom font instead of using osd-font

### DIFF
--- a/uosc.conf
+++ b/uosc.conf
@@ -52,6 +52,9 @@ menu_hjkl_navigation=no
 menu_opacity=0.8
 menu_font_scale=1
 
+# Menu font (Uncomment to use)
+#font=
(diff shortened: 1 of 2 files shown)


tomasklaen:
I made uosc to use osd font so it matches mpv UI and is configurable by setting osd font in the main config file. Why would you need the UI (mpv default and uosc) to use 2 different fonts?

I'm trying hard to limit the settings to a necessary bare minimum so it's not too overwhelming for people configuring uosc, and this seems like bloat. At the very least, it looks like something that will be used by 1 person.

Thank you for contribution, but I don't see merging this without at least a good potential use case that might be useful for more people.

PR #68, opened 2022-03-20
Fix delete-file-next; add flash-top-bar command; allow '#menu:' as keyword

microraptor:
Thanks for this awesome mpv script. Here are some very small improvement suggestions:

[keep playlist on delete-file-next command](https://github.com/darsain/uosc/commit/0005120302085f518e11d14e68d06e6f0d038e38):
This is actually a bug fix. Before when using the delete-file-next binding in a playlist, it would just skip to the next file in the folder and disregard the playlist. Now it moves to the next file in the playlist as it is described in the readme.

[add flash-top-bar command](https://github.com/darsain/uosc/commit/f237f70bcefa8bf7251f2202b8947703596e4092):
Adds another flash binding, just like the other ones. I found it very helpful to flash the top bar when using a "next" shortcut, to see the file title.

[Also allow '#menu:' as keyword instead of '#!' in input.conf](https://github.com/darsain/uosc/commit/5f80046b393222fb6321b4b087fd93749ccdd3c8):
This adds only one if statement and is useful for more compatibility with mpv.NET, which uses the input.conf to build a context menu just like uosc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,7 +379,7 @@ Adding items to menu is facilitated by commenting your keybinds in `input.conf` 
 
 Comment has to be at the end of the line with the binding.
 
-Comment has to start with `#!`.
+Comment has to start with `#!` (or `#menu:`).
 
 Text after `#!` is an item title.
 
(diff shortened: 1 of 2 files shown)


tomasklaen:
All seem reasonable. Thx!

PR #87, opened 2022-07-10
Consistant scrolling direction

christoph-heinrich:
Scrolling up to change a value is semantically associated with
increasing that value.
Volume already worked this way, and the same is true for seeking
by scrolling outside of the UI, but scrolling up on the timeline
seeked backwards and scrolling up on the speed widget reduced speed.
Now scrolling on the timeline or outside of the UI seeks in the same
direction and speed also increases when scrolling up on the widget.
The restriction of only allowing positive numbers for `timeline_step`
has also been lifted, allowing users to set negative numbers
if they want to reverse the scrolling direction of timeline seeking.
This was already possible for speed and volume.

### DIFF
--- a/uosc.lua
+++ b/uosc.lua
@@ -2387,10 +2387,10 @@ elements:add('timeline', Element.new({
 		if this.pressed then this:set_from_cursor() end
 	end,
 	on_wheel_up = function(this)
-		if options.timeline_step > 0 then mp.commandv('seek', -options.timeline_step) end
+		mp.commandv('seek', -options.timeline_step)
 	end,
 	on_wheel_down = function(this)
-		if options.timeline_step > 0 then mp.commandv('seek', options.timeline_step) end
+		mp.commandv('seek', options.timeline_step)
 	end,
 	render = render_timeline,
 }))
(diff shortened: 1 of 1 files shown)


tomasklaen:
If you're scrolling horizontally with a wheel, scrolling down to scroll right is a convention, that's why timeline and speed go that way. You can see that all over the place, for example firefox tabs, or any horizontally scrollable area in any browser when you hold shift and scroll with a wheel.

So I'm not gonna change the default behavior, but making it configurable with negative numbers is a good addition. I'll look it through when I have some time.

christoph-heinrich:
All other media players that I'm aware of seek forward when scrolling up, so you're going against an already established convention for media players.
I've separated it out into two commits, so that you can merge the first one on it's own if you really don't want to change scrolling direction.

FichteFoll:
For the record, the [default mpv bindings](https://github.com/mpv-player/mpv/blob/284fecc0bdd8152a35a35a387faf5d4675b8f0c8/etc/input.conf#L40) do this as well. Personally, I don't use the scroll wheel for seeking, so I don't have a personal preference here. I couldn't even tell you what I find more natural.

PR #88, opened 2022-07-10
Fix initial static pause indicator

christoph-heinrich:
When starting the player paused, the static pause indicator was not shown.
By ignoring the initial invocation only for the flash indicator, this is now fixed.

### DIFF
--- a/uosc.lua
+++ b/uosc.lua
@@ -329,7 +329,7 @@ local state = {
 	media_title = '',
 	duration = nil,
 	position = nil,
-	pause = false,
+	pause = mp.get_property_native('pause'),
 	chapters = nil,
 	chapter_ranges = nil,
 	border = mp.get_property_native('border'),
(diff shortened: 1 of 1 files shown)


christoph-heinrich:
This has turned into more then just fixing the initial static pause indicator.

I've been hunting down a bug that seemed like a race condition, which caused the pause indicator to sometimes flash in the beginning when opened with `--pause`. Throwing the timing slightly off by e.g. having something running in the background would prevent it from happening. Whenever it happened the observe callback would be called with `true` for the initial call, then `false`, then `true` again.

Turns out it only happens when I have [reload.lua](https://github.com/4e6/mpv-reload/blob/master/reload.lua) in my script directory. I haven't looked what about that script could possibly be causing that.

While figuring that out I have ended up with some code for the pause indicator that I prefer over my original commit.

PR #91, opened 2022-07-12
Add UI scaling

### LINKED ISSUE #81: UI elements are too small on high resolution displays
I am currently on a 14" 2880x1800 laptop display with 200% scaling enabled in Windows settings.

Without using uosc:
image

Using uosc (empty `uosc.conf`):
image
mpv-220626-1715-2hNyajMDsT

The default mpv OSD seems to respect the Windows scale factor, but uosc does not. I'm aware that there are granular size configurations, but it would be a lot of work just to get each UI element working properly.
Is it possible to configure uosc to follow the Windows UI scale factor? If not, is it possible to configure the scale factor of all UI elements at once (maybe `scale=1`)?

christoph-heinrich:
Uses a new option `ui_scale` for scaling the UI as well as the `display-hidpi-scale` property.
Has been inspired by the scaling in [console.lua](https://github.com/mpv-player/mpv/blob/0044c19f0de29ff752eb708365322b0ec16b53e7/player/lua/console.lua#L132-L138). 

Console observes the `display-hidpi-scale` property and renders whenever it changes, however this results in an error in line 1453 `attempt to index upvalue 'forced_key_bindings' (a nil value)`. That's why I commented it out. It's already way too late, so I'm not gonna look into that now. If you know how to fix that, that would be great.

Fixes #81

### DIFF
--- a/uosc.conf
+++ b/uosc.conf
@@ -74,6 +74,8 @@ top_bar_title=yes
 window_border_size=1
 window_border_opacity=0.8
 
+# scale the interface by this factor
+ui_scale=1
 # pause video on clicks shorter than this number of milliseconds, 0 to disable
 pause_on_click_shorter_than=0
 # flash duration in milliseconds used by `flash-{element}` commands
--- a/uosc.lua
+++ b/uosc.lua
@@ -95,6 +95,8 @@ top_bar_title=yes
 window_border_size=1
 window_border_opacity=0.8
 
+# scale the interface by this factor
+ui_scale=1
 # pause video on clicks shorter than this number of milliseconds, 0 to disable
 pause_on_click_shorter_than=0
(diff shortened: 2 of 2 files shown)


tomasklaen:
I'll look into it, and your other PRs, when I have some time. No ETA, a lot on my plate now.

But thx for finding out about the `display-hidpi-scale` prop. I've missed it when I was searching docs for it.

christoph-heinrich:
Turns out the problem was that I accidentally added `()` behind the function in the hook.

christoph-heinrich:
Forgot adding the new option to the default options at the top in `uosc.lua`

tomasklaen:
Sorry for ignoring all the PRs. Don't have much time 😞. Should get to it sometime next week.

PR #92, opened 2022-07-12
Automatically choose rendering frequency

christoph-heinrich:
I think there is slightly less jitter with `display-fps` then with `estimated-display-fps`, although it's really hard to tell and I don't have multiple monitors with different refresh rates to test with.
Simply swap them around if you want to try it yourself.

I've also noticed that `estimated-display-fps` does not get updated when the video is paused, so when there was a low estimated fps because of e.g. heavy background activity, then the refresh rate during pause will also be low. This could be worked around by adding `state.pause and` in front of it.

### DIFF
--- a/uosc.lua
+++ b/uosc.lua
@@ -302,7 +302,9 @@ local options = {
 }
 opt.read_options(options, 'uosc')
 local config = {
-	render_delay = 0.03, -- sets max rendering frequency
+	-- sets max rendering frequency in case the
+	-- native rendering frequency could not be detected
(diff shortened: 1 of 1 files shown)


tomasklaen:
You're dividing `1` with `config.render_delay`, which is itself divided by 1.

Also, is it wise to call `mp.get_X` every time `request_render()` is called? Isn't it expensive? I'd rather it be kept in state and updated in a `mp.observe_property` hook. There's already a lot of prop updating hooks at the bottom above keybindings.

PR #93, opened 2022-07-12
Add chapter title to hovered time

### LINKED ISSUE #85: Show chapter name together with timestamp when hovering over seek bar
Especially when watching videos from youtube, chapters can be quite useful for determining where in the video you're currently hovering at as an alternative to the chapter list you can already open explicitly. A very similar feature is provided by youtube itself, e.g. on https://www.youtube.com/watch?v=XMn0OM0eWDU

PS: searching for "chapter" in this repo's issue almost exclusively yields issues from myself, huh.

christoph-heinrich:
Closes #85

### DIFF
--- a/uosc.lua
+++ b/uosc.lua
@@ -562,34 +562,115 @@ end
 
 function text_width_estimate(text, font_size)
 	if not text or text == "" then return 0 end
-
 	local text_width = 0
-	local char_width = font_size
-	local byte_len = #text
-	local byte_end = 0
+	for _, _, width in utf8_iter(text) do
+		text_width = text_width + width
+	end
+	return text_width * font_size * options.font_height_to_letter_width_ratio
+end
 
-	for byte_start = 1, byte_len do
-		if byte_start > byte_end then
-			local char_width = char_width
-			local char_byte = string.byte(text, byte_start)
+function utf8_iter(string)
+	local byte_start = 1
+	local byte_count = 1
 
-			local byte_count = 1;
-			if     char_byte < 192 then byte_count = 1
-			elseif char_byte < 224 then byte_count = 2
-			elseif char_byte < 240 then byte_count = 3
-			elseif char_byte < 248 then byte_count = 4
-			elseif char_byte < 252 then byte_count = 5
-			elseif char_byte < 254 then byte_count = 6
+	return function()
+		if #string < byte_start then return nil end
+
+		local char_byte = string.byte(string, byte_start)
+
+		byte_count = 1;
+		if     char_byte < 192 then byte_count = 1
+		elseif char_byte < 224 then byte_count = 2
+		elseif char_byte < 240 then byte_count = 3
+		elseif char_byte < 248 then byte_count = 4
+		elseif char_byte < 252 then byte_count = 5
+		elseif char_byte < 254 then byte_count = 6
+		end
+
+		local start = byte_start
+		byte_start = byte_start + byte_count
+
+		return start, byte_count, (byte_count > 2 and 2 or 1)
+	end
+end
+
+function wrap_text(text, line_width_requested)
+	local line_width = 0
+	local wrap_at_chars = {' ', '　', '-', '–'}
+	local remove_when_wrap = {' ', '　'}
+	local lines = {}
+	local line_start = 1
+	local before_end = nil
+	local before_width = 0
+	local before_line_start = 0
+	local before_removed_width = 0
+	local max_width = 0
+	for char_start, count, char_width in utf8_iter(text) do
+		local char_end = char_start + count - 1
+		local char = text.sub(text, char_start, char_end)
+		local can_wrap = false
+		for _, c in ipairs(wrap_at_chars) do
+			if char == c then
+				can_wrap = true
+				break
 			end
-
-			if byte_count ~= 1 then char_width = char_width * 2 end
-
-			text_width = text_width + char_width
-			byte_end = byte_start + byte_count - 1
+		end
+		line_width = line_width + char_width
+		if can_wrap or (char_end == #text) then
+			local remove = false
+			for _, c in ipairs(remove_when_wrap) do
+				if char == c then
+					remove = true
+					break
+				end
+			end
+			local line_width_after_remove = line_width - (remove and char_width or 0)
+			if line_width_after_remove < line_width_requested then
+				before_end = remove and char_start - 1 or char_end
+				before_width = line_width_after_remove
+				before_line_start = char_end + 1
+				before_removed_width = remove and char_width or 0
+			else
+				if (line_width_requested - before_width) <
+				   (line_width_after_remove - line_width_requested) then
+					lines[#lines+1] = text.sub(text, line_start, before_end)
+					line_start = before_line_start
+					line_width = line_width - before_width - before_removed_width
+					if before_width > max_width then max_width = before_width end
+				else
+					lines[#lines+1] = text.sub(text, line_start, remove and char_start - 1 or char_end)
+					line_start = char_end + 1
+					line_width = remove and line_width - char_width or line_width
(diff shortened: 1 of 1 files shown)


christoph-heinrich:
Chapter ranges now work too. Took me a while to wrap my head around those.

christoph-heinrich:
At first I wanted to make the change as unintrusive as possible to how to original code worked, but I think the second commit is much better. If you agree I'll squash it.

uosc.lua
```diff
@@ -1674,15 +1676,28 @@ function render_timeline(this)
...
+		local text_len = math.max(time_formatted:len(), chapter_title:len())
+		local box_half_width_estimate = text_width_estimate(text_len, this.font_size) / 2
 		ass:new_event()
 		ass:append('{\\blur0\\bord1\\shad0\\1c&H'..options.color_background_text..'\\3c&H'..options.color_background..'\\fn'..config.font..'\\fs'..this.font_size..bold_tag..'')
 		ass:append(ass_opacity(math.min(options.timeline_opacity + 0.1, 1)))
-		ass:pos(math.min(math.max(cursor.x, box_half_width_guesstimate), display.width - box_half_width_guesstimate), fay)
+		ass:pos(math.min(math.max(cursor.x, box_half_width_estimate), display.width - box_half_width_estimate), fay)
 		ass:an(2)
-		ass:append(mp.format_time(hovered_seconds))
+		ass:append(chapter_title .. '\\N')
```
FichteFoll, line 1699: I wonder if it would be better to show them on separate ass lines so that they could be left- and right-margined independently, i.e. the timestamp would be closer to the hovered area for long titles.

Illustrations, where `o` represents the currently hovered position, i.e. `cursor.x`:

*current code*
```
this is a very long chapter title
            00:00:00
|o---------------|-----------------|------|

          this is a very long chapter title
                      00:00:00
|----------------|-----------------|-----o|
```

*my suggestion*
```
this is a very long chapter title
00:00:00
|o---------------|-----------------|------|

     this is a very long chapter title
                 00:00:00
|o---------------|---o-------------|------|

          this is a very long chapter title
                                   00:00:00
|o---------------|-----------------|-----o|
```

christoph-heinrich: Good idea.

Natural-Harmonia-Gropius:
Misaligned with CJK characters.

image
image

christoph-heinrich:
That's unavoidable with the way character width is estimated, happens with latin characters too.

christoph-heinrich:
Although I do wonder... since it is possible to use [this](https://mpv.io/manual/stable/#command-interface-compute-bounds) to get the exact bounds, maybe we could call that once for each string and store the result for future use. This wouldn't just be useful for hovered time and chapters, but also the width estimation in menus and I believe the title at the top also uses width estimation, so a more general solution that could be used in all those places would be useful.

I'll look into it once all my PRs that touch anything to do with width estimation have been merged, otherwise it will just lead to a bunch of conflicts.

Natural-Harmonia-Gropius:
> Although I do wonder... since it is possible to use [this](https://mpv.io/manual/stable/#command-interface-compute-bounds) to get the exact bounds, maybe we could call that once for each string and store the result for future use. This wouldn't just be useful for hovered time and chapters, but also the width estimation in menus and I believe the title at the top also uses width estimation, so a more general solution that could be used in all those places would be useful.
> 
> I'll look into it once all my PRs that touch anything to do with width estimation have been merged, otherwise it will just lead to a bunch of conflicts.

That's sound diffcult. maybe better to make things simple, for now.
https://github.com/Natural-Harmonia-Gropius/uosc/commit/eb26f9786cc287064fbcad94f3653b3d0b69aeeb works for me. byte ranges got from google, coverd most cases.
image
image

christoph-heinrich:
The line ```local char = string.sub(text, i, byte_end)``` is left over form debugging I assume?
Otherwise LGTM, please create a PR.

Natural-Harmonia-Gropius:
> Otherwise LGTM, please create a PR.

https://github.com/christoph-heinrich/uosc/pull/1

> Would you mind if I add that code to [console.lua](https://github.com/mpv-player/mpv/blob/master/player/lua/console.lua) (and maybe other places in mpv) at some point?

Sure, no problem.

christoph-heinrich:
> [christoph-heinrich#1](https://github.com/christoph-heinrich/uosc/pull/1)

I meant a PR for this repo, because it's not specific to this change (although it does currently depend on it, so @darsain needs to be careful when merging).

tomasklaen:
How about do this purely in ass tags, completely agnostic to the width of the text?

Just do a bottom center alignment with `\an2` so that it grows upwards from the center, smart wrapping (each line approximately equally long) with `\q0`, and then limit the position to be at least `font_size * 10` from the edge.

There is no reason we should render chapter titles on a single line.

And if there is a chapter title rendered, the time tooltip should also inherit the same edge limits, so that it is directly below the title.

christoph-heinrich:
I don't know how I can get it to wrap. I've added the `\q0` but it still went outside the screen instead of wrapping. Adding clipping also didn't help. Sorry but I'm not all that good with ass tags :upside_down_face:.

tomasklaen:
Damn. Was hoping `\q0` would wrap when it touches the edges. I guess too much to ask for.

So now we need to wrap ourselves. Best place for that to happen is during chapters serialization. Add an additional `title_wrapped` and `title_wrapped_length` properties, where the length one is the length of the longest line in the wrapped title.

Wrapping should happen at every space closest to the 25th character in a current line.

Then in rendering we use `title_wrapped_length` to decide the center point, and use that for both chapter title, and hovered time so they stick together.

Though I'm not 100% sure how to do wrapping for asian text without breaking it 😶.

I'm going to merge the improved text width estimation PR so you can use it here.

FichteFoll:
`\q0` only works for automatically positioned subtitles, i.e. those without an explicit `\pos`.

christoph-heinrich:
I've made an iterator out of the code from @Natural-Harmonia-Gropius and used that for the text wrapping thing.
The code of the text wrapping function is kinda horrible, but at this point I'm just glad that it works reliably.
Maybe I'll try making it better tomorrow. Any suggestions are welcome.
text_wrap_example
text_wrap_example_jap
text_wrap_example_rus

I made the byte count for double width > 2 , because otherwise the russian text would look like this:
text_wrap_example_rus_1

tomasklaen:
Looks good. The wrapping function does look a bit over-complicated, but I'm too lazy to come up with something myself :)

The only nitpick I have, I'd like an extra space between hovered time and chapter title, otherwise it looks like it's just another line of the title.

Natural-Harmonia-Gropius:
I prefer to render time and chapters separately，The time jumpping when width of chapter in timeline smaller than hover warpper.

tomasklaen:
Hm... that's true, there will be an annoying jumping for chapters at the edges. Makes sense to decouple them then.

christoph-heinrich:
Time is once again separate from the chapter title and there is some extra spacing between them.

tomasklaen:
Tried it out, and apart of the two nitpicks below it works great!

1\. It looks better when chapter title is forced bold, so just replace the optional `..bold_tag..` on it with `\\b1`.

2\. I found a title that wraps correctly, but the shorter line is returned as text width and used to calculate limits:

```
Abaurin's punishment and pound of flesh
```

... or at least that's what I'm guessing is going on, since the upper line `Abaurin's punishment and` clips outside of the screen (on the right side), while the lower line `pound of flesh` is correctly limited and stops moving when it touches the edge.

christoph-heinrich:
You're right, that string doesn't behave correctly and I already found the mistake. Was a pretty obvious one tbh, I'm surprised it never came up during testing.

tomasklaen:
👍

FichteFoll:
Thanks for the change, very cool.

PR #96, opened 2022-07-19
Remove 100ms leeway in chapter menu

christoph-heinrich:
Due to the 100 leeway, a chapter starting at 00:00 would not be active,
when the playback time also was 00:00.


tomasklaen:
Hm... I can't recall why I put that leeway there. Maybe it was supposed to be adding instead of subtracting. Either way, was wrong, thx for PR.

PR #99, opened 2022-07-20
Show more information for tracks

### LINKED ISSUE #58: Ambiguous tracks in audio track selection
Different audio tracks can become indistinguishable in files with multiple audio tracks of the same language. Often, those tracks contain different formats, channel layouts or sample rates which could be displayed in addition to the language of the track.

An example of this being implemented in the track selection in mpv.net:
example

This would make track selection much easier (in certain cases).

christoph-heinrich:
Closes #58

### DIFF
--- a/uosc.lua
+++ b/uosc.lua
@@ -2985,9 +2985,27 @@ function create_select_tracklist_type_menu_opener(menu_title, track_type, track_
 			if track.type == track_type then
 				if track.selected then active_item = track.id end
 
+				local hint_vals = {
+					track.lang and track.lang:upper() or nil,
+					track['demux-h'] and (track['demux-w'] and track['demux-w'] .. 'x' .. track['demux-h']
+					                      or track['demux-h'] .. 'p'),
+					track['demux-fps'] and string.format('%.5gfps', track['demux-fps']) or nil,
+					track.codec,
+					track['audio-channels'] and track['audio-channels'] .. ' channels' or nil,
+					track['demux-samplerate'] and string.format('%.3gkHz', track['demux-samplerate']/1000) or nil,
+					track.forced and 'forced' or nil,
+					track.default and 'default' or nil,
+				}
+				local hint_vals_filtered = {}
+				for i = 1, #hint_vals do
+					if hint_vals[i] then
+						hint_vals_filtered[#hint_vals_filtered+1] = hint_vals[i]
+					end
+				end
+
 				items[#items + 1] = {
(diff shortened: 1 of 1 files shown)


uosc.lua
```diff
@@ -2985,9 +2985,27 @@ function create_select_tracklist_type_menu_opener(menu_title, track_type, track_
...
 				if track.selected then active_item = track.id end
 
+				local hint_vals = {
+					track.lang and track.lang:upper() or nil,
+					track['demux-h'] and (track['demux-w'] and track['demux-w'] .. 'x' .. track['demux-h']
+					                      or track['demux-h'] .. 'p'),
+					track['demux-fps'] and string.format('%.5gfps', track['demux-fps']) or nil,
+					track.codec,
+					track['audio-channels'] and track['audio-channels'] .. ' channels' or nil,
+					track['demux-samplerate'] and string.format('%.3gkHz', track['demux-samplerate']/1000) or nil,
```
FichteFoll, line 2995: I was just wondering if a utility function streamline this kind of "process if not nil" behavior would be helpful.

My Lua is a bit rusty, but something like this:

```lua
-- function first so that it can handle varargs
function ifnotnil(x, ...)
    local args = {...}
    for i = 1, #args do
        if not hint_vals[i] then
            return nil
        end
    end
    if (type(x) == 'function') then
        return x(...)
    else
        return x
    end
end

hint_vals = {
    ifnotnil(string.upper, track.lang),
    ifnotnil(function(w, h) return string.format('%sx%s', w, h); end, track['demux-w'], track['demux-h'])
        or ifnotnil(function(h) return h .. 'p' end, track['demux-h']),
    ifnotnil(function(fps) return string.format('%.5gfps', fps); end, track['demux-fps']),
    track.codec,
    ifnotnil(function(c) return c .. ' channels'; end, track['audio-channels']),
    ifnotnil(function(rate) return string.format('%.3gkHz', rate/1000); end, track['demux-samplerate']),
    ifnotnil('forced', track.forced),
    ifnotnil('Default', track.default),
}
```

FichteFoll: Yeah it' unfortunate that Lua doesn't have a more concise lambda syntax (and Moonscript has a whole other bag of issues).

christoph-heinrich:
I just came across this and wanted to leave it as a reminder in case we want to add more info at some point https://github.com/mpv-player/mpv/blob/48eb77207bff01ae766a9fb33b4e4c35cfad62c9/player/command.c#L1958-L2013

PR #100, opened 2022-07-20
Fix speed dial crash

### LINKED ISSUE #78: change the speed via the widget may make uosc crash
Hello, here is the another weird bug.
1. add the speed by the default shortcut
2. press the speed button and quickly move 
3. Sometimes it would crash the uosc (even mpv)

Snipaste_2022-05-27_20-25-32

christoph-heinrich:
When dragging the speed dial and letting go over the timeline,
`this:fadeout` was called, but that function does not exist.
Introduced in 755194352f6eacb3488fd519be6ecbc21137463d

I think this was simply forgotten in the aforementioned commit.

Closes #78

### DIFF
--- a/uosc.lua
+++ b/uosc.lua
@@ -2649,9 +2649,6 @@ if options.speed then
 			end
 		end,
 		on_global_mbtn_left_up = function(this)
-			if this.dragging and elements.timeline.proximity_raw == 0 then
-				this:fadeout()
-			end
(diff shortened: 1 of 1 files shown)


po5:
Thank tou

po5:
Thank you. Keep encountering this but like Fitche I never bothered fixing it.

PR #102, opened 2022-07-20
WIP: Integration with quality-menu

christoph-heinrich:
This is more of a proof of concept then something ready to merge.
It uses the changes from the [script_interface](https://github.com/christoph-heinrich/mpv-quality-menu/tree/script_interface) branch to let users select video and audio formats fetched and formatted by quality-menu.
The formatting assumes a monospace font to display things in a table, but it doesn't look to bad this way either.

Would you be open to something like this in general?

### DIFF
--- a/uosc.lua
+++ b/uosc.lua
@@ -3908,6 +3908,65 @@ mp.add_key_binding(nil, 'stream-quality', function()
 		active_index = active_index,
 	})
 end)
+
+function create_stream_quality_menu(menu_title, format_type)
+	return function()
+		local menu_type = format_type .. '_formats'
+		if menu:is_open(menu_type) then menu:close() return end
+
+		local path = ''
+
+		local function options_callback(url, foptions, format_id)
+			if url == path and menu:is_open(menu_type) then
+				foptions = utils.parse_json(foptions)
+				local items = {}
+				local active_index = nil
+				for i, option in ipairs(foptions) do
+					items[i] = {
+						title = option.label,
+						value = {format_id = option.format, url = url},
+					}
+					if option.format == format_id then active_index = i end
+				end
+				elements.menu:update({
+					items = items,
+					active_index = active_index
+				})
+			end
+		end
+
+		local function selection_handler(format_value)
+			mp.commandv('script-message-to', 'quality_menu', format_type .. '_format_set',
+			            format_value.url, format_value.format_id)
+		end
+
+		local function path_change_handler(_, url)
+			elements.menu:update({items = {}, active_index = nil})
+			path = url
+			mp.commandv('script-message-to', 'quality_menu', format_type .. '_formats_get', url, mp.get_script_name())
+		end
+
+		-- Items and active_index are set in the handle_prop_change callback, since adding
+		-- a property observer triggers its handler immediately, we just let that initialize the items.
+		menu:open({}, selection_handler, {
+			type=menu_type,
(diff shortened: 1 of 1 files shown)


christoph-heinrich:
I've just tried displaying the format information as hints, and while it can look nice, it certainly is less readable.
The usability also goes down dramatically with narrow videos (e.g. YouTube shorts).
formats_hints
formats_labels

tomasklaen:
Hm... I'd rather it be the other way around. Rather than uosc maintaining menus for any script out there, I think it'd be better that uosc provides an API for scripts to send a serialized menu that uosc should display.

Something like (pseudo types):

```
Menu {
  title: string | nil;
  items: Item[];
}

Item {
  title: string | nil;
  hint: string | nil;
  action: Action;
}

Action = ActionSubmenu | ActionCommand | ActionCommandv;

ActionSubmenu extends Menu {
  type: 'submenu';
}

ActionCommand {
  type: 'command';
  command: string;
}

ActionCommandv {
  type: 'commandv';
  params: string[];
}
```

christoph-heinrich:
Currently the title gets clipped in favor of the hint. But I think in some cases (like the track menu in the screenshot) this is actually preferable.
track_menu

I implemented it this way around because I expected it to be much simpler code on both sides, but I'll add a commit to this and the corresponding branch over on quality-menu that reverses the relationship. Adding a commit instead of overwriting to make comparing easier. Not sure when I can get to it though and I can't do any development from this Saturday until the next one, so it might take a while.

Btw I've been using this ever since I wrote it, and having the quality-menu integrated in the UI is fantastic.

tomasklaen:
Have you tried the uosc's built in `stream-quality` menu? I always felt like it's good enough. I just select the quality I want and it selects something nearest to it available.

christoph-heinrich:
quality-menu has the same functionality built in, might have even been the default back when i forked it, but I never liked it.
I want to select an actually available format and not just select something preconfigured and hope for the best.

tomasklaen:
So I've added uosc detection API:

```lua
-- Register response handler
mp.register_script_message('uosc-version', function(version)
	print('uosc version', version)
end)

-- Ask for version
mp.commandv('script-message-to', 'uosc', 'get-version', mp.get_script_name())
```

And menu API:

```lua
local utils = require('mp.utils')
local menu = {
    title = 'Custom menu',
    items = {
        {title = 'Foo', hint = 'foo', value = 'string command args'},
        {title = 'Bar', hint = 'bar', value = {'table', 'command', 'args'}},
        {title = 'Submenu', items = {...}},
    }
}
local json = utils.format_json(menu)
mp.commandv('script-message-to', 'uosc', 'show-menu', json)
```

That should be all that you need.

christoph-heinrich:
I've added uosc support via the new interface in the [script_interface](https://github.com/christoph-heinrich/mpv-quality-menu/tree/script_interface) branch, in case you want to try that out.

tomasklaen:
Yes, that's what the `type` prop is for (documented in readme). If you request opening a menu with a type, and menu of that type is already open, it'll not re-open the old one, but close the current one. So just pass down the menu type and you get toggling for free.

christoph-heinrich:
I've released a new version of [quality-menu](https://github.com/christoph-heinrich/mpv-quality-menu/releases) with uosc integration. This is now completely obsolete.

PR #104, opened 2022-07-22
Update and render on remaining time changes

christoph-heinrich:
Pausing a live stream leads to the remaining time changing without any actual playback, but so far the UI would only update the remaining time when the playback time changed.


FichteFoll:
Thanks for these various usability fixes. Some of these like this have also bothered me for a while, but not enough to spent the time working myself into the code base to fix them (or even reporting them, now that I think of it).

FichteFoll (approved):

PR #109, opened 2022-08-09
Show volume-bar only when audio is available

Natural-Harmonia-Gropius:
Shows volume bar when blank or view a still image feels weird and noisy, so hide it.


tomasklaen:
Looks good, but sort out the inconsistent indentation :) Especially at the bottom you've re-indented some lines that weren't changed.

Natural-Harmonia-Gropius:
> Especially at the bottom you've re-indented some lines that weren't changed.

addressed

PR #110, opened 2022-08-10
Better text_width_estimate

Natural-Harmonia-Gropius:
rebase needed after https://github.com/darsain/uosc/pull/93 been merged.
Edit : 😓 Sorry for those messes of force push.

before:
image
image

after:
image
image

[Advance measure](https://developer.mozilla.org/en-US/docs/Glossary/advance_measure) is a better way to do this. But for now, this is good enough.

### DIFF
--- a/uosc.conf
+++ b/uosc.conf
@@ -42,6 +42,7 @@ speed_size_fullscreen=68
 speed_persistency=
 speed_opacity=1
 speed_step=0.1
+speed_step_is_factor=no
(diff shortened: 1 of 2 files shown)


christoph-heinrich (approved):

tomasklaen:
Could you re-apply the PR on current master, but only include the `text_width_estimate()` improvement, without touching the rendering or including chapter title? Let's leave that for #93.

PR #112, opened 2022-08-15
Better text_width_estimate

Natural-Harmonia-Gropius:
The first byte of character greater than or equal to `1100 0000` will be estimated as double width.
https://en.wikipedia.org/wiki/UTF-8#Encoding

[Advance measure](https://developer.mozilla.org/en-US/docs/Glossary/advance_measure) is a better way to do this. But for now, this is good enough.


tomasklaen:
Thank you!

PR #115, opened 2022-08-16
Menu width configuration

christoph-heinrich:
I was wondering why the menu _appears_ wider when windowed then when in fullscreen.
Turns out they actually were the same width `menu_min_width`, but in windowed it felt too wide because of the change in font size and item height. I made the width configurable and introduced a new one for fullscreen that is roughly 1.4 times the size of the windowed width, because item height and font size also change by that factor when in fullscreen.

While trying out different minimum widths, one of the entries with a submenu got clipped by the submenu icon.
Turns out that icon was not considered during the width estimation.

Now people can make their menu as narrow as the text allows.

I personally prefer
```
menu_min_width=185
menu_min_width_fullscreen=260
```
but I left the default for windowed the way it was.

### DIFF
--- a/uosc.lua
+++ b/uosc.lua
@@ -69,9 +69,12 @@ speed_font_scale=1
 # controls all menus, such as context menu, subtitle loader/selector, etc
 menu_item_height=36
 menu_item_height_fullscreen=50
+menu_min_width=260
+menu_min_width_fullscreen=360
 menu_wasd_navigation=no
 menu_hjkl_navigation=no
 menu_opacity=0.8
+menu_parent_opacity=0.4
 menu_font_scale=1
 
 # menu button widget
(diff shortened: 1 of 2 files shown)


uosc.lua
```diff
@@ -1052,8 +1056,12 @@ function Menu:open(items, open_item, opts)
 			local estimated_max_width = 0
 			for _, item in ipairs(this.items) do
 				local spacings_in_item = item.hint and 3 or 2
+				local has_submenu = item.items ~= nil
+				-- M as a stand in for icon
+				local rhs = item.hint or (has_submenu and 'M' or nil)
+				local rhs_size = item.hint and this.font_size_hint or this.font_size
```
tomasklaen, line 1062: Use words. I can't tell what `rhs` stands for 😶.

christoph-heinrich: rhs stands for right hand side. I changed it to hint_icon.

PR #117, opened 2022-08-18
Fix accessing member of nil value.

tinkernels:
ref: load-subtitles

### DIFF
--- a/uosc.lua
+++ b/uosc.lua
@@ -3378,7 +3378,7 @@ mp.add_key_binding(nil, 'load-subtitles', function()
 		path = os.getenv("HOME")
 	end
 	open_file_navigation_menu(
-		serialize_path(path).dirname,
+		path,
 		function(path) mp.commandv('sub-add', path) end,
 		{
 			type = 'load-subtitles',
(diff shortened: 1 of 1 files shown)


tomasklaen:
I think this needs a bit different approach:

```lua
if path then
	if is_protocol(path) then
		path = false
	else
		local serialized_path = serialize_path(path)
		path = serialized_path ~= nil and serialized_path.dirname or false
	end
end
if not path then
	path = os.getenv("HOME")
end
open_file_navigation_menu(
	path,
```

Just paste this there, test that it works, and update the PR.

tinkernels:
> I think this needs a bit different approach:
> 
> ```lua
> if path then
> 	if is_protocol(path) then
> 		path = false
> 	else
> 		local serialized_path = serialize_path(path)
> 		path = serialized_path ~= nil and serialized_path.dirname or false
> 	end
> end
> if not path then
> 	path = os.getenv("HOME")
> end
> open_file_navigation_menu(
> 	path,
> ```
> 
> Just paste this there, test that it works, and update the PR.

Modified and tested here.

tomasklaen:
You forgot to pass the `path` we've just spent the last 11 lines normalizing into `open_file_navigation_menu(path)` as in my snippet. It's still accepting `serialize_path(path).dirname`.

tinkernels:
> You forgot to pass the `path` we've just spent the last 11 lines normalizing into `open_file_navigation_menu(path)` as in my snippet. It's still accepting `serialize_path(path).dirname`.

Modified now.

PR #119, opened 2022-08-20
Custom chapter size

po5:
Lets the user customize the width of chapter markers.  
Also handles the case of not overlapping chapters that are very close together (in `timeline_chapters=lines` style, dots can still overlap).

### DIFF
--- a/uosc.conf
+++ b/uosc.conf
@@ -30,6 +30,7 @@ timeline_font_scale=1
 # timeline chapters style: dots, lines, lines-top, lines-bottom
 timeline_chapters=dots
 timeline_chapters_opacity=0.2
+timeline_chapters_width=1
 # which timeline elements to render, and in what order
 # available: ranges,chapters,progress,cache,time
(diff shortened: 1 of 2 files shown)


tomasklaen:
Set the default `timeline_chapters_width` to `6`. Then split the current `chapter_size` variable into `chapter_width` and `chapter_height`, and determine their values in the if/else switch above the rendering. For example for dots:

```diff
- chapter_size = math.min(6, (foreground_size / 2) + 1)
+ chapter_width = math.min(options.timeline_chapters_width, (foreground_size / 2) + 1)
```

Then use them in rendering.

And remove the overlapping prevention. It's valuable to have a visual cue that there are multiple chapters on top of each other, otherwise it'll look like a chapter range bar.

tomasklaen:
Sorry I've touched the whole code block with chapters in last commit, you'll have to rebase 😯.

tomasklaen:
Had a bit of time and wanted to include this in the next release, so I went ahead.

PR #121, opened 2022-08-20
Add speed_bar_offset option

po5:
Lets the user reposition the speed bar (e.g. `speed_bar_offset=82` to display it closer to the top bar).

### DIFF
--- a/uosc.conf
+++ b/uosc.conf
@@ -48,6 +48,7 @@ volume_font_scale=1
 speed=no
 speed_size=46
 speed_size_fullscreen=68
+speed_bar_offset=0
 speed_persistency=
 speed_opacity=1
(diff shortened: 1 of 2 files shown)


tomasklaen:
I want all timeline adjacent elements to fade in with the timeline, otherwise there's too many things popping in and out, making it look like a whack-a-mole.

And regarding the offset, the use here would be to position it to the top, but the speed bar is currently not designed to be at the top. And I don't think supporting that adds much value to warrant the rendering code required to flip it vertically.

At least not now. Maybe I'll reconsider after implementing menu builder, since users that put too many things there might want to free up space by moving sped to the top.

Regardless, I definitely don't want to allow and maintain offsetting here. Sorry about that.

PR #122, opened 2022-08-20
Add text_border_size option

po5:
As this comes from my old personal fork, I don't know if I caught all relevant cases.


po5:
Adding a border to the speed widget, top bar etc. would be a better option.

PR #123, opened 2022-08-20
Hovered chapter title fallback near start

po5:
When hovering to the left edge, displays the first chapter's title even if that chapter is tagged a few milliseconds after 00:00:00.00.  
It may be an optimization to break early if `chapter.time >= 1`, but this is such a niche case that adding a confusing condition isn't warranted.

Using extra variables and only setting the title if everything fails, the naive approach below would incorrectly always display the first chapter when there are multiple chapters within the first second.
```diff
-	if hovered_seconds >= chapter.time then
+	if hovered_seconds >= chapter.time or (hovered_seconds < 1 and chapter.time < 1) then
```
If you have a nicer approach that wouldn't require going through the chapter list twice, I will update this PR.

### DIFF
--- a/uosc.lua
+++ b/uosc.lua
@@ -1662,15 +1662,26 @@ function render_timeline(this)
 		local hovered_seconds = state.duration * (cursor.x / display.width)
 		local chapter_title = ''
 		local chapter_title_width = 0
+		local chapter_title_0 = ''
+		local chapter_title_width_0 = 0
+		local chapter_time_0 = 1
 		if (rendered_chapters and state.chapters) then
 			for i = #state.chapters, 1, -1 do
 				local chapter = state.chapters[i]
+				if chapter.time < chapter_time_0 then
+					chapter_title_0 = chapter.title_wrapped
+					chapter_title_width_0 = chapter.title_wrapped_width
+				end
 				if hovered_seconds >= chapter.time then
 					chapter_title = chapter.title_wrapped
 					chapter_title_width = chapter.title_wrapped_width
 					break
 				end
 			end
+		end
+		if chapter_title_width == 0 then
+			chapter_title = chapter_title_0
+			chapter_title_width = chapter_title_width_0
 		end
 		local time_formatted = mp.format_time(hovered_seconds)
 		local margin_time = text_width_estimate(time_formatted, this.font_size) / 2


tomasklaen:
So if we are hovering at `00:00:00.00`, but 1st chapter is set at `00:00:00.50`, display it anyway?

Hm... I don't know. What if we are dealing with insanely short clip with high framerate and the chapter timing is intended to be this way? What if chapter actually starts at `0.50`?

I'd rather be correct than pretend to know what chapter author intended.

But if this is a common issue, for example if there is some tool that is known to always set starting chapter at something above `0ms`, I might reconsider.

Anyone else has an input on this?

---

And regarding the solution, if we decide that this is a good thing to do, the ideal would I guess be to do this during chapter serialization. Just check if 1st one starts at `<1` seconds and move it to `0`.

po5:
Not sure why framerate would have anything to do with this as chapters work by effective time, they have no concept of framerate. It would still be a chapter marker unhelpfully-close-but-not-snapped to the start time.

I've noticed this on multiple releases I have around, and having to aim for the chapter title to display is annoying.  
This snapping only applies to the first chapter, if its start time is <1s. I don't see the harm in any "inaccuracy" this would bring.  
Having a chapter at 00:00:00 is common, and your theoretical example of a very precisely chaptered video with a chapter intended a few ms after start would no doubt also include another chapter at real 00:00:00.00.

Will write the serialization-time solution when I can.

tomasklaen:
> framerate would have anything to do with this as chapters work by effective time, they have no concept of framerate

I've meant stuff like super high framerate slow motion videos which might have tons of chapters in 1st second (bullet hits, bullet exits, ...). In this video, there might be a chapter that starts at `0.5`, and truly is meant to start at `0.5`, but your proposal would display it at the whole duration of `0-0.5`. But are these kind of videos even encoded in 1:1 time? It might be that their time is just scaled so that they are still like ~30 fps, but take a lot longer to play than their real time. Whatever the case, there are potential issues with ignoring chapter timing.

And I need to clarify what is this actually trying to fix, because I'm not getting it clearly from the opening post. Are you just annoyed that there is a couple of pixels at the beginning of timeline that don't display a chapter title when you hover over them?

christoph-heinrich:
I think the best way about this would be to move the first chapter to 0 in the chapter serialization, if it is sufficiently close to 0, instead of modifying the code for hovering.

Regarding high framerate videos, both varieties exist, but i believe the one where frame times are stretched out to result in ~30fps is more common then having an actual 1000fps video.

To prevent any problems with very short high framerate videos, we could could not only have `if first_chapter.time < 1` but also include something like `first_chapter.time < state.duration/100`.

I have never encountered a video with a chapter > 0 and < 1 though.

tomasklaen:
But again, what is this fixing exactly? That you have a couple pixels at the beginning where you don't get a chapter title when hovering? That doesn't seem to warrant the extra conditionals and sacrificing correctness, which might be undesirable to many.

For example, if I'm authoring some video, and I open it in mpv, and I see the 1st chapter is not exactly at the start due to fat fingering when placing it in some software, I'll go and fix it, but I'll never notice if mpv hides this. Or if I'm the rare case that needs close to 0 but not exactly 0 chapters, and then I open it in mpv to check and I see them at 0, it'd drive me crazy trying to debug it.

I'd definitely prefer correctness here over UI pretending to know what I want to see.

So if there isn't any other benefit than to not have a couple pixels at the beginning without chapter title, than this is a wontfix.

po5:
Wontfix it is.

PR #124, opened 2022-08-22
File picker support for navigating to other drives on Windows

Hill-98:
Let's bring a better experience to Windows users!

It works on Windows Vista and later.

PS: I'm not a lua programmer, so the code might be a little bad.

### DIFF
--- a/uosc.lua
+++ b/uosc.lua
@@ -3169,9 +3169,9 @@ function open_file_navigation_menu(_directory, handle_select, menu_options)
 	table.sort(directories, word_order_comparator)
 
 	-- Pre-populate items with parent directory selector if not at root
-	local items = is_root and {} or {
-		{title = '..', hint = 'parent dir', value = directory.dirname}
-	}
+	local items = is_root and
+		(state.os == 'windows' and {{title = '..', hint = 'Drives', value = '$drives'}} or {}) or
+		{{title = '..', hint = 'parent dir', value = directory.dirname}}
 
 	for _, dir in ipairs(directories) do
 		local serialized = serialize_path(utils.join_path(directory.path, dir))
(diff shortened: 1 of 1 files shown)


uosc.lua
```diff
@@ -3169,9 +3198,9 @@ function open_file_navigation_menu(_directory, handle_select, menu_options)
 	table.sort(directories, word_order_comparator)
 
 	-- Pre-populate items with parent directory selector if not at root
-	local items = is_root and {} or {
-		{title = '..', hint = 'parent dir', value = directory.dirname}
-	}
+	items = is_root
+		and (function() return state.os == 'windows' and {{title = '..', hint = 'Drives', value = '$drives'}} or {} end)()
```
tomasklaen, line 3202: Why is this a function and not just a parentheses?

uosc.lua
```diff
@@ -3149,6 +3149,35 @@ end
 -- **active_path** - full path of a file to preselect
 -- Rest of the options are passed to `menu:open()`
 function open_file_navigation_menu(_directory, handle_select, menu_options)
+	local items = {}
+
+	if _directory == '$drives' then
```
tomasklaen, line 3154: This should be a separate function `open_drives_menu()`.

tomasklaen (changes requested):
Cool! Tested it and works fine, just address the nitpicks and I'll merge.

uosc.lua
Hill-98: Are you saying `(state.os == 'windows' and {{title = '..', hint = 'Drives', value = '$drives'}} or {})`?

I tried it, and it does work. Since I'm very new to Lua, I didn't think of some tricks.

I will change it.

uosc.lua
Hill-98: OK, no problem, that's correct!

tomasklaen (approved):

PR #128, opened 2022-08-24
feat: always disable the original osc

hooke007:
We don't need manually set `osc=no` in mpv.conf any more.
And also in case that some actions would re-activate the osc "luckily" on the fly.


FichteFoll:
Could this lead to flickering when first opening the player because the default osc is first shown and then hidden as the plugin is loaded?

hooke007:
I have not observed this phenomenon.

hooke007:
Anyway, I think it is worth because
> some actions would re-activate the osc "luckily" on the fly

Or revert README's part in this commit? In case of some unexpected situations.

tomasklaen:
Not just flickering, but I'd also be afraid that osc might not unregister and unbind all the things it needs to when disabled after the start up.

I'd still like to merge this, but yes, revert the readme change.

PR #140, opened 2022-08-28
feat: add `audio-device` Command

hooke007:


hooke007:
It's not perfect for I have no idea how to implement `active_prop` in this case.

christoph-heinrich (changes requested):
I don't think `active_prop` is needed here, you can simply return an `active_index` in the `list_serializer` like is already the case in the function `create_select_tracklist_type_menu_opener()`.

Not sure about the best way of getting that `active_index`. The only way I have found would be to iterate over the `audio-device-list` and match the `name` with `audio-device` like is done [here](https://gist.github.com/bitingsock/ad58ee5da560ecb922fa4a867ac0ecfd#file-cycle-adevice-lua-L5).

I suspect that won't work when no audio device has been manually selected yet, as the console shows `Autoselect device` for that property. Simply return `nil` for the active_index when it couldn't be found.

Edit: I've just tested it and `audio-device` returns "auto" if it hasn't already been manually set.
Maybe it's also a good idea to filter the list by audio output driver using `current-ao`, but be careful that might be nil, depending on if there is currently an audio output or not.

tomasklaen:
1. Rename to `audio-device`.
2. I don't know about the default menu addition. This is something 99% people will never use. But at the very least, put it below aspect ratio. And if you're adding it to the default menu, also add it to the default menu example in the readme at the same spot. Use `alt+s` as a shortcut.
3. Return the active index from serializer (`return items, active_index`), which should be set to the index of whatever was returned by `mp.get_property_native('audio-device')` (default it to `auto` if it returns `nil` for some reason).

hooke007:
Fixed 1&2.
Unfamiliar with Lua, I was in a bind of point 3.

christoph-heinrich:
To help you out with Lua, I've made a PR over on your fork.
@tomasklaen what do you think of the filtering? I got the idea from [this script](https://gist.github.com/bitingsock/ad58ee5da560ecb922fa4a867ac0ecfd).

hooke007:
@christoph-heinrich Thanks but there is a little issue there. After switching to one wasapi device, other api's selection-items disappear.

christoph-heinrich:
Yes, that's because of the filtering. I've only tested it on my main system, and there were 45 entries, so I thought the filtering idea from that linked script sounds good. Just leave out the last commit if you don't want that. With filtering I was down to 5 and honestly I don't see why one would want to change the api, but maybe there are some use cases that I'm not aware of.

hooke007:
Just my obsolete usage.
I think it's up to @tomasklaen to make final decision.

tomasklaen:
The filtering looks good.

I'd also want to do something about those huge hints. Since we are filtering only the current driver, the `wasapi/` part is redundant. And the ID should probably also be truncated. Similar to how github shortens commit IDs: truncate to first dash, and if the output is longer than **20** chars, then truncate that to first **10**.

There's already a split util above so you can just use that (`split(value, '-')`).

The final hint should look maybe like `{5d8b9r9v...}`.

christoph-heinrich:
Not everyones devices look like this. For me it looks like:
device_list
And for alsa there are devices with hints like:
device_hint
So not sure about truncating on `-`.

With your suggestions mine would look like this (except less wide, I forgot shortening it during editing):
device_list_edited

tomasklaen:
Yes, that would be dumb. So only drop the `driver/` prefix. But only if it's a prefix. Keep it if it's the sole part of the hint (2nd item on screenshots above).

christoph-heinrich:
Prefix is now stripped out if available.

christoph-heinrich:
> > one would want to change the api
> 
> Just my obsolete usage.

I tried changing the `ao` via selecting a different one for the audio-device, but it doesn't work. I don't get an error or anything, it just seems to fail silently and then falls back to the original `ao`.
Based on that we didn't actually remove any possible uses for the menu by filtering by `current-ao`. I was a bit worried about that as this is your PR and your comment indicated that you would like to do that.
Setting `ao` during runtime does work however, in case you actually need that.

hooke007:
> I was a bit worried about that as this is your PR and your comment indicated that you would like to do that.

Never mind, for it's pushed to tomasklaen's master branch, just follow his preference.
@tomasklaen If no more thoughts I would merge christoph-heinrich's commit into this pr.

tomasklaen:
> If no more thoughts I would merge christoph-heinrich's commit into this pr.

Yes, that's what I'm waiting for :)

hooke007:
Done

PR #145, opened 2022-08-30
Timeline progress smoothness

christoph-heinrich:
Ever since implementing #137 I felt like the trailing edge sometimes wasn't as smooth as the leading edge for some reason.

I've now finally looked into it and the brightness of the video in the background influences how smooth it looks, and that's why it wasn't always the same.
Removing the clipping from the timeline background resolves the issue, but that's not really a proper solution.
I've also tried rendering the rectangle in different ways (e.g. the other way around) because at first I suspected it might be a deficiency in libass, but that didn't help.

Any ideas on how to solve this properly?

Also here are two recordings to demonstrate the problem and how it looks with the clipping removed.
With clipping:

https://user-images.githubusercontent.com/8932183/187437694-f9912113-deae-4f91-a3b5-91dcb8abbad2.mp4

Without clipping:

https://user-images.githubusercontent.com/8932183/187437739-2405a27d-205c-4f04-85df-03d6eb8e4973.mp4


tomasklaen:
The issue is that ass `iclip()` on background sometimes doesn't exactly match fg coordinates, even when they are both the exact same rounded integers. There is an off by one error somewhere in libass probably. I was trying to fix it sometime last week and it drew me insane for like half an hour until I gave up.

The solution is I guess to use ass clipping only for timestamps, and for background, draw 2 rectangles: one before the line and one after, instead of drawing just one and clipping the fg coordinates. That way we are not dealing with ass clipping, and everything should match perfectly.

tomasklaen:
Actually, that was an issue **with** rounding. When you submitted the PR removing it, I knew it's gonna be blurry, but it's blurry in a different way because of this off by one iclip thing anyway, and removing rounding will at least make it smooth, so whatever.

Rendering 2 rectangles instead of 1 with iclip might still help though.

christoph-heinrich:
It seems to me as if libass either completely clips a pixel or doesn't clip it at all (so no interpolation with opacity or anything).
It also seems to always round the number down, which is why it hasn't been a problem for the leading edge, as that would only get clipped when the pixel already gets fully drawn, but on the trailing edge we would end up with pixels that are barely drawn but still clipped. I've changed the `foreground_coordinates` to ceil and floor the x coordinates and that works well in my testing.
Drawing more rectangles instead of clipping still seems like the better solution though.
Clipping based on rounded numbers looks better for the time text.

christoph-heinrich:
Added the background rendering thing as a separate commit in case you want to compare the results, but looks good to me.

tomasklaen:
Can you re-apply on current master? I did some repo refactoring that makes resolving too complex for github.

christoph-heinrich:
Is rebased.

PR #146, opened 2022-08-30
Add border x offset to progress line

christoph-heinrich:
The progress line started at an x of 0 instead of `bax`, resulting in no border at the left side and twice the border at the right side.


tomasklaen:
Can you re-apply on current master? I did some repo refactoring that makes resolving too complex for github.

christoph-heinrich:
Is rebased

PR #150, opened 2022-08-30
fix: cursor line has same position in foreground and background

christoph-heinrich:
The cursor line used to be one pixel further left when on foreground then when on background.
Now it keeps the same position relative to the cursor.


christoph-heinrich:
Omg that is way too obvious, I can't believe that I didn't think of that yesterday.
I've changed it accordingly.

PR #152, opened 2022-08-31
Timeline cursor pixel middle

christoph-heinrich:
I don't know if that's a desirable change or not, but I was wondering why, when seeking to a position and having an odd number as line width, it would end up with blurry edges at the progress line. After all with an odd number there is exactly one pixel in the middle and so it should be able to cleanly display the progress line.

After ruling out faulty time calculation and rendering, it turns out the problem was that cursor x coordinate points at the left edge of the pixel, while what I visualizing was in the middle of the pixel.

Technically this also makes hovered chapters more accurate, because depending how the chapter line gets rounded, without this change it might still show the previous chapter even one pixel after the line.
Not that anyone would ever really notice that without looking reeeaaally closely.

Also this depends on #151. I'll need to rebase after the other one gets merged.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -1728,7 +1728,8 @@ function render_timeline(this)
 
 	-- Hovered time and chapter
 	if (this.proximity_raw == 0 or this.pressed) and not (elements.speed and elements.speed.dragging) then
-		local hovered_seconds = this:get_time_at_x(cursor.x)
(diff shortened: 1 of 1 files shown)


tomasklaen:
Rebase away. Also, add a comment above that line on what is the point of that `+ 0.5`.

christoph-heinrich:
Is rebased and added comment

PR #153, opened 2022-08-31
feat: don't show chapter end on hover

### LINKED ISSUE #125: Don't show title on hover for an "end of range" chapter
The kind of range is only/mostly relevant while the range itself is active, not afterwards. If you use the sponsorblock plugin while watching videos from YouTube, you will occasionally see a sponsor or a self-plug range or similar where the "end" chapter will be shown on hover for almost the entire video.

It would be more useful if instead uosc showed "intro segment (info)" after the first marker and nothing after the second marker of a range. I don't think it makes sense to apply a change here to the sponsorblock plugin because technically it does use the correct way of implementing this and when navigating the chapter menu for example, being able to select the "… end" marker is a use case. It's just that with the new UI that it doesn't work too well.

Example video: https://www.youtube.com/watch?v=oSSCRUmaquo

cc @po5

christoph-heinrich:
Chapters that end a chapter range and are matched by something other
then `.*` are now not shown when hovering over that chapter.

Closes #125

### DIFF
--- a/script-opts/uosc.conf
+++ b/script-opts/uosc.conf
@@ -152,6 +152,10 @@ default_directory=~/
 # They only need to occur in a title, not match it completely.
 # Matching is case insensitive.
 #
+# Chapters that end a range and got matched by an `end_pattern` other then `.*`
+# will not be shown on hover, unless that chapter is also used to start another
+# range.
+#
 # `color` is a `bbggrr` hexadecimal color code.
 # `opacity` is a float number from 0 to 1.
 #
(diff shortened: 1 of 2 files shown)


FichteFoll:
Seems to work for me. :+1: 

However, the play and pause icons seem to be broken on this branch.

christoph-heinrich:
Added documentation for that behavior to `uosc.conf`

tomasklaen:
Well, lets see how soon someone creates an issue "My chapter tooltip is not showing up" 😄.

PR #155, opened 2022-08-31
fix: compatiblity with `unpack` and `table.unpack`

christoph-heinrich:
In Lua 5.2 the unpack global has been moved to table.unpack
To keep compatiblity with 5.1 and newer, choose depending on what's
available.


tomasklaen:
Thx for caching these.

PR #156, opened 2022-08-31
fix: nil error in `show-menu` api

christoph-heinrich:
The local `menu` shadowed the global `menu`, thus accesing it's methods didn't work.


tomasklaen:
Damn 🥺.

PR #162, opened 2022-09-01
refactor: move chapter range handling into loop and break early

christoph-heinrich:
This was originally intended for the other PR, but I had to quickly do something before committing/pushing and by the time I got back it was already merged, so another PR it is. Wasn't sure if you would approve of that change anyway, so it's probably better to be separate.


tomasklaen:
Sure, looks much simpler.

PR #164, opened 2022-09-01
fix: simplify the feasible workaround for smooth

hooke007:


hooke007:
uosc's rendering can be smooth without interpolation. For it is only the fast framerepeat and there is no extra comsumption for 'generating the new frame'.

christoph-heinrich:
I've tested it and you're right, in fact any of the `display-*` options result in a smooth interface.

hooke007:
Yeah, `display-resample` ought to be the the most robust one except `audio` for majority.

There are other solutions such as `vf=fps=fps=60` but too hacky to as a suggestion.

christoph-heinrich:
True, `display-tempo` might also enter that list one day, but with audio filters not doing so well with speeds very close to 1, it'll probably be a niche option (if it would finally get merged).

PR #165, opened 2022-09-02
feat: better default config for chapter ranges

christoph-heinrich:
I have recently noticed that a lot of videos have "Intro" and "Outro" chapters, Should we match on those as well?

### DIFF
--- a/script-opts/uosc.conf
+++ b/script-opts/uosc.conf
@@ -170,4 +170,4 @@ default_directory=~/
 # ```
 # chapter_ranges=sponsor start<3535a5:.5>sponsor end, segment start<3535a5:0.5>segment end
 # ```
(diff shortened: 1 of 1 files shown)


tomasklaen:
Sure, but match it exclusively with `^intro&`. Don't want it to match stuff like `Intro to xxx`.

christoph-heinrich:
Intro and outro are now also matched, with the same colors as opening and ending.

For my own config I was a bit more liberal with the patterns and also try a different color.
```
^intro$|intro:? | intro$<7ca535:0.5>.*, ^outro$<7ca535:0.5>.*|{eof}
```
But I haven't tried how well that would work with the default config, as the cached ranges are also green by default.
Looks like this for me:
intro
Not sure about the color though.

PR #166, opened 2022-09-02
feat: support `osc-margins` shared-script-property

christoph-heinrich:
The documentation cautions against using shared script properties, but
console.lua relies on the margin to not overlap with the bottom bar from
the default osc. Without support for that in uosc, using the console
was always annoying.
Now it also doesn't overlap with the uosc timeline.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -162,6 +162,8 @@ local state = {
 	first_real_mouse_move_received = false,
 	playlist_count = 0,
 	playlist_pos = 0,
+	margin_top = 0,
+	margin_bottom = 0,
 }
 local forced_key_bindings -- defined at the bottom next to events
 
@@ -1512,6 +1514,27 @@ function update_human_times()
 	end
 end
 
+function update_margins()
+	-- margins are normalized to window size
+	local top, bottom = 0, 0
+
+	local timeline_size = elements.timeline:get_effective_size() or 0
+	bottom = (timeline_size / display.height)
+
+	local top_bar = elements.top_bar
+	if top_bar.enabled and top_bar:get_effective_proximity() ~= 0 then
+		top = (top_bar.size or 0) / display.height
+	end
+
+	if top == state.margin_top and bottom == state.margin_bottom then return end
+
+	state.margin_top = top
+	state.margin_bottom = bottom
+
+	utils.shared_script_property_set("osc-margins",
+		string.format("%f,%f,%f,%f", 0, 0, top, bottom))
+end
+
 -- ELEMENT RENDERERS
 
 function render_timeline(this)
@@ -2248,6 +2271,8 @@ function render()
 	osd.data = ass.text
 	osd.z = 2000
 	osd:update()
+
+	update_margins()
 end
 
 -- STATIC ELEMENTS


tomasklaen:
Isn't it a concern to set this on each frame? Shouldn't we only set when it changes?

tomasklaen:
I'd even be for just setting it once on load, using the max size options for timeline and top bar. That should be a simple one-liner and achieve all we want to achieve with no performance or bloat hit.

tomasklaen:
Oh, they are fractions. So I guess on display resolution change then.

christoph-heinrich:
The osc updates them on initializaion, in the render after a change to fullscreen and after a change to visibility mode, visibility.

I would certainly prefer not just setting it to the max size, as if both elements are always displayed in their max size.

I'll try to limit how often it gets called, without having inserting some sort of `request_margin_update()` all over the place.

tomasklaen:
> The osc updates them on initializaion, in the render after a change to fullscreen and after a change to visibility mode, visibility.

Does it do that even when disabled? Are we fighting each other then?

christoph-heinrich:
> Does it do that even when disabled? Are we fighting each other then?

I've tested it and with `osc=no` the script doesn't even get loaded, so nothing to worry about there.
Because of to #128 there is also the possibility of both scripts fully initializing, however in my testing that isn't actually a problem.
osc.lua sets the script property a few times at the very beginning and then nothing afterwards, doesn't matter what I'm doing.
In fact it is as if it stops getting any kind of updates whatsoever, so I guess mpv must have special handling for osc to just completely disable it, even if `osc=no` gets set at runtime, because the code of the osc doesn't have any special handling for unobserving all the properties etc. It even receives the 'shutdown' event.

I ended up going down the `request_margin_update()` road, but as it turns out the only way to update every time the size of the timeline changes, is to update every time it gets rendered. So instead I added the margin size to the state, and then only update the script property when one of the two actually changed.

po5:
It's a nice feature and works as described, but the extended commit message is misleading.

> The documentation cautions against using shared script properties, but
> console.lua relies on the margin to not overlap with the bottom bar from
> the default osc.

None of the `osc-layout` options by default make the console script move away from mpv's osc, in fact if you read the osc.lua code you'll see that the `osc-margins` values it sets come from `video-margin-ratio-*` properties.  
https://github.com/mpv-player/mpv/blob/master/player/lua/osc.lua#L161-L166  
https://github.com/mpv-player/mpv/blob/master/player/lua/osc.lua#L2171-L2190

Only with `osc-visibility=always` and `osc-layout=bottombar` does this behavior seemingly kick in without setting `video-margin-ratio-*`.

https://github.com/mpv-player/mpv/commit/4787eb594a879619b9acd266bac015bfd0032c50

> If the OSC is not set to permanently visible, export no margins, because
> it would look weird to move the console depending on the mouse movement.

This script property seems more like a hacky solution to the specific problem they ran into with console.lua, with this specific osc configuration, rather than something we should actually re-use.  
May be worth extending the behavior in mpv's original osc, so that other scripts can observe the osc's (and in the future all 3rd party osc's) actual reserved space as to not overlap with custom UI elements, for moving subtitles out of the way of the display through `sub-pos`, etc.

christoph-heinrich:
> None of the `osc-layout` options by default make the console script move away from mpv's osc

All of the `osc-layout` options set the top margin, I tested it with e.g. `mpv --no-config --script-opts=osc-layout=box --border=no`, but the `topbar` layout advertises a top margin that is a little bit bigger then it actually is.

> in fact if you read the osc.lua code you'll see that the `osc-margins` values it sets come from `video-margin-ratio-*` properties. 

In fact if you read the osc.lua code, you'll see that the values for the `margins` table, which is used for setting the `osc-margins` shared script property,  do not get set in the code you've linked. That code is for **setting** the [`video-margin-ratio-*`](https://mpv.io/manual/stable/#options-video-margin-ratio-left) properties, not reading from them. And they do get set to the same values as the osc-margins if `boxvideo` is used. They are there to scale the video down to prevent it from overlapping with the interface.

> Only with `osc-visibility=always` and `osc-layout=bottombar` does this behavior seemingly kick in.

`bottombar` is the only layout that sets the bottom margin. Not sure why that decision was made. Maybe they thought that `box` and `slimbox` are supposed to hover over everything anyways. Or maybe they are too far up to move other things out of the way for them? If that bothers you, you can ask on the on the IRC why that is and maybe even make a PR, but the mpv devs are not too fond of touching the osc code (because it's too fragile, according to them).

And yes they only set the margins when `osc-visibility=always`. Makes more sense this way for them, as the UI is only temporarily visible with `osc-visibility=auto` and that would cause things that respect the margin (like the console) to jump around too much.
But I think for the way uosc works, always setting them is the way to go.

> This script property seems more like a hacky solution to the specific problem they ran into with console.lua, with this specific osc configuration, rather than something we should actually re-use.

That would be why it's not documented :wink:, but I don't see a downside to supporting it here. It's not like that code is particularly invasive on the rest of the code or anything.

> May be worth extending the behavior in mpv's original osc, so that other scripts can observe the osc's (and in the future all 3rd party osc's) actual reserved space as to not overlap with custom UI elements, for moving subtitles out of the way of the display through `sub-pos`, etc.

Do you mean advertising the space that it _may_ need, even if it currently does not? Sounds useful.

PR #172, opened 2022-09-04
fix: pixel to time calculation

christoph-heinrich:
While things were sort of fine with `timeline_style=line`,
it was impossible to set the time to 0% or 100% when using
`timeline_style=bar`.

Now the time calculation is correct regardless of style and the
rendering is pixel perfect.

However I'm not entirely sure about the way the `timeline_style=bar` gets rendered now. It is technically correct in the sense that mouse input can go from 0% to 100% and the bar always gets displayed perfectly to the interpolated point between those two mouse input positions, so when seeking the result always lines up perfectly with the mouse input.

But that also means that when seeking to 0% the timeline bar isn't completely empty anymore, and when seeking to 100% it is also not completely full. Instead it always draws half a pixel at the edge.

Alternatively we could:
1. set the line width in the rendering part to 0, resulting in the same rendering as it used to be (empty bar at 0% and completely full at 100%)
2. not add the `line_width / 2` to `fbx`, resulting in an empty bar at 0%, but with the last pixel not drawn at all at 100%
3. add `line_width` instead of `line_width / 2` to `fbx`, resulting in the first pixel being fully drawn at 0%, and the last pixel being fully drawn at 100%

2 and 3 result in a clean edge when seeking with the mouse, while 1 basically has the pixel at the edge interpolated over the whole width (e.g. being half drawn at 50% time).

Let me know which one you prefer.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -2629,9 +2629,11 @@ function Timeline:update_dimensions()
 end
 
 function Timeline:get_time_at_x(x)
-	-- padding serves the purpose of matching cursor to timeline_style=line exactly
-	local padding = (options.timeline_style == 'line' and self:get_effective_line_width() or 0) / 2
-	local progress = math.max(0, math.min((x - self.ax - padding) / (self.width - padding * 2), 1))
+	-- line width 1 for timeline_style=bar so mouse input can go all the way from 0 to 1 progress
+	local line_width = (options.timeline_style == 'line' and self:get_effective_line_width() or 1)
+	local time_width = self.width - line_width
+	local progress_x = x - self.ax - line_width / 2
+	local progress = math.max(0, math.min(progress_x / time_width, 1))
 	return state.duration * progress
 end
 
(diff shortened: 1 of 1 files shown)


tomasklaen:
I think you're looking at this from the wrong angle. `this.width * progress` is the correct representation of current time, and that can't change. If there are cursor-progress alignment issues, the fix needs to happen in `set_from_cursor()`.

christoph-heinrich:
That's exactly the behavior of alternative 1, which results in `fbx = this.width * progress + bax` when the whole line width thing is simply removed from the rendering for `timeline_style=bar` (so exactly what it used to be).

That was my preferred option too.

christoph-heinrich:
It's changed now.

tomasklaen:
Am I dumb, or is this producing the exact same numbers as the current behavior? So nothing changed?

christoph-heinrich:
The progress calculation for mouse input with `timeline_style=bar` should be different. It used to be the case that seeking all the way to the left would calculate a progress of (I think) 0.0005something%, and seeking all the way to the right would calculate a progress of 0.000995something%, which isn't off by a lot, but it's enough to not end up at the very first or the very last frame.

While working on this I ended up drawing things up on paper and redoing all the progress and corrdinate calculations from scratch, that's how the change to the line coordinate calculation came to be. I think it's a bit easier to understand that way, but the result should be the same, so I can undo that again if you want to.

I'll have time to fix up my open PRs in 6-ish hours.

christoph-heinrich:
With `timeline_style=bar` on master in fullscreen I can seek from 0.00026041666... up to 0.99973958333...
With this change it goes from 0 to 1.

The reason for that is that mouse input goes from 0.5 pixels to width-0.5 pixels, which needs to be considered during the time calculation.

I didn't test that now, but before that 0.5 pixel mouse input change it must have been possible to seek to 0, but impossible to seek to 1 in fullscreen.

PR #174, opened 2022-09-05
fix: timeline render order

christoph-heinrich:
I was wondering why chapter seem to disappear when they are on the progress line and the timeline is not entirely minimized.

In my testing the visibility of the progress line isn't a problem with this order. And the only chapter style that would cover the progress bar from top to bottom is 'lines', so even if the chapter was completely opaque it would only become a problem if very thick lines were used.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -1649,37 +1649,31 @@ function render_timeline(this)
 	ass:draw_stop()
 
 	-- Progress
-	local function render_progress()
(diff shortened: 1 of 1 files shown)


tomasklaen:
Well, ok. But if we are not shifting rendering order, than these don't have to be functions anymore. So just un-function those code blocks.

christoph-heinrich:
The code has been un-functioned and then rebased because of merge conflicts.
Also wanted to mention that #175 solved the visibility problem with thick opaque chapter lines.

tomasklaen:
This is now part of the refactor.

PR #175, opened 2022-09-05
fix: fancy chapter rendering

christoph-heinrich:
Differentiate between parts of chapter indicator that are covered by the
progress line/bar and parts that are not.

Looks much better that way, particularly with wide indicators (e.g. the default 6 wide dots).
Performance shouldn't be that bad, as it does bounds checks and only renders what's necessary.

Not sure if this is "fix" or "feat"


christoph-heinrich:
Don't merge yet. a line with width 1 isn't visible on top of the progress bar. I'll have to look into that. (width 2 renders no problem interestingly)

christoph-heinrich:
Even better now :smile:

PR #185, opened 2022-09-10
Add more border and font scale options

po5:
Defaults unchanged, added options for everything I thought useful to tweak.

### DIFF
--- a/script-opts/uosc.conf
+++ b/script-opts/uosc.conf
@@ -96,6 +96,7 @@ volume_font_scale=1
 # Playback speed widget: mouse drag or wheel to change, click to reset
 speed_persistency=
 speed_opacity=1
+speed_border=1
 speed_step=0.1
 speed_step_is_factor=no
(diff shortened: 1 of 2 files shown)


tomasklaen:
I don't want every pixel on the screen to have a configuration option. I'm trying to find a balance to satisfy picky people while still having as little options as possible, and I'm kind of loosing. The config file is quite overwhelming already.

The only option from these that I wouldn't mind adding is `top_bar_font_scale`, as we control user's font rendering. The rest is now either not needed anymore after the recent changes, or should be determined by the UI.

po5:
I agree there are way too many options, a (set of?) unified font scale option(s) with an actual impact on *all* text would be great.  
Haven't had a chance to play around with the refactor much, will complain when necessary :^)

tomasklaen:
Yeah consolidating all `*_font_scale` options into one is not a bad idea.

PR #186, opened 2022-09-11
fix: `has_playlist` control disposition

christoph-heinrich:
I kept changes to a minimum for now, but there are a bunch of places that check for `state.playlist_count > 1` that could now be replaced with `state.has_playlist`.

Also there are a few places that use `mp.get_property_native('playlist-count')` instead of `state.playlist_count`. Is there a good reason for that or should those be replaced with the state property?

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -387,6 +387,7 @@ local state = {
 	is_image = nil,
 	is_stream = nil,
 	has_audio = nil,
+	has_playlist = nil,
 	cursor_autohide_timer = mp.add_timeout(mp.get_property_native('cursor-autohide') / 1000, function()
 		if not options.autohide then return end
 		handle_mouse_leave()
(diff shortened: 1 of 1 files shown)


tomasklaen:
> should those be replaced

Yeah, probably old leftovers from when the value was not in the state map.

christoph-heinrich:
I was bothered by the fact that `state.has_playlist` is redundant with the existence of `state.playlist_count`, so I changed it.
In exchange the implementation of the check now has an `if`. Not sure which variant is preferable :neutral_face:

tomasklaen:
No, I want the `has_playlist` back 😶.

christoph-heinrich:
Ok, but I'll wait with that until your changes land. Then I can also replace all instances of `state.playlist_count > 1` with `state.has_playlist` without running into conflicts.

tomasklaen:
Refactor now on main. I'm expecting tons of bugs in it 👼.

christoph-heinrich:
Rebased, brought back `has_playlist` and all the `playlist-count` related changes applied.

PR #195, opened 2022-09-14
Revert border_opacity for ass:rect()

christoph-heinrich:
The border for the speed notches was always drawn with full opacity. That was particularly noticeable when the speed widget was barely visible.
before
speed_before
after
speed_after

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -1028,7 +1028,7 @@ end
 ---@param ay number
 ---@param bx number
 ---@param by number
----@param opts? {color?: string; border?: number; border_color?: string; opacity?: number; border_opacity?: number; clip?: string, radius?: number}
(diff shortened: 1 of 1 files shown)


tomasklaen:
This bug was introduced when I decided to change `ass_mt:rect()` opacity prop to apply as `\1a` (primary color alpha) instead of `\alpha` (alpha of all components), because I needed border alpha for something to be separate, and at the time it seemed ok to do. That something didn't make it in, so border alpha is not needed, and making `opacity` set the background opacity is wrong I think.

If I'd want to set background opacity from border opacity separately, they should be a separate props from `opacity`.

But right now, nothing needs these, so just fix it by reverting the `opacity` prop to apply `\alpha` again, and remove `border_opacity` from `ass_mt:rect()` altogether.

christoph-heinrich:
`border_opacity` has been reverted now, `rect()` uses `\alpha` again.

PR #197, opened 2022-09-14
fix: chapter ranges

christoph-heinrich:
I can't tell what the plan was here, but rendering and calling serialization still referenced `state.chapter_ranges`, so I simply moved it back into the state to restore the functionality.

Edit: If you want to move it out of the state, might I suggest moving it into the timeline? Because that's the only element that cares about them.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -373,6 +373,7 @@ local state = {
 	duration_or_remaining_time_human = nil, -- depends on options.total_time
 	pause = mp.get_property_native('pause'),
 	chapters = nil,
+	chapter_ranges = nil,
 	border = mp.get_property_native('border'),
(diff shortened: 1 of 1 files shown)


tomasklaen:
I moved it out of state because it didn't make sense to me there. State is for properties that change during session, while chapter_ranges are a constructed serializers that are static per session. They are they own thing so should be their own thing.

Timeline also shouldn't be concerned with creating and storing chapter range serializers, maybe only using them.

So the best to do here imo would be to keep chapter range serializers as their own thing, and make timeline tap into chapters and use the serializers to generate ranges, which will then be stored on the timeline.

Currently chapter range serializers store the ranges on themselves, which is also weird.

tomasklaen:
Don't waste time, I'll be reworking chapter ranges tomorrow.

Issue #177, opened 2022-09-09
Allow customization top_bar_title 


maaldheefee:

tomasklaen:
How? There's nothing to customize. I won't be adding a different color option just for title, and lowering opacity of the background will make the text unreadable when on top of other text, see #168.

maaldheefee:
I meant allow changing of the text using property expansion.

tomasklaen:
You mean this #24?

maaldheefee:
Yes

dyphire:
@tomasklaen I think it is necessary to set the `--title` to be customizable, because in some cases the media title is not useful information. 
For example:
![](https://cdn.jsdelivr.net/gh/dyphire/PicGo/img/2022/09/14/20220914202347.png)
![](https://cdn.jsdelivr.net/gh/dyphire/PicGo/img/2022/09/14/20220914202357.png)
As you can see, the title does not show useful content. This should be displayed as the `{filename}`

dyphire:
> @tomasklaen I think it is necessary to set the `--title` to be customizable, because in some cases the `media-title` is not useful information. 

At the same time, the streaming file should always use the `media-title`, because the `filename` is also not a valid information for it.
If the user is not allowed to customize the `--title`, I think it is more correct to always display the `filename` first, and only force the `media-title` to be used for streaming files.

dyphire:
> But let's say I give you an option to customize title template. What will you do exactly? Switch it to filename? Then when you open a stream or file downloaded from a service that changed its name to some hash, it will display its URL or that hash as a title

As long as we provide customizable title options, we can do more control. Whether in the input.conf or mpv.conf file.
For example, you can use automatic condition configuration in mpv.conf to switch the title for the stream file.
```
[media-title]
 profile-cond=path:match('://')~=nil or path:match('^magnet:')~=nil  or ...
 profile-restore=copy
 script-opts-append=uosc-title=${media-title}

```
Similarly, we can choose to use the media title as the default value, and switch file names for specific files through automatic condition configuration:
```
[fliename]
 profile-cond=filename:match("cXcY@FRDS")~=nil or ...
 profile-restore=copy
 script-opts-append=uosc-title=${fliename}

```
This is the advantage of the `--title` option provided by mpv's official osc.lua script. 
I don't understand why to restrict user behavior, which denies more possibilities for this script.

tomasklaen:
```
profile-cond=filename:match("cXcY@FRDS")~=nil or ...
[...]

PR #199, opened 2022-09-14
fix: react to changes in the `title` property

christoph-heinrich:
As mentioned [here](https://github.com/tomasklaen/uosc/issues/177#issuecomment-1247147261) it currently does not react to changes in the title property.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -4065,13 +4065,19 @@ end
 mp.set_key_bindings(mouse_keybinds, 'mouse_movement', 'force')
 mp.enable_key_bindings('mouse_movement', 'allow-vo-dragging+allow-hide-cursor')
 
-mp.register_event('file-loaded', function()
-	parse_chapters()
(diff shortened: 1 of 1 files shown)


scripts/uosc.lua
```diff
@@ -4064,14 +4064,20 @@ if options.pause_on_click_shorter_than > 0 then
...
 	set_state('title', mp.command_native({"expand-text", title_template}))
+end
+mp.register_event('file-loaded', function()
+	parse_chapters()
+	update_title(mp.get_property_native('title'))
 end)
 mp.register_event('end-file ', function() set_state('title', nil) end)
+mp.observe_property('title', 'string', function(_, title)
+	-- Don't change title if there is currently none
+	if not state.title then return end
```
tomasklaen, line 4078:
```lua
if state.title then update_title(title) end
```

christoph-heinrich: That was way too obvisous :disappointed:.

PR #200, opened 2022-09-14
fix: mouse input with fractional scaling

### LINKED ISSUE #192: Cannot seek using the timeline when mouse pointer is at the bottom edge of screen when fullscreen
I believe the bottom most pixels should trigger the timeline. Instead triggers my MBTN_LEFT action.

1 pixel from the bottom edge
image

at the bottom edge
image

christoph-heinrich:
This is simply the fix I posted in #192. Seems to work pretty well so far.

Really good hidpi support would need more invasive changes, but this is good enough for now.

Closes #192


tomasklaen:
@hooke007 @maaldheefee can you guys confirm that this fixes it for you?

hooke007:
Works well.

PR #201, opened 2022-09-14
Chapter stuff

christoph-heinrich:
Most of that was already written when you closed the other PR, so I thought not offering it was a bit of a waste.

The first commit mostly came to be because I found it odd that parsing chapters would need the duration. Ended up as a general improvement to chapter procurement. `normalize_chapters()` doesn't really need to return the table, but I left it as it was in case the implementation changes at some point.

The second commit now has the `options.chapter_ranges` replaced with a table of the parsed chapter ranges.
`state.chapter_ranges` now only contains the ranges that have been matched to the current chapters, not the whole definition + matches.
The serialization function has been factored out into a standalone function, instead of storing new function in each `options.chapter_range`.

If it doesn't get merged, maybe parts of it will be of use for your rework tomorrow :wink:

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -326,6 +326,43 @@ local config = {
 			}
 		end
 	end)(),
+	chapter_serializers = (function()
+		-- Parse `chapter_ranges` option into workable data structure
+		local chapter_ranges = {}
+		for _, definition in ipairs(split(options.chapter_ranges, ' *,+ *')) do
+			local start_patterns, color, opacity, end_patterns = string.match(
+				definition, '([^<]+)<(%x%x%x%x%x%x):(%d?%.?%d*)>([^>]+)'
+			)
+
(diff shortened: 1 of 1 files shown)


christoph-heinrich:
It's rebased now because I'm actively using it.

With chapter ranges still being broken, I think it's better if this gets merged. It's not like this hinders you from doing your rework when you get around to it.

tomasklaen:
I didn't want to review and potentially go back and forth on a PR that will get replaced in a day or two, but ok.

Can't replace `options.chapter_ranges` with a table, type checker will complain about assigning table to a string, that's why I extracted this stuff from options. So better to put it on its own `chapter_serializers` variable.

christoph-heinrich:
Weird, the type checker didn't complain to me about it. But since this isn't the only option that gets parsed into a table, it's now treated the same as the other ones.

PR #204, opened 2022-09-15
Support input-conf option

Hill-98:
If you use someone else's set of configuration files (git based), you may want to customize input.conf.


po5:
What a weird feature. Should support it anyway. tomas was fast on the merge :^)

PR #207, opened 2022-09-15
feat: add `has_sub` and `has_chapter` for `{disposition}`

dyphire:


scripts/uosc.lua
```diff
@@ -3129,7 +3131,7 @@ function Controls:serialize()
 		for _, disposition in ipairs(dispositions) do
 			local value = disposition:sub(1, 1) ~= '!'
 			local name = not value and disposition:sub(2) or disposition
-			local prop = (name == 'has_audio' or name == 'has_playlist') and name or 'is_' .. name
+			local prop = (name == 'has_audio' or name == 'has_sub' or name == 'has_chapter' or name == 'has_playlist') and name or 'is_' .. name
```
christoph-heinrich, line 3134: This is over the 120 char limit.

I'd format it like this:
``` Lua
local prop = (name == 'has_audio' or
              name == 'has_sub' or
              name == 'has_chapter' or
              name == 'has_playlist') and name or 'is_' .. name
```

Alternatively this is slowly approaching a point where something like this starts to make sense:
``` Lua
local has_props = {'has_audio', 'has_playlist', 'has_chapter', 'has_sub'}
local prop = itable_index_of(has_props, name) and name or 'is_' .. name
```
But that has some runtime overhead. Don't know which @tomasklaen would prefer.

tomasklaen:
```lua
name:sub(1, 4) == 'has_' ? name or 'is_' .. name
```

scripts/uosc.lua
```diff
@@ -392,6 +392,8 @@ local state = {
 	is_image = nil,
 	is_stream = nil,
 	has_audio = nil,
+	has_sub = nil,
+	has_chapter = nil,
```
tomasklaen, line 396: Flip all of the initial `has_` and `is_` flags to `false`.

scripts/uosc.lua
```diff
@@ -4098,11 +4101,18 @@ mp.observe_property('track-list', 'native', function(name, value)
...
+	set_state('has_sub', has_sub)
 	set_state('is_video', is_video)
 	set_state('is_stream', is_protocol(path))
 	Elements:trigger('dispositions')
 end)
-mp.observe_property('chapter-list', 'native', parse_chapters)
+mp.observe_property('chapter-list', 'native', function(_, chapters)
+	local has_chapter = false
+	if #chapters > 0 then has_chapter = true end
+	set_state('has_chapter', has_chapter)
```
tomasklaen, line 4112: We are already almost at 4.5k lines, we need to be conservative :)

```lua
set_state('has_chapter', #chapters > 0)
```

tomasklaen:
TY!

PR #212, opened 2022-09-17
feat: add `current/count` tooltip for list item elements

dyphire:


dyphire:
Optimize code implementation. Now the current chapter is saved as a state variable.

tomasklaen:
I don't see a value in telling people the index of currently selected sub/audio/... track. And I especially don't like this in a tooltip, as it's annoying to get to that info. And I further don't like that this is available only for built in shorthands. And even more how much code this not very valuable feature adds.

So I won't be merging this. I'm sorry you wasted your time, I wish you created an issue to discuss it first :(.

Though I do see the idea of displaying counters interesting. I'll try implementing it somehow.

dyphire:
In this pr, I want to add the list information prompt function that other OSC scripts have.
Example effect:

Personally, I would like to see this information, but It's true that I don't know how many users are interested in using this script.
> And even more how much code this not very valuable feature adds.
Though I do see the idea of displaying counters interesting. I'll try implementing it somehow.

The relevant implementation code is really not concise enough. I hope you can find a better implementation method next.

tomasklaen:
Control bar elements now support `#{badge}` syntax which will give button a badge. `{badge}` can be `sub`, `audio`, `video`, `playlist` for internally managed counters, or any mpv property that makes sense to have as a badge. If property is an array, it'll display its size.

Here's how the default control bar looks like now:

image

dyphire:
> Control bar elements now support `#{badge}` syntax which will give button a badge. `{badge}` can be `sub`, `audio`, `video`, `playlist` for internally managed counters, or any mpv property that makes sense to have as a badge. If property is an array, it'll display its size.

This implementation looks great, especially as a syntax option.

At the same time, I still hope there is a way to realize the prompt of current orbit/position. 
It doesn‘t necessarily need to be customizable. We just need to provide this function for the built-in tool controls. Maybe we can directly share the corresponding items of the submenu. This should not involve complex code.

dyphire:
In addition, can you also provide the function of custom icons for built-in tool controls? Now I can only modify the script to achieve this, which is not what I want.

dyphire:
https://github.com/tomasklaen/uosc/blob/08d4dd9e4d243100d80ba7785bc400ff90e3997a/scripts/uosc.lua#L933-L934
By the way, it seems that you forgot to change it.
It should be `opts.font = 'uosc_icons'`

christoph-heinrich:
> In addition, can you also provide the function of custom icons for built-in controls? Now I can only modify the script to achieve this, which is not what I want.

Can't you just take the definition of the shorthand and modify it according to your needs? e.g. `command:theaters:script-binding uosc/audio#audio?Audio` for audio button with video icon.

tomasklaen:
> current orbit/position

I won't be adding position to these badges. I don't see a value in it for stuff like subs, audio, etc.

The only one where that makes sense is the playlist count, but that position is already in the window title. Also, the badge looks pretty bad if there's more than 2-3 letters in it, which might be common for playlist counter even without the index in it. Yeah, I guess I'll just drop playlist counter altogether, and make the one in the title more pronounced instead.

> It should be `opts.font = 'uosc_icons'`

`uosc_icons` is just a file name, `MaterialIconsRound-Regular` is the font name/id that is added by that file.

tomasklaen:
I've removed the playlist counter from badges, and improved the top bar position instead:

image

dyphire:
> I've removed the playlist counter from badges, and improved the top bar position instead:
> 
> image

It looks good, but the potential audio/subtitle track counter problem is still unresolved.

tomasklaen:
> It looks good, but the potential audio/subtitle track counter problem is still unresolved.

It's not a problem. Why would you want to know the index of the active subtitle track? What value does it provide?

I'd be interested in language/format of current subtitles, or language/channels/bitrate for audio, all of which are too long to display in a badge or anywhere else in the UI for that matter, so you understandably have to pull out a menu for this.

But **index**? This is a completely useless piece of information to clutter the UI with.

dyphire:
> It's not a problem. Why would you want to know the index of the active subtitle track? What value does it provide?

You've misunderstood. I mean that:
> > the badge looks pretty bad if there's more than 2-3 letters in it
>
> This is also very common for audio/subtitles, especially web-dl video streaming.

tomasklaen:
I've never seen, heard of, or can think of a use case for a file with 4 digit number of audio or subtitle tracks 😨. How is this _common_ for you?

dyphire:
> I've never seen, heard of, or can think of a use case for a file with 3 digit number of audio or subtitle tracks 😨. How is this _common_ for you?

It is also rare for me to have a 3 digit number of audio or 
 subtitle track. I only know that the number of subtitles automatically generated by Youtube can be as large as this, which can be obtained through ytdl.

Two digit numbers of audio or subtitle tracks are common. It would be nice if this did not constitute a potential problem.

ps:I can personally accept this level of icon coverage.

tomasklaen:
More than 2 digits doesn't mean `>2`, it means `>99`.

```
10
└┴──> 2 digits
100
└┴┴─> 3 digits
```

maaldheefee:
Can we have the option to hide 0 (and maybe 1) values?

FichteFoll:
Would it make sense to dim buttons if their badge as well with the same or a similar condition as the badge not being shown?

tomasklaen:
The buttons are still track selecting buttons that can load the track if missing. Dimmed would imply they are disabled or something. Doesn't seem right to me.

dyphire:
At present, counters can be used to distinguish, and Dimmed seems unnecessary.

FichteFoll:
> […] that can load the track if missing.

Right, I forgot about this. Disregard my suggestion.

dyphire:
> (Not for video though, iirc.)

When will you need this? Is that add an external cover for the audio file?

PR #213, opened 2022-09-18
feat: thumbfast intergration

### LINKED ISSUE #7: Thumbnails in timeline
Display thumbnails when hovering the timeline, similar to https://github.com/TheAMM/mpv_thumbnail_script

po5:
Very minimal integration. Get the script here: https://github.com/po5/thumbfast

Does what you'd expect, generates and displays thumbnails. The client is in charge of positioning.  
Appears stacked on top of timestamp and chapter title if there is one.

I'd like to draw (on the uosc side) a background that acts as a border when covered up by the thumbnail, and as a placeholder while it's rendering if you don't see any issues with that, but you or @christoph-heinrich would know better how to make it look nice.  
Just keep in mind the image overlay is always rendered on top of osd overlays.

Closes #7.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -634,7 +634,7 @@ function wrap_text(text, target_line_length)
 		lines[#lines + 1] = string.sub(text, line_start)
 		if line_length > max_length then max_length = line_length end
 	end
-	return table.concat(lines, '\n'), max_length
+	return table.concat(lines, '\n'), max_length, #lines
 end
 
 -- Escape a string for verbatim display on the OSD
@@ -894,7 +894,7 @@ function serialize_chapters(chapters)
 	if not chapters then return end
 	for index, chapter in ipairs(chapters) do
 		chapter.index = index
-		chapter.title_wrapped, chapter.title_wrapped_width = wrap_text(chapter.title, 25)
+		chapter.title_wrapped, chapter.title_wrapped_width, chapter.title_wrapped_lines = wrap_text(chapter.title, 25)
 		chapter.title_wrapped = ass_escape(chapter.title_wrapped)
 	end
 	return chapters
@@ -2879,6 +2879,7 @@ function Timeline:render()
 			if chapter and not chapter.is_end_only then
 				chapter_title = chapter.title_wrapped
 				chapter_title_width = chapter.title_wrapped_width
+				chapter_title_lines = chapter.title_wrapped_lines
 			end
 		end
 
@@ -2902,10 +2903,12 @@ function Timeline:render()
 
 		-- Thumbnail
 		if not thumbfast.disabled and thumbfast.width ~= 0 and thumbfast.height ~= 0 then
+			local final_scale = options.ui_scale * state.hidpi_scale
 			mp.commandv("script-message-to", "thumbfast", "thumb",
 				hovered_seconds,
-				math.min(display.width - thumbfast.width - 10, math.max(10, cursor.x - thumbfast.width / 2)),
-				fay - thumbfast.height - (2 + self.font_size * (chapter_title and 2.8 or 1.4))
+				math.min(display.width * final_scale - thumbfast.width - 10 * final_scale, math.max(10 * final_scale, cursor.x * final_scale - thumbfast.width / 2)),
+				fay * final_scale - thumbfast.height - (2 + self.font_size * (chapter_title and 1.8 or 1.4) +
+					(chapter_title_lines and self.font_size * chapter_title_lines or 0)) * final_scale
 			)
 		end
 	else
@@ -4189,7 +4192,7 @@ mp.observe_property('osd-dimensions', 'native', function(name, val)
 	update_display_dimensions()
 	request_render()
 end)
-mp.observe_property('display-hidpi-scale', 'native', update_display_dimensions)
+mp.observe_property('display-hidpi-scale', 'native', create_state_setter('hidpi_scale', function() update_display_dimensions() end))
 mp.observe_property('demuxer-via-network', 'native', create_state_setter('is_stream', function()
 	set_state('uncached_ranges', state.is_stream and state.duration and {0, state.duration} or nil)
 	Elements:trigger('dispositions')
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -435,6 +435,11 @@ local state = {
 	margin_top = 0,
 	margin_bottom = 0,
 }
+local thumbfast = {
+	width = 0,
+	height = 0,
+	disabled = false
+}
 
 --[[ HELPERS ]]
 
@@ -2894,6 +2899,19 @@ function Timeline:render()
 				text_length_override = chapter_title_width,
 			})
 		end
+
+		-- Thumbnail
+		if not thumbfast.disabled and thumbfast.width ~= 0 and thumbfast.height ~= 0 then
+			mp.commandv("script-message-to", "thumbfast", "thumb",
+				hovered_seconds,
+				math.min(display.width - thumbfast.width - 10, math.max(10, cursor.x - thumbfast.width / 2)),
+				fay - thumbfast.height - (2 + self.font_size * (chapter_title and 2.8 or 1.4))
+			)
+		end
+	else
+		if thumbfast.width ~= 0 and thumbfast.height ~= 0 then
+			mp.commandv("script-message-to", "thumbfast", "clear")
+		end
 	end
 
 	return ass
@@ -4593,5 +4611,13 @@ mp.register_script_message('update-menu', function(json)
 		local menu = data.type and Menu:is_open(data.type)
 		if menu then menu:update(data)
 		else open_command_menu(data) end
+	end
+end)
+mp.register_script_message('thumbfast-info', function(json)
+	local data = utils.parse_json(json)
+	if type(data) ~= 'table' or not data.width or not data.height then
+		msg.error('thumbfast-info: received json didn\'t produce a table with thumbnail information')
+	else
+		thumbfast = data
 	end
 end)


po5:
Obligatory video.
https://user-images.githubusercontent.com/42466980/190887348-7c8320a2-2e67-4236-9d63-c6ba60769baa.mp4

dyphire:
This is great implementation, but it will be broken when used with `--hidpi-window-scale` option.

po5:
I don't know how to reproduce this, `--hidpi-window-scale` does nothing since my monitor is only 1080p.  
Best of luck to whoever tries fixing it.

christoph-heinrich:
I think the coordinate calculation simply has to be multiplied by `display-hidpi-scale`. Currently it's not saved as a state variable, so that would have to be done (add to state at the top and then set with `set_state()` in the `mp.observe_property('display-hidpi-scale',...)`

Natural-Harmonia-Gropius:
I think `width` and `height` can be remove, just control it in thumbfast is fine.
And thumbfast.disabled may be automactically controlled by something exports from thumbfast like this https://github.com/christoph-heinrich/mpv-quality-menu/blob/c56183646e40514d918d6eec89120a376ddb37b8/quality-menu.lua#L787

tomasklaen:
This looks great!

The only nitpicks I have is first, add a readme with instructions and config example. I'll add an installation step for thumbnails where I'll send people to your repo, so it'd be nice if they had all the necessary info without having to inspecting the code.

And second, maybe instead of `max_width` & `max_height`, which might not be ideal for portrait or ultrawide videos, have a one `resolution` option, that'd accept values like:

```config
resolution=300x200
resolution=300*200
resolution=60000
```

Where you'll have to split the first 2 examples and multiply the values to get the pixels, and then calculate width & height based on videos aspect ratio that fits this resolution.

po5:
> I think `width` and `height` can be remove, just control it in thumbfast is fine.

We can't always know where the client expects the thumbnail to be positioned, e.g. if it would rather restrict thumbnail movement farther from the edge of the window to match its seekbar, if the seekbar is at the top of the UI (x/y would have to turn into top-left coordinates requiring alignment options).  
It also would make resulting thumbnail position unpredictable to clients unless they copy the entire positioning code, discouraging decorations (and no the script will never draw its own).

> And `thumbfast.disabled` may be better to automactically controlled by something exports from thumbfast

The value does come from a script message, there is no way for a user of uosc to manipulate this variable through e.g. a config.  
It's meant to communicate states when thumbfast itself refuses to run (e.g. network video if not allowed in config) and no decorations should be drawn by the osc.

> readme with instructions and config example

Will definitely be doing that, wanted to make sure you didn't see anything fundamentally wrong with the way it currently works.

> And second, maybe instead of `max_width` & `max_height`, which might not be ideal for portrait or ultrawide videos, have a one `resolution` option

It's specifically because of wide aspect videos that it works like this. I don't want the client to be expected to handle a 2000x30 *thumbnail* that may break its math and go out of the screen, and it breaks user expectations.  
This is especially a concern since the script matches your current cropping, the user can get themselves in a situation with an oversized thumbnail.  
The option can always be added later if people care as this is independent from the script-message API, and all on the thumbfast side.

Will write a nice readme, attempt to fix the hidpi issue, and draw a background/border.

Mac users: I'd like help to get it to work, as I don't have a mac myself.

tomasklaen:
`max_width` and `max_height` could be used to handle the extremes, but this is just a tiny nitpick :) It's fine as it is now too, and definitely simpler to implement. No issues if you won't add `resolution`.

API and all looks fine to me, I have no objections. This is pretty much exactly how I envisioned it to work. All `width`, `height`, and `disabled` data coming from info message are definitely needed.

> and draw a background/border

Nooo, make thumbfast just draw the image, let me do the styling in uosc.

po5:
> Nooo, make thumbfast just draw the image, let me do the styling in uosc.

Yes, I that's what I meant as in the original post.

> I'd like to draw (on the uosc side) a background that acts as a border when covered up by the thumbnail, and as a placeholder while it's rendering

But I'll let you do the styling if you insist :^)

po5:
Nice readme is done.

christoph-heinrich:
This quick and dirty patch makes the thumbnail display at the right position with hidpi scaling.
``` diff
index dc4b9ac..f9298f3 100644
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -2956,8 +2956,8 @@ function Timeline:render()
 		if not thumbfast.disabled and thumbfast.width ~= 0 and thumbfast.height ~= 0 then
 			mp.commandv("script-message-to", "thumbfast", "thumb",
 				hovered_seconds,
-				math.min(display.width - thumbfast.width - 10, math.max(10, cursor.x - thumbfast.width / 2)),
-				fay - thumbfast.height - (2 + self.font_size * (chapter_title and 2.8 or 1.4))
+				math.min(display.width * state.hidpi_scale - thumbfast.width - 10, math.max(10, cursor.x * state.hidpi_scale - thumbfast.width / 2)),
+				fay * state.hidpi_scale - thumbfast.height - (2 + self.font_size * (chapter_title and 2.8 or 1.4) * state.hidpi_scale)
 			)
 		end
 	else
@@ -4158,7 +4158,7 @@ mp.observe_property('osd-dimensions', 'native', function(name, val)
 	update_display_dimensions()
 	request_render()
 end)
-mp.observe_property('display-hidpi-scale', 'native', update_display_dimensions)
+mp.observe_property('display-hidpi-scale', 'native', create_state_setter('hidpi_scale', function() update_display_dimensions() end))
 mp.observe_property('demuxer-cache-state', 'native', function(prop, cache_state)
 	local cached_ranges = cache_state and cache_state['seekable-ranges'] or {}
 	local uncached_ranges = nil
```

The current implementation doesn't take into account chapter titles with multiple lines. With this patch it does:
``` diff
index dc4b9ac..5e5e393 100644
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -713,7 +713,7 @@ function wrap_text(text, target_line_length)
 		lines[#lines + 1] = string.sub(text, line_start)
 		if line_length > max_length then max_length = line_length end
 	end
-	return table.concat(lines, '\n'), max_length
+	return table.concat(lines, '\n'), max_length, #lines
 end
 
 -- Escape a string for verbatim display on the OSD
@@ -920,7 +920,7 @@ function serialize_chapters()
 	end
 
 	for _, chapter in ipairs(chapters) do
-		chapter.title_wrapped, chapter.title_wrapped_width = wrap_text(chapter.title, 25)
+		chapter.title_wrapped, chapter.title_wrapped_width, chapter.title_wrapped_lines = wrap_text(chapter.title, 25)
 		chapter.title_wrapped = ass_escape(chapter.title_wrapped)
 	end
 
@@ -2919,7 +2919,7 @@ function Timeline:render()
 	-- Hovered time and chapter
 	if (self.proximity_raw == 0 or self.pressed) and not (Elements.speed and Elements.speed.dragging) then
 		local hovered_seconds = self:get_time_at_x(cursor.x)
-		local chapter_title, chapter_title_width = nil, nil
+		local chapter_title, chapter_title_width, chapter_title_lines = nil, nil
 
 		if (options.timeline_chapters ~= 'never' and state.chapters) then
 			for i = #state.chapters, 1, -1 do
@@ -2928,6 +2928,7 @@ function Timeline:render()
 					if not chapter.is_end_only then
 						chapter_title = chapter.title_wrapped
 						chapter_title_width = chapter.title_wrapped_width
+						chapter_title_lines = chapter.title_wrapped_lines
 					end
 					break
 				end
@@ -2957,7 +2958,8 @@ function Timeline:render()
 			mp.commandv("script-message-to", "thumbfast", "thumb",
 				hovered_seconds,
 				math.min(display.width - thumbfast.width - 10, math.max(10, cursor.x - thumbfast.width / 2)),
-				fay - thumbfast.height - (2 + self.font_size * (chapter_title and 2.8 or 1.4))
+				fay - thumbfast.height - (2 + self.font_size * (chapter_title and 1.8 or 1.4) +
+					(chapter_title_lines and self.font_size * chapter_title_lines or 0))
 			)
 		end
 	else
```

christoph-heinrich:
The current implementation doesn't take into account chapter titles with multiple lines. With this patch it does:
``` diff
index dc4b9ac..5e5e393 100644
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -713,7 +713,7 @@ function wrap_text(text, target_line_length)
 		lines[#lines + 1] = string.sub(text, line_start)
 		if line_length > max_length then max_length = line_length end
 	end
-	return table.concat(lines, '\n'), max_length
+	return table.concat(lines, '\n'), max_length, #lines
 end
 
 -- Escape a string for verbatim display on the OSD
@@ -920,7 +920,7 @@ function serialize_chapters()
 	end
 
 	for _, chapter in ipairs(chapters) do
-		chapter.title_wrapped, chapter.title_wrapped_width = wrap_text(chapter.title, 25)
+		chapter.title_wrapped, chapter.title_wrapped_width, chapter.title_wrapped_lines = wrap_text(chapter.title, 25)
 		chapter.title_wrapped = ass_escape(chapter.title_wrapped)
 	end
 
@@ -2919,7 +2919,7 @@ function Timeline:render()
 	-- Hovered time and chapter
 	if (self.proximity_raw == 0 or self.pressed) and not (Elements.speed and Elements.speed.dragging) then
 		local hovered_seconds = self:get_time_at_x(cursor.x)
-		local chapter_title, chapter_title_width = nil, nil
+		local chapter_title, chapter_title_width, chapter_title_lines = nil, nil
 
 		if (options.timeline_chapters ~= 'never' and state.chapters) then
 			for i = #state.chapters, 1, -1 do
@@ -2928,6 +2928,7 @@ function Timeline:render()
 					if not chapter.is_end_only then
 						chapter_title = chapter.title_wrapped
 						chapter_title_width = chapter.title_wrapped_width
+						chapter_title_lines = chapter.title_wrapped_lines
 					end
 					break
 				end
@@ -2957,7 +2958,8 @@ function Timeline:render()
 			mp.commandv("script-message-to", "thumbfast", "thumb",
 				hovered_seconds,
 				math.min(display.width - thumbfast.width - 10, math.max(10, cursor.x - thumbfast.width / 2)),
-				fay - thumbfast.height - (2 + self.font_size * (chapter_title and 2.8 or 1.4))
+				fay - thumbfast.height - (2 + self.font_size * (chapter_title and 1.8 or 1.4) +
+					(chapter_title_lines and self.font_size * chapter_title_lines or 0))
 			)
 		end
 	else
```

christoph-heinrich:
It's probably impractical with the way things are right now, but if we were to send the script messages in `Timeline:on_mouse_move()` and `Timeline:on_mouse_leave()`, depending on how the timing works out, the thumbnail could be rendered one frame earlier.

tomasklaen:
Should we be accounting for hidpi? That should be handled in thumbfast, no?

And yes, I was already planning to move the `thumbfast clear` message to `on_mouse_leave` hook after merging this, but `thumbfast thumb` with tooltip limits would be way too weird in `on_mouse_move`. It's fine where it is now.

christoph-heinrich:
I've realized in #192 that for really good hidpi support (avoiding blurry lines just like without scaling), the rendering has to be done at native resolution. Because of that I don't think thumbfast should scale it's rendering resolution. It would make sense to scale the thumbnail size with hidpi scaling, but positioning should be native.

po5:
I realized ui_scale caused the same issue as hidpi_scale, so I was finally able to test it (and also fixed handling of ui_scale).  
Would still like someone with a hidpi display to make sure it's fine in combination with both default and non-default ui_scale.

Implemented @christoph-heinrich's chapter title wrap calculation, I have mixed feeling about how it makes the thumbnail jump vertically between chapters with different line counts. But it's better than forcing it to be at the maximum height of *any* chapter, as a single very long chapter would throw the whole thing off.

When hidpi is confirmed to work, fine for merge?

tomasklaen:
> I have mixed feeling about how it makes the thumbnail jump vertically between chapters with different line counts

Thumbnail should definitely be below chapter title, precisely because of this issue, but I can deal with styling and stuff like this after the merge.

> fine for merge?

Just rebase the serialize_chapters conflict 🤐.

dyphire:
> Would still like someone with a hidpi display to make sure it's fine in combination with both default and non-default ui_scale.

For reference only, the pr can work normally with use `hidpi-window-scale` or `ui_scale=1.5` for me.

po5:
I'll let you deal with the fancy background stuff and chapter title positioning. chapter_title_lines stuff is probably useless now.

Zabooby:
Quick question, will thumbfast work from within uosc, meaning that I won't have to download any extra scripts?

christoph-heinrich:
> Quick question, will thumbfast work from within uosc, meaning that I won't have to download any extra scripts?

Quick answer, no.

omaru-ok:
> > Quick question, will thumbfast work from within uosc, meaning that I won't have to download any extra scripts?
> 
> Quick answer, no.

quack question., 
im still waitin for new realease :O

dyphire:
> Thumbnail should definitely be below chapter title, precisely because of this issue, but I can deal with styling and stuff like this after the merge.

Just out of curiosity, the chapter title and thumbnail cannot be displayed together at present. Is this because the reconstruction has not been completed yet?

tomasklaen:
Are you using latest main? There was an error whenre tooltip was trying to position itself down when too high up on the screen, which shouldn't happen for timeline chapters. but that has been fixed, and everything works fine for me:

Screenshot

dyphire:
> Are you using latest main? There was an error whenre tooltip was trying to position itself down when too high up on the screen, which shouldn't happen for timeline chapters. but that has been fixed, and everything works fine for me:

Yes, I'm using latest main, Somehow it didn't work for me.

dyphire:
I noticed that our thumbnail styles are also different, and I suddenly felt that this might be a `hidpi-window-scale` issue.

So I tried to disable the `hidpi-window-scale` option and it works.

So the issue is still `hidpi-window-scale`.

tomasklaen:
And your timeline timestamps are also not centered. Damn dealing with this hidpi thing is annoying.

dyphire:
> And your timeline timestamps are also not centered. Damn dealing with this hidpi thing is annoying.

After testing, the `ui_scale` option also caused issues.

When I try to disable `hidpi-window-scale` and use the `ui_scale=1.5`, the same results appear as `hidpi-window-scale`. 
But when I use the `hidpi-window-scale=no` and `ui_scale=1.5`, I have a more strange bug.

I have to say these strange bugs are maddening.

tomasklaen:
Btw, I'm working on new chapter ranges right now. I'm hoping since @christoph-heinrich added all of the hidpi code into uosc he takes over and deals with it 👼. If not, I'll look into it when I'm done.

christoph-heinrich:
I'll rework the whole scaling thing, but that will take some time. Don't expect anything today, and I don't know if I'll have enough time tomorrow.

christoph-heinrich:
Actually I'll make it work today, because I'm not so sure about the other approach to scaling. The result should be better, but it's also a lot more invasive to the code.

PR #214, opened 2022-09-18
fix: show fully uncached range for network streams

christoph-heinrich:
The timeline should show to be uncached as soon as can.

Not sure if the callback is really necessary. In my testing the `demuxer-via-network` observer always gets called once at the beginning of the file right before the `demuxer-cache-state` observer and right after the `duration` observer, so it shouldn't really make a difference. But I don't think that's guaranteed, so I've added the callback just in case.

The state already had a `is_stream` variable that wasn't used.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -4100,7 +4100,6 @@ mp.observe_property('duration', 'number', create_state_setter('duration', update
 mp.observe_property('speed', 'number', create_state_setter('speed', update_human_times))
 mp.observe_property('track-list', 'native', function(name, value)
 	-- checks the file dispositions
-	local path = mp.get_property_native('path')
 	local has_audio, has_sub, is_video, is_image = false, false, false, false
 	for _, track in ipairs(value) do
 		if track.type == 'audio' then has_audio = true end
@@ -4117,7 +4116,6 @@ mp.observe_property('track-list', 'native', function(name, value)
 	set_state('has_audio', has_audio)
 	set_state('has_sub', has_sub)
 	set_state('is_video', is_video)
-	set_state('is_stream', is_protocol(path))
 	Elements:trigger('dispositions')
 end)
 mp.observe_property('chapter-list', 'native', function(_, chapters)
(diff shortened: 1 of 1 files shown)


tomasklaen:
Why did you remove the conditionals in the loop that is inverting cached ranges? According to the docs, cached ranges can overlap or be very close to each other. That special handling was facilitating concatenation of overlapping or too close ranges, as well as ignoring the ones that are too short, and rendering them would be just a noise on the timeline.

tomasklaen:
Had to add another condition because I've noticed a redundant uncached range indicator at the end on some streams (last cached range end time was lower than reported duration by a ~100ms or so).

christoph-heinrich:
I've split up the change in two commits, as they are pretty much independent of each other.

Filtering the ranges afterwards is easier then trying to do it all at once (easy to miss edge cases). Performance is potentially maybe a little bit worse this way, but I don't think that it makes much of a difference either way.

tomasklaen:
Damn, mine was buggy 😭. Swhat I get for not testing it.

Do we still need the `set_state('is_stream', is_protocol(path))` in `track-list` observer when this is now set by `demuxer-via-network`? If not remove it and I'll merge.

tomasklaen:
Wait, the `track-list` is also responsible for triggering `dispositions` event. So where do we put it? Is `demuxer-via-network` available inside `track-list` prop event? Maybe we could just change the current set_state to:

```lua
set_state('is_stream', mp.get_property_native('demuxer-via-network'))
```

And do only the default assignemnt of uncached ranges inside `demuxer-via-network` one.

christoph-heinrich:
Yes, I totally missed how is_stream used to be set (because it doesn't show up by looking at references).
I've now already changed the disposition trigger. Is it fine like that?

tomasklaen:
`dispositions` event listeners are a bit expensive (controls bar destruction, serialization, re-creation), so I don't want it to trigger too much if not necessary. And it's already triggering like 3 times each file load.

I guess this is unavoidable, and it'd better be solved by event throttling.

christoph-heinrich:
Maybe we could serialize the control bar once at the beginning, and create an object for each item in the configuration, independent of if it's condition is met or not. Then on disposition trigger it only runs through each object checking if the condition is met or not. Objects who's condition is not met are not rendered and don't take up any space.

Issue #203, opened 2022-09-15
Maximize window button not working correctly


NaiveInvestigator:
If the fullscreen is used, using the other button fullscreen button or using `cycle fullscreen` doesn't minimize the window. 
![image](https://user-images.githubusercontent.com/72654596/190341813-7e86a0c4-a2f9-49de-96dd-e846a7f553dd.png)

tomasklaen:
Yes that's a known issue (can't find where we discussed this before though). All uosc does when you click the button is `cycle window-maximized` command, so it's on mpv to handle that correctly, and it currently doesn't.

This comment is the best explanation I could find as to what is happening: https://github.com/mpv-player/mpv/issues/7887#issuecomment-655529206

tomasklaen:
I think I'll give the button some logic to work around this. It's not acceptable for that button to be this broken.

Natural-Harmonia-Gropius:
There's another issue that `set fullscreen no; set window-maximized no` will not work properly when you maximize the window and then turn it to fullscreen.

Try this, although in the above scenario you need to click twice.
https://github.com/Natural-Harmonia-Gropius/uosc/blob/c184cc1c2956770827ddefddd84f2eca13ebf7f5/scripts/uosc.lua#L2959

hooke007:
Not sure about mpv's behavior when using `no-border` on Linux. I would never touch that button on Windows for one part of output will be covered by win-taskbar.

> I think I'll give the button some logic to work around this. 

How about this idea?

- if `border=no`
→→ button_win-max: `cycle fullscreen` (same as button_fullscreen)

- if `border=yes`
  - if `fullscreen=yes`
→→ button_win-max: `set fullscreen no` `cycle window-maximized`
  - if `fullscreen=no`
→→ button_win-max: `cycle window-maximized`

tomasklaen:
This should be the best what we can do in this situation. I personally couldn't find a state where it still does something bad.

It's pretty much the logic described by @hooke007 above, but I added `set window-maximized no;cycle fullscreen` to `border=no` state.

christoph-heinrich:
I think we need to make this behavior platform specific or turn the maximize button in effectively a fullscreen button, because it's horribly broken for me on kde plasma on wayland.
Related https://github.com/mpv-player/mpv/issues/11575

I never use that button, so I never noticed before.
![image](https://github.com/tomasklaen/uosc/assets/8932183/9d63f70b-815b-4c30-9386-7915dd4111bf)
[...]

PR #216, opened 2022-09-19
Fullscreen button fix

NaiveInvestigator:
This is more of a temporary patch than a proper fix  for #203, but I think it could work fine for the time being. I tested this and haven't come across any issues (yet). I hope this helps!


tomasklaen:
This can't just cycle fullscreen. If you're maximized, it won't do anything, so this just introduces different, although slightly less common bugs. I just need to get to fixing it.

PR #221, opened 2022-09-19
fix: anchor texture instead of align and shift

christoph-heinrich:
The idea behind this is that the texture tiles as if it started tiling at the anchor. This has the benefit that if multiple textured areas share the same anchor, their pattern lines up. That's noticeable with in the timeline, where if you seek forward into an uncached section, the pattern on the left would move a bit on master, but with this everything stays perfectly aligned.

It's a replacement for `align` because having both doesn't make sense. Anything that could be done with align can also be done with anchoring. Later realized the same is true for `shift`.

~The timeline is currently anchored to `fby`, but I think it looks better in motion when it's anchored to `fay`.
However while testing I noticed a problem that is also present on master with anchoring to the top of the timeline (`align=7` on master). For some reason the pattern starts further down then the coordinates given to `ass:txt()`. Bigger texture size leads to bigger gap.~
~(I've made the pattern and timeline size very big to make it easier to notice)~
timeline_texture_gap

~Do you know why that is and if there is something that can be done about that?~

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -1044,21 +1044,20 @@ end
 ---@param bx number
 ---@param by number
 ---@param char string Texture font character.
----@param opts {size?: number; color: string; align?: number; shift?: number; opacity?: number; clip?: string}
+---@param opts {size?: number; color: string; opacity?: number; clip?: string; anchor_x?: number, anchor_y?: number}
 function ass_mt:texture(ax, ay, bx, by, char, opts)
 	opts = opts or {}
-	local width, height, align = bx - ax, by - ay, opts.align or 1
+	local anchor_x, anchor_y = opts.anchor_x or ax, opts.anchor_y or ay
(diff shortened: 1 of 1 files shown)


christoph-heinrich:
I've now also tried changing the math to work with with `align=1` but anchoring to the top of the timeline still results in the same gap.

christoph-heinrich:
With centered glyphs it renders properly now. But I've edited the font with fontforge, so now I'll have to also edit the svgs and then figure out what's going on with that json file.

christoph-heinrich:
Realized that `shift` can also be replaced by the new `anchor_x`. And while I was at it, I added a commit for centering the black lines between the white ones in the timeline and the volume slider.

tomasklaen:
Looks good, but revert the alignment to `by`, instead of `ay`. When anchored to the top it looks like it's _sliding in_ from the outside, but the timeline and every element in it is designed to look like its _growing in_, so I want the texture to adhere to that too.

PR #223, opened 2022-09-20
feat: add multiple subtitles as filters in menu

MasouShizuka:
feat: back to parent dir when in uosc/open-file menu
fix: preselect correctly when back to the parent dir
fix: correct the chapter index in the chapter menu
feat:
multi-subtitles: select from current directory or mkv built-in subtitles and load them as filters
playlist-prev-keep-multi-subtitles-status: select the previous episode and update the subtitles of the corresponding video
playlist-next-keep-multi-subtitles-status: select the next episode and update the subtitles of the corresponding video

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -4025,7 +4025,7 @@ function open_file_navigation_menu(directory_path, handle_select, opts)
 		if path.is_directory then
 			--  Preselect directory we are coming from
 			if path.is_to_parent then
-				inheritable_options.selected_path = directory.path
+                inheritable_options.active_path = directory.path
 			end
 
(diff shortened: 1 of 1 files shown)


dyphire:
> feat: back to parent dir when in uosc/open-file menu 
fix: preselect correctly when back to the parent dir 
fix: correct the chapter index in the chapter menu 

These commits seem to have nothing to do with this PR topic. 
Maybe it is better to split them into a new PR.

MasouShizuka:
Sorry, I am not familiar with github.
Actually, the 4th commit is related to the topic, and the other 3 are other parts that I have modified locally.

PR #224, opened 2022-09-20
fix: add .vob and .lrc in *_types

dyphire:
I am not sure whether to add `.iso` to media types.
What do you think?


po5:
iso files can play decently well, but only with the dvd:// prefix.  
It's largely undesirable to open an iso without that prefix, but do we want to do that for the user when their mpv may not be compiled with dvd support?  
Same thing with unpacked vob files (though those may actually be standalone vobs and not part of a dvd structure).

dyphire:
> Same thing with unpacked vob files (though those may actually be standalone vobs and not part of a dvd structure).

The vob files is the same as the properties of the m2ts files. Since .m2ts have been added in previous commit, it is reasonable to support vob files.

> iso files can play decently well, but only with the dvd:// prefix.

Or maybe bd:// prefix.

> It's largely undesirable to open an iso without that prefix.

I agree with this statement, so we all agree not to add .iso to media types.

dyphire:
> Just don't add any files that will cause issues when opened with file opening menu. Not sure which ones that are, so you have to duke it out among yourself :)

These won't. It is the iso file that may cause issues, so I decided not to add it.

PR #225, opened 2022-09-20
feat: back to parent dir when in uosc/open-file menu

MasouShizuka:


tomasklaen:
So if I open a file open menu and I press back, it's going to navigate up the directory tree? This makes no sense. The implementation is also very duck-tape-y. Why should menu care what the items' hints are, nevertheless rely on them for back button behavior?

And you can't activate current directory when navigating up the directory tree. The directory is not the active file, the file within it is. Just pre-selecting it was correct before.

Well, there's nothing here I'd like to merge. Sorry.

MasouShizuka:
I just want to navigate up the directory tree more easily, just click `left` or `h`.
And I know it should't rely on the hint word, but actually I don't know much about lua language.

If possible, I hope you can implement a function that navigate up the directory tree and select it.

PR #226, opened 2022-09-20
fix: correct the chapter index in the chapter menu

MasouShizuka:


tomasklaen:
What is wrong with the current code? It seems to do the same thing.

And I don't want to select 1st chapter if nothing was found. It's just not correct. If current time is lower than first chapter, than first chapter is just not active yet.

MasouShizuka:
If you choose chapter 1, then mpv will jump to the position of chapter 1.
And reopen the chapter menu, it will be the chapter 2.

I think the menu should reflect your current state, like what sub you choose now.
After jumping the chapter 1, it's between chapter 1 and chapter 2, and I think it's now belonging to chapter 1, and maybe you want to jump to chapter 2.

MasouShizuka:
Of course, about the first chapter, I agree with you.
`If current time is lower than first chapter, than first chapter is just not active yet.`

I do this just for convenience, no need click down to select.

tomasklaen:
I see the bug now. So..

1. Remove the `active_found` variable, it's not needed anymore.
2. Remove the pre-activation of 1st item. It's incorrect, and not necessary. If nothing is activated, menu pre-selects the first item automatically.

PR #227, opened 2022-09-20
feat: add multi-subtitles as filters

MasouShizuka:
feat:
multi-subtitles: select from current directory or mkv built-in subtitles and load them as filters
playlist-prev-keep-multi-subtitles-status: select the previous episode and update the subtitles of the corresponding video
playlist-next-keep-multi-subtitles-status: select the next episode and update the subtitles of the corresponding video

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -4578,3 +4578,126 @@ mp.register_script_message('update-menu', function(json)
 		else open_command_menu(data) end
 	end
 end)
+
+local multi_subtitles_info = {}
+
+function multi_subtitles_clear()
+    for index, value in ipairs(multi_subtitles_info) do
+        mp.command(value.remove)
+
+    multi_subtitles_info = {}
+    end
+end
+
+function multi_subtitles_exist(id)
+    for index, value in ipairs(multi_subtitles_info) do
+        if value.id == id then
+            return true, index
+        end
+    end
+
+    return false, -1
+end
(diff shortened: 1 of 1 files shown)


tomasklaen:
What is this doing? 😨 From reading it for like 5 minutes I think it's trying to select same subtitles when navigating to the next file? But it's doing something with filters, and files? and ... I'm not sure.

MasouShizuka:
For example, the video directory is like:
```
video 01.mkv
video 01_sc.ass
video 01_tc.ass
video 02.mkv
video 02_sc.ass
video 02_tc.ass
```

When playing the `video 01.mkv`, you can load `video 01_sc.ass` and load `video 01_tc.ass` as filter using `multi-subtitles` menu.
And by `playlist-next-keep-multi-subtitles-status`, `video 02.mkv` will load `video 02_tc.ass` as filter.

po5:
Is this a way to load a secondary subtitle but keep its styling? This would be more suited as its own script.

MasouShizuka:
If you use filter to load sub, it will keep the style.

hooke007:
There is no doubt it should be standalone.

tomasklaen:
Yeah you should make this as your own script. You can use uosc's [`open-menu` message API](https://github.com/tomasklaen/uosc#open-menu-menu_json-submenu_id) if you want to use uosc's menus.

PR #228, opened 2022-09-20
Chapter on window title

RexAckermann:
Show chapter details on title bar if available


tomasklaen:
I don't don't think chapter belongs there, nor would I personally ever want it there. The implementation is also incredibly wasteful, but that's irrelevant.

I think I have a better idea. Gonna try something out.

tomasklaen:
Here's how it looks like:

image

PR #238, opened 2022-09-21
fix: thumbnail decoration with scaling

christoph-heinrich:
Fixes the scaling as noted in https://github.com/tomasklaen/uosc/pull/213#issuecomment-1253839467

Screenshot is with 1.6 scaling factor.

timeline_thumbnail_scaling

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -2912,8 +2912,8 @@ function Timeline:render()
 				display.width * scale - thumbnail.width - thumb_x_margin
 			))
 			local thumb_y = round(tooltip_anchor.ay * scale - thumb_y_margin - thumbnail.height)
-			local ax, ay = thumb_x - border, thumb_y - border
-			local bx, by = thumb_x + border + thumbnail.width, thumb_y + border + thumbnail.height
+			local ax, ay = (thumb_x - border) / scale, (thumb_y - border) / scale
+			local bx, by = (thumb_x + border + thumbnail.width) / scale, (thumb_y + border + thumbnail.height) / scale
 			ass:rect(ax, ay, bx, by, {
 				color = options.foreground, border = 1, border_color = options.background, radius = 3, opacity = 0.6,
 			})


hooke007:
@tomasklaen  @christoph-heinrich  Sometimes the preview output is too close to the left and top side.

christoph-heinrich:
Another thing to have a look at ._.

christoph-heinrich:
So... I did find a bug in my implementation, but it's a pretty minor thing and didn't change anything how things look on my end.
I was able to reproduce the bug, but the difference in border size if far less then what's happening for you. What scaling factor are you using?

christoph-heinrich:
So... I have something now that seems to work, but I don't really understand why yet. That whole mixing real coordinates with scaled ones gets really confusing.
``` diff
index 3aa55a9..d73f006 100644
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -2918,13 +2918,13 @@ function Timeline:render()
 		-- Thumbnail
 		if not thumbnail.disabled and thumbnail.width ~= 0 and thumbnail.height ~= 0 then
 			local scale = options.ui_scale * state.hidpi_scale
-			local border, margin_x, margin_y = 2, 10, 5
-			local thumb_x_margin, thumb_y_margin = (border + margin_x) * scale, (border + margin_y) * scale
-			local thumb_x = round(clamp(
+			local border, margin_x, margin_y = round(2 * scale), round(10 * scale), round(5 * scale)
+			local thumb_x_margin, thumb_y_margin = border + margin_x, border + margin_y
+			local thumb_x = clamp(
 				thumb_x_margin, cursor.x * scale - thumbnail.width / 2,
 				display.width * scale - thumbnail.width - thumb_x_margin
-			))
-			local thumb_y = round(tooltip_anchor.ay * scale - thumb_y_margin - thumbnail.height)
+			)
+			local thumb_y = tooltip_anchor.ay * scale - thumb_y_margin - thumbnail.height
 			local ax, ay = (thumb_x - border) / scale, (thumb_y - border) / scale
 			local bx, by = (thumb_x + border + thumbnail.width) / scale, (thumb_y + border + thumbnail.height) / scale
 			ass:rect(ax, ay, bx, by, {
```

hooke007:
Snipaste_2022-09-22_17-43-29

It is Closer...

christoph-heinrich:
wtf is going on with that white background around your time and chapter title, there is no way my patch introduced that. If you use a ui_scale of 0,8333333333333333 (1/1.2), does it render correctly?

Honestly at this point I have no Idea why this could possibly be going wrong. Seems like a bug of libass to me, but then why does the rest of the UI look fine? one way to find out for sure would be to render everything in native resolution, but that would require a ton of work.

tomasklaen:
It seems like @hooke007 is using subtitle override settings (#241) and that's the source of the white backgrounds. Maybe he's doing also something that messes up with positioning too? Because I've noticed that timestamps in timeline are note vertically centered on all of his screenshots.

hooke007:
> wtf is going on with that white background around your time and chapter title

https://github.com/tomasklaen/uosc/issues/241

hooke007:
ui_scale=1/1.2

Snipaste_2022-09-22_18-16-02

tomasklaen:
And how is the current time tooltip and chapter title in the right position, but your timeline timestamps are not centered? Oh maybe it's just you consistently clipping the bottom of the timeline out of every screenshot at around the same place :)

hooke007:
Left side of timeline's preview

Snipaste_2022-09-22_18-21-27

christoph-heinrich:
This has to be a problem in libass... I've now tried rendering the rectangle without any border or radius. With scaling=1 I get clean edges, as one would expect. But with scaling=1.3 I have blurry edges, despite the real coordinates being integers. And when I round the coordinates to have integers in scaled coordinates it's still blurry. Scaling 1.35 creates clean edges again. Shit makes no sense.

tomasklaen:
Honestly, atm I'd rather have blurry UI for non-integer scaling than have to scale every size variable in the file. Especially since increasing scale is usually done for displays with higher dpi, where the blurriness might not even be noticeable.

But I might change my mind if I find time to wrap ma head around this problem, because atm the implementation of hidpi and scaling you added to uosc is transparent to me :)

christoph-heinrich:
@hooke007 At this point the only advice I have for you is to play with the `ui_scale` value until you have something that's about the right size and also looks alright. For me 1.3 doesn't render quite right, 1.35 looks great, and 1.4 is a bit blurry again. Don't know why, the numbers we hand to libass all look right. Maybe I'll have some epiphany after sleeping it over.

Since it looks worse for you then anything I can replicate, maybe try a minimal configuration with only uosc and thumbfast and see if it's better. I don't know what could be messing things up, but I'm out of other ideas.

hooke007:
Actually I found that I could use fullscreen to reduce these strange pixel shift.

Issue #239, opened 2022-09-21
The timeline's cache indication should not only be shown when playing network stream


Hill-98:
uosc will currently only display the cache indication when playing network stream.

But there are some files that also come from the network, such as files on a NAS, but mpv doesn't set `demuxer-via-network` to `yes`. so we need a more refined detection method.

I noticed that if `cache` is set to `auto`, mpv will automatically enable caching when playing files from the network drive, and local files are not cached, I'm not sure how mpv does this.

Can we change it to show the cache indication whenever the mpv has cached data? (no matter where the file came from)

PR #240, opened 2022-09-21
fix: show cache indication whenever cache is available

Hill-98:
#239

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -4172,7 +4172,7 @@ end))
 mp.observe_property('demuxer-cache-state', 'native', function(prop, cache_state)
 	local cached_ranges = cache_state and cache_state['seekable-ranges'] or {}
 	local uncached_ranges = nil
-	if state.duration and state.is_stream then
+	if state.duration and #cached_ranges > 0 then
 		-- Normalize
 		local ranges = {}
 		for _, range in ipairs(cached_ranges) do


scripts/uosc.lua
```diff
@@ -4172,15 +4172,18 @@ end))
 mp.observe_property('demuxer-cache-state', 'native', function(prop, cache_state)
 	local cached_ranges = cache_state and cache_state['seekable-ranges'] or {}
 	local uncached_ranges = nil
-	if state.duration and state.is_stream then
+	if state.duration then
```
christoph-heinrich, line 4175: Wouldn't it be simpler to just check for `#cached_ranges > 0` here instead of counting in the for?

christoph-heinrich:
This brings up the question how `state.is_stream` should be handled. arguably it should always be true when there is a cache or not?

scripts/uosc.lua
Hill-98: I'm not a Lua programmer, I found this method when I searched the web how to get the table size, I'll change it when I wake up.

Hill-98:
> This brings up the question how `state.is_stream` should be handled. Arguably it should always be true when there is a cache.

I'm used to setting `--cache` to `yes` so that it will be cached no matter what, but it doesn't have to be a stream.

tomasklaen:
Add:

```lua
if not state.is_stream then
	set_state('is_stream', true)
end
```

at the end of the `if state.duration and #cached_ranges > 0 then` condition.

Hill-98:
As I said above, there is not necessarily a 'stream' when there is a cache.

So I feel like either get rid of `is_stream`, or keep the status quo, and there are not particularly many features currently relying on `is_stream`.

christoph-heinrich:
This now overwrites the full cached range set by the `demuxer-via-network` observer for early display of uncached range ._.
That's why I suggested adding the `#cached_ranges > 0` check before the other if :neutral_face:

tomasklaen:
Oh I misunderstood your request. I saw `#cached_ranges > 0` in the updated PR and thought that was it and merged 🤷‍♂️. I'll fix it.

PR #251, opened 2022-09-23
fix: integer resolution when scaling

christoph-heinrich:
The resolution passed to ass was not necessarily an integer while scaling, and I think that's what's caused the issues noted in #238. In my testing the thumbnail decoration renders nicely now with all sorts of fractional scaling factors.

@hooke007 Can you please test if this resolves your issues?

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -377,7 +377,7 @@ end
 
 --[[ STATE ]]
 
-local display = {width = 1280, height = 720, aspect = 1.77778}
+local display = {width = 1280, height = 720, scale_x = 1, scale_y = 1}
 local cursor = {hidden = true, x = 0, y = 0}
 local state = {
 	os = (function()
@@ -428,6 +428,7 @@ local state = {
 	playlist_pos = 0,
 	margin_top = 0,
 	margin_bottom = 0,
+	hidpi_scale = 1,
 }
 local thumbnail = {width = 0, height = 0, disabled = false}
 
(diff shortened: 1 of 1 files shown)


hooke007:
Snipaste_2022-09-23_14-36-37

hooke007:
hidpi: set 1.2 in system settings (But it was strange that mpv report it was 1.198)

current-window-scale : 0.789
Snipaste_2022-09-23_14-56-42

current-window-scale : 0.627
Snipaste_2022-09-23_15-00-18

current-window-scale : 0.754
Snipaste_2022-09-23_15-01-51

tomasklaen:
Can I merge or are you still working on this?

christoph-heinrich:
Still working on it. hooke's ability to find problems is truly astonishing.

christoph-heinrich:
If you're looking for something to do, I noticed that playlist items now advance by 2 (so one gets skipped). Haven't had time to look into it or make a proper issue about it (with video).

tomasklaen:
How are you advancing? Both letting mpv open the next item or doing `uosc/next` advances correctly to the next item.

christoph-heinrich:
Simply letting it run out. I've noticed it happens only when I have speed >= 2. And I've also had it not happen once while speed was == 2.

https://user-images.githubusercontent.com/8932183/191990907-c18957e9-fe32-4f32-aff0-43ab1138d8c2.mp4

tomasklaen:
Double jump, along with 2 other bugs this made me notice is fixed now 😭.

christoph-heinrich:
Just noticed that depending on window size, despite border being a rectangle that stratches over the whole width, which is the same width that ass gets, it gets deteched from the window edge.
Screenshot_20220923_175238
Looks like libass really doesn't like rendering with scaled resolution. Arguably that's a bug in libass.

christoph-heinrich:
After trying absolutely everything regarding scaling and rounding that I could think of, this is the best we can do. It's not great, far from perfect, but I hope it's good enough.

`math.ceil()` on the border because bigger sizes make the problem less noticeable, but it certainly still is noticeable unfortunately.

hooke007:
Excellent now!

christoph-heinrich:
Still excellent after the last change? because that got pushed only seconds before your comment, so not sure if you missed it or not.

hooke007:
> Still excellent after the last change?

Yes

tomasklaen:
Awesome!

PR #256, opened 2022-09-23
fix: possible nil error in uncached ranges generation

christoph-heinrich:
Managed to get a nil error there, but it happens rarely.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -4304,7 +4304,12 @@ mp.observe_property('demuxer-via-network', 'native', create_state_setter('is_str
 	Elements:trigger('dispositions')
 end))
 mp.observe_property('demuxer-cache-state', 'native', function(prop, cache_state)
-	local cached_ranges = cache_state and cache_state['seekable-ranges'] or {}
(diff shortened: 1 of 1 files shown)


scripts/uosc.lua
```diff
@@ -4322,8 +4322,8 @@ mp.observe_property('demuxer-cache-state', 'native', function(prop, cache_state)
 		}
 	end
 	table.sort(ranges, function(a, b) return a[1] < b[1] end)
-	if cache_state['bof-cached'] then ranges[1][1] = 0 end
-	if cache_state['eof-cached'] then ranges[#ranges][2] = state.duration end
+	if cache_state and cache_state['bof-cached'] then ranges[1][1] = 0 end
+	if cache_state and cache_state['eof-cached'] then ranges[#ranges][2] = state.duration end
```
tomasklaen, line 4326: Wrap it in a single `if cache_state then`.

christoph-heinrich: At the beginning of the function is also a nil check for cache_state. I could initialize everthing at the beginning in a single if.

christoph-heinrich: Is this good?

PR #262, opened 2022-09-24
feat: support single chapter ads

christoph-heinrich:
So far only ad ranges from the sponsorblock script were supported, but when downloading with ytdl the sponsored parts are only a single chapter starting with `[SponsorBlock]:`. Also added `^sponsors?` because that's what I had in my old config and it never had any false positives for me.

ref #261

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -892,7 +892,7 @@ function serialize_chapter_ranges(normalized_chapters)
 		-- Sponsor blocks
 		if config.chapter_ranges.ads then
 			local id = chapter.lowercase_title:match('segment start *%(([%w]%w-)%)')
-			if id then
+			if id then -- ad range from sponsorblock
 				for j = i + 1, #chapters, 1 do
 					local end_chapter = chapters[j]
 					local end_match = end_chapter.lowercase_title:match('segment end *%(' .. id .. '%)')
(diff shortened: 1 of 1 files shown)


scripts/uosc.lua
```diff
@@ -907,6 +907,16 @@ function serialize_chapter_ranges(normalized_chapters)
 					end
 				end
 			end
+			-- single chapter for ad
+			if chapter.lowercase_title:find('%[sponsorblock%]:') or chapter.lowercase_title:find('^sponsors?') then
```
tomasklaen, line 911: This should be `elseif`.

Also, `[sponsorblock]` is just what this particular youtube creator decided to call his sponsor timestamp, so it's not going to match sponsor segment of anyone else. I think these 2 matches should be consolidated into `:find('sponsor')`, otherwise we won't match much, and might as well just not include this extra matching at all.

And how common would it be that someone puts a `sponsor` word into their chapter and it's not actually a sponsor segment?

christoph-heinrich: > This should be `elseif`.

oops, it used to be elseif, but then I tried something different and it never became elseif again.

> Also, `[sponsorblock]` is just what this particular youtube creator decided to call his sponsor timestamp, so it's not going to match sponsor segment of anyone else.

For testing I've downloaded a different video from a different creator and it still was `[sponsorblock]:`, so I think that's generally the case for downloaded videos with `yt-dlp --sponsorblock-mark "all"`.

> And how common would it be that someone puts a `sponsor` word into their chapter and it's not actually a sponsor segment?

I remember seeing some, enough to add that to my range config. I'll hunt down one for you :wink:

tomasklaen: Wait, the `[sponsorblock]:` thing is inserted by an alternative sponsorblock plugin built into yt-dlp?

If so, just change the `elseif`, and keep the your original 2 matches.

christoph-heinrich: The [WAN Show](https://www.youtube.com/watch?v=dOyrx5HOCyY) always has chapters with `Sponsor - Companyname`

PR #265, opened 2022-09-24
fix: typo

hooke007:


hooke007:
conf's value is diff from script. I am not sure if it was typo.

hooke007:
The previous url links to the `filled` , as far as I know different Variant has diff num of available icons.

PR #267, opened 2022-09-25
fix: render when thumbnail size changes

christoph-heinrich:
With recent discoveries in https://github.com/po5/thumbfast/pull/31 this is necessary because the thumbnail size may change at any time, but it seems like a good thing to do in general.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -4709,5 +4709,6 @@ mp.register_script_message('thumbfast-info', function(json)
 		msg.error('thumbfast-info: received json didn\'t produce a table with thumbnail information')
 	else
 		thumbnail = data
+		request_render()
 	end
 end)


Natural-Harmonia-Gropius:
It will already be adjusted.
[EXPAND.webm](https://user-images.githubusercontent.com/50797982/192249350-ccdcfc65-e430-40e5-9390-46d21048ae65.webm)

christoph-heinrich:
That file is great, thanks.
It is possible to run intro situations with that where no new thumbnail gets generated until you seek.
Interestingly the thumbnail resolution does not change as the underlying video resolution changes, which I would have expected to happen. Don't understand why tbh, because usually ffmpeg is rather eager in changing the output resolution in order to get closer to the original aspect ratio.
It is also possible with that file to seek to a position where the generated thumbnail differs in both height and width, which my function for guessing the real resolution can't handle. I guess I'll have to implement a full blown number factoring algorithm then.

tomasklaen:
Is the condition necessary? Why not just `request_render()` whenever message arrives? It's a throttled function after all.

christoph-heinrich:
There, condition removed.

PR #270, opened 2022-09-26
fix: new progress indicator line behavior

christoph-heinrich:
The math is much simpler then it used to be and there are some more simplifications I can do, but that will have to wait until tomorrow.
Mouse input works correctly but the code is horrible. Again, tomorrow.

I don't really like how the progress bar behaves when seeking, but this is technically correct behavior.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -2741,11 +2741,15 @@ function Timeline:update_dimensions()
 end
 
 function Timeline:get_time_at_x(x)
-	-- line width 1 for timeline_style=bar so mouse input can go all the way from 0 to 1 progress
-	local line_width = (options.timeline_style == 'line' and self:get_effective_line_width() or 1)
-	local time_width = self.width - line_width
-	local progress_x = x - self.ax - line_width / 2
-	local progress = clamp(0, progress_x / time_width, 1)
+	local line_width = (options.timeline_style == 'line' and self:get_effective_line_width() - 1 or 0)
+	local time_width = self.width - line_width - 1
+	local fax = (time_width) * state.time / state.duration
+	local fbx = fax + line_width
+	-- time starts 0.5 pixels in
+	x = x - self.ax - 0.5
+	if x > fbx then x = x - line_width
+	elseif x > fax then x = fax end
+	local progress = clamp(0, x / time_width, 1)
 	return state.duration * progress
 end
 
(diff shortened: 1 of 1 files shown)


christoph-heinrich:
Cleaned up the code and behavior is now also pixel perfect, both in line and bar style.

tomasklaen:
Try to make an always wide line, pause a video, pick a pixel somewhere in the last 20% of the timeline, and start clicking the same pixel. The video keeps seeking slightly forward until the cursor is at the start of the line. Can we do something about that? Because this is just weird.

And that chapter diamond stretching looks horrible, like something's broken. Call `lxa()` only on the `chapter_x`, not the individual diamond coordinates.

christoph-heinrich:
> Try to make an always wide line, pause a video, pick a pixel somewhere in the last 20% of the timeline, and start clicking the same pixel. The video keeps seeking slightly forward until the cursor is at the start of the line. Can we do something about that? Because this is just weird.

Only happens to me in windows mode, but not in fullscreen... I'm sure there is something that can be done about that, I'll have to do some more investigating. 

> And that chapter diamond stretching looks horrible, like something's broken. Call `lxa()` only on the `chapter_x`, not the individual diamond coordinates.

 Done.

christoph-heinrich:
The time shifting when seeking thing has also been resolved now.

tomasklaen:
The past (left to the line) x values adjustment seems to be off. Both ranges and chapters don't jump to the start of the line as they should, but a couple pixels in front of it:

[2022-09-27-20-35-35.webm](https://user-images.githubusercontent.com/47283320/192608498-ac8d4666-99c5-47f2-92d3-d8f105926eda.webm)

christoph-heinrich:
Wtf, I specifically tested for that at some point. I wonder how that got broken again.

christoph-heinrich:
Should be fine now. That's what I get for testing it in fullscreen.

tomasklaen:
Chapter ranges are broken now. I have this shortcut that inserts a couple testing chapters:

```lua
mp.add_forced_key_binding('Alt+Ctrl+v', 'foo', function()
    local duration = mp.get_property_native("duration")
    local new_chapters = {}
    new_chapters[#new_chapters + 1] = { title = 'Opening', time = duration * 0.03 }
    new_chapters[#new_chapters + 1] = { title = 'Chapter 2', time = duration * 0.1 }
    new_chapters[#new_chapters + 1] = { title = 'Chapter 3', time = duration * 0.55 }
    new_chapters[#new_chapters + 1] = { title = 'Sponsor segment start (afr54w)', time = duration * 0.8 }
    new_chapters[#new_chapters + 1] = { title = 'Sponsor segment end (afr54w)', time = duration * 0.9 }
    mp.set_property_native("chapter-list", new_chapters)
end)
```

And the ranges are in the wrong places now.

christoph-heinrich:
That now brings up the question how ranges that are adjacent to the beginning and the end should be handled, because right now they get forced to 0 or bbx if they are sufficiently close to it, but that creates a situation where a range might be over the progress bar, while the associated chapter is still in front of it.
Screenshot_20220927_220013

christoph-heinrich:
Now a chapter at 0 is always in front of the progress line, I think it's fine that way.

tomasklaen:
Remove that half a second rounding completely. I don't think it provides much value.

christoph-heinrich:
Half a pixel rounding provides the value of chapters at 0 being the same distance from the window edge as chapters at `state.duration` and it's also required for things to line up with mouse input.

PR #271, opened 2022-09-27
feat: add is_idle dispositions

dyphire:
This is helpful for controls visibility in mpv [pseudo gui mode](https://mpv.io/manual/master/#pseudo-gui-mode).

### DIFF
--- a/script-opts/uosc.conf
+++ b/script-opts/uosc.conf
@@ -51,6 +51,7 @@ timeline_chapters_opacity=0.8
 # Item visibility control:
 #   `<[!]{disposition1}[,[!]{dispositionN}]>` - optional prefix to control element's visibility
 #   - `{disposition}` can be one of:
+#     - `idle` - true if mpv is in idle mode (no file loaded)
 #     - `image` - true if current file is a single image
 #     - `audio` - true for audio only files
 #     - `video` - true for files with a video track
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -416,6 +416,7 @@ local state = {
 	volume = nil,
 	volume_max = nil,
 	mute = nil,
+	is_idle = false,
 	is_video = false,
 	is_audio = false, -- true if file is audio only (mp3, etc)
 	is_image = false,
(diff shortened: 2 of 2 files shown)


script-opts/uosc.conf
```diff
@@ -51,6 +51,7 @@ timeline_chapters_opacity=0.8
 # Item visibility control:
 #   `<[!]{disposition1}[,[!]{dispositionN}]>` - optional prefix to control element's visibility
 #   - `{disposition}` can be one of:
+#     - `idle` - true for if mpv is idle state
```
tomasklaen, line 54: Change to: "true if mpv is in idle mode (no file loaded)"

scripts/uosc.lua
```diff
@@ -4280,7 +4281,11 @@ mp.observe_property('playlist-count', 'number', function(_, value)
 end)
 mp.observe_property('fullscreen', 'bool', create_state_setter('fullscreen', update_fullormaxed))
 mp.observe_property('window-maximized', 'bool', create_state_setter('maximized', update_fullormaxed))
-mp.observe_property('idle-active', 'bool', create_state_setter('idle'))
+mp.observe_property('idle-active', 'bool', function(_, idle)
+	set_state('is_idle', idle)
+	create_state_setter('idle')
```
tomasklaen, line 4286: Remove the `create_state_setter('idle')` call. It doesn't do anything here.

tomasklaen (changes requested):

scripts/uosc.lua
christoph-heinrich: Could simply replace `'idle'` with `'is_idle'` in the original code and that should be fine or not?

tomasklaen: Depends on the order in which these events are being dispatched. If `idle-active` is last after everything else that triggers `dispositions` than this won't be reliable. I guess better be safe.

christoph-heinrich: Oh right, I missed the disposition because I only looked at that excerpt.

dyphire: > Remove the `create_state_setter('idle')` call. It doesn't do anything here.

Done.

PR #274, opened 2022-09-27
feat: make `*_persistency` support idle

hooke007:
Actually I think making those default value be `idle,paused` would be more reasonable for we do not had the logo/msg showing in idle. But I want to hear more opinions.


christoph-heinrich:
I think we would even be allowed to show the logo while idle. I've been wondering about adding that.

PR #276, opened 2022-09-29
fix: buggy active audio-device detection in audio-device menu

dyphire:


scripts/uosc.lua
```diff
@@ -4662,10 +4661,10 @@ mp.add_key_binding(nil, 'audio-device', create_self_updating_menu_opener({
 					hint = hint,
 					value = device.name,
 				}
-				if device.name == current_device then active_index = #items end
+				if device.name == current_device then items[#items].active = true end
```
tomasklaen, line 4664: Move the condition into the table above as `active = device.name == current_device` property.

tomasklaen (changes requested):

scripts/uosc.lua
dyphire: Done.

Issue #20, opened 2020-05-11
Add menu for editions


FichteFoll:
Editions are a seldomly used feature in Matroska files, since they require ordered chapters – an almost unknown magic –, but they exist.

Definitely not a priority for me, as I can easily cycle editions with a native mpv binding (which happens *very* rarely), but for the sake of completeness I thought I should mention this.

darsain:
First time seeing these editions. Looking at mpv docs and still don't really understand what they are.

Is it like sets of different chapters for same video file that you can switch between?

FichteFoll:
That's a good summary. An edition is a list of chapters, or more precisely: ordered chapters, which means a second edition could for example provide a different order for the same chapters without needing a separate video file. They an even reference chapters from other files (in the same folder) through ordered chapters.

https://mod16.org/hurfdurf/?p=8 provides the best overview on this that I know of.

darsain:
K. I'll look into this when more people request it, since it looks like none is using it atm, or when I'll have absolutely nothing else to do.

Todo when adding support for this:
- stop ensuring chapters are ordered chronologically by sorting them, and sort them only for purpose of generating chapter ranges

FichteFoll:
I don't how how mpv provides ordered chapters internally. It may already translate the ordered chapters into a virtual chapter stream for scripts to use.

PR #277, opened 2022-09-29
feat: add `editions` menu,`editions` control and  `has_many_editon` disposition

dyphire:
Todo:
 - [x] add `editions` control.
 
 ~I have no idea how to add a new control. Can anybody help me?~
Close https://github.com/tomasklaen/uosc/issues/20

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -266,6 +266,7 @@ local function create_default_menu()
 				{title = '2.35:1', value = 'set video-aspect-override "2.35:1"'},
 			},},
 			{title = 'Audio devices', value = 'script-binding uosc/audio-device'},
+			{title = 'Editions', value = 'script-binding uosc/editions'},
 			{title = 'Screenshot', value = 'async screenshot'},
 			{title = 'Show in directory', value = 'script-binding uosc/show-in-directory'},
 			{title = 'Open config folder', value = 'script-binding uosc/open-config-directory'},
@@ -3208,7 +3209,8 @@ function Controls:init()
 		['audio-device'] = 'command:speaker:script-binding uosc/audio-device?Audio device',
 		video = 'command:theaters:script-binding uosc/video#video>1?Video',
 		playlist = 'command:list_alt:script-binding uosc/playlist?Playlist',
-		chapters = 'command:bookmarks:script-binding uosc/chapters#chapters>0?Chapters',
+		chapters = 'command:bookmark:script-binding uosc/chapters#chapters>0?Chapters',
+		['editions'] = 'command:bookmarks:script-binding uosc/editions#editions>1?Editions',
 		['stream-quality'] = 'command:high_quality:script-binding uosc/stream-quality?Stream quality',
 		['open-file'] = 'command:file_open:script-binding uosc/open-file?Open file',
 		['items'] = 'command:list_alt:script-binding uosc/items?Playlist/Files',
@@ -4266,6 +4268,10 @@ mp.observe_property('track-list', 'native', function(name, value)
 	set_state('has_many_video', types.video > 1)
 	Elements:trigger('dispositions')
 end)
+mp.observe_property('editions', 'number', function(_, editions)
+	if editions then set_state('has_many_edition', editions > 1) end
+	Elements:trigger('dispositions')
+end)
 mp.observe_property('chapter-list', 'native', function(_, chapters)
 	local chapters, chapter_ranges = serialize_chapters(chapters), {}
 	if chapters then chapters, chapter_ranges = serialize_chapter_ranges(chapters) end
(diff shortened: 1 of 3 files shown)


scripts/uosc.lua
```diff
@@ -247,6 +247,7 @@ local function create_default_menu()
 		{title = 'Open file', value = 'script-binding uosc/open-file'},
 		{title = 'Playlist', value = 'script-binding uosc/playlist'},
 		{title = 'Chapters', value = 'script-binding uosc/chapters'},
+		{title = 'Editions', value = 'script-binding uosc/editions'},
```
tomasklaen, line 250: Move this under `Utils`, below `Audio devices`. None is using editions, and they are not available in 99+% of files, so this shouldn't be in the main menu for most common features.

scripts/uosc.lua
```diff
@@ -4266,6 +4267,10 @@ mp.observe_property('track-list', 'native', function(name, value)
 	set_state('has_many_video', types.video > 1)
 	Elements:trigger('dispositions')
 end)
+mp.observe_property('edition-list/count', 'native', function(_, editons)
+	if editons then set_state('has_many_editon', editons > 1) end
```
tomasklaen, line 4271: As with chapters, `has_many_*` is not needed here. Only add `has_edition`, which should be `editions > 0` check, not `1`.

scripts/uosc.lua
```diff
@@ -4487,6 +4492,26 @@ mp.add_key_binding(nil, 'chapters', create_self_updating_menu_opener({
 	end,
 	on_select = function(time) mp.commandv('seek', tostring(time), 'absolute') end,
 }))
+mp.add_key_binding(nil, 'editions', create_self_updating_menu_opener({
+	title = 'Editions',
+	type = 'editions',
+	list_prop = 'edition-list',
+	list_serializer = function(_, editions)
+		local items = {}
+		local editions = mp.get_property_native('edition-list', {})
```
tomasklaen, line 4501: `editions` are already passed as a function parameter, no need to retrieve them again here.

tomasklaen (changes requested):
To add to control shorthands, search for `shorthands`.

Also document editions menu command in readme, and new control in `uosc.conf`

scripts/uosc.lua
dyphire: Most Matroska files have a version, and `editions` menu or control can only be useful for files with more than 1 editions. In most cases, it is not needed, so `editions > 1` is necessary.

scripts/uosc.lua
dyphire: In fact, in my test. if we don't do this, it will get a nil value when switching edition and crash the script, just like this:
https://github.com/tomasklaen/uosc/blob/f9c062f9927985ec911fd97a8bb30987e5a53b69/scripts/uosc.lua#L4450-L4454

scripts/uosc.lua
dyphire: This is the display of the open editions menu on most mkv files.
image

Obviously, it is unnecessary to use the editions menu in this case, especially the display of editions control.

dyphire:
> Also document editions menu command in readme, and new control in `uosc.conf`

Done.
> To add to control shorthands, search for `shorthands`.

I found that I made a mistake before, and now I have successfully added the `editions` control.
Do you have any suggestions for the icon of control? I can't find a better one.

hooke007:
> icon of edition

How about `toc` or `segment` ?

dyphire:
> > icon of edition
> 
> How about `toc` or `segment` ?

The `segment` looks good and more appropriate.
Let's wait for others' opinions

christoph-heinrich:
Chapters currently use `bookmarks`, and from my understanding editions are different sets of chapters, so something like a collection of bookmarks would be great, but that doesn't exist (`collections_bookmark` maybe?). But if we were to change chapters to `bookmark`, then editions could be `bookmarks`.

It's challenging because the chapters are already a collection, end so an editions are a collection of collections.

dyphire:
> But if we were to change chapters to `bookmark`, then editions could be `bookmarks`.

Do this.
> It's challenging because the chapters are already a collection, end so editions are a collection of collections. I don't know how editions are usually used in practice, so maybe my thinking is completely wrong here.

Such concerns should be unnecessary. Although different editions can have different chapter-list, the behavior of the editions is more like imitating what multiple video tracks need to do. ref: [edition-flags](https://www.ietf.org/archive/id/draft-ietf-cellar-matroska-06.html#name-edition-flags).

In my test, the script can work normally for both editions and chapters.
This is an example of editions menu under a video file with multiple editions:
image

christoph-heinrich:
> > But if we were to change chapters to `bookmark`, then editions could be `bookmarks`.
> 
> Done this.

I was just throwing it out there as an idea, not saying you *should* change it :smile:.
Let's see what @tomasklaen has to say about that whole icon situation.

dyphire:
> I was just throwing it out there as an idea, not saying you _should_ change it 😄. Let's see what @tomasklaen has to say about that whole icon situation.

Ah, I just like this suggestion from a personal point of view. 😄

tomasklaen:
> Also I just noticed that there is literally an icon for audio tracks but the default config doesn't use that? Pretty sure it used to.

`audiotrack` icon is a note, which evokes music in me, so I've changed it to the audio wave icon, as it is more appropriate.

PR #284, opened 2022-10-02
fix: keep no menu selection on update

### LINKED ISSUE #282: Self updating menus loose selection on active prop change
Started happening with 0ad72e5f85002406e962042a07f4120b981b2701.

This used to be particularly noticeable with the chapter menu, because it's active prop was `playback-time` and so this constantly happened during playback. That has improved a lot with the recent commit 5b70a762e0f51591fa1279d0ea513ca4a160296e, but it does still happen when the current chapter changes.

I still had an mpv instance open from before that commit, which is where I noticed it. Here is a recording from that instance.

https://user-images.githubusercontent.com/8932183/193411758-f249b61a-4393-483b-b829-c2b5dd9c634f.mp4

To solve this we could either bring back a separate serializer for the active prop, or always open a new menu where we select the same item as before instead of updating it, but loosing the selection does make sense when the list itself actually changes.

christoph-heinrich:
I hope this doesn't break anything. From testing and looking at the code it *should* be fine, but with all the ways a menu can be opened/updated, I might have missed something.

Closes #282.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -1741,10 +1741,11 @@ function Menu:update(data)
 	local current_selected_index = self.current.selected_index
 
 	self:update_content_dimensions()
+	self:reset_navigation()
 	-- `update_content_dimensions()` triggers `select_item_below_cursor()`
+	-- `reset_navigation()` sets an index if current one is nil
 	-- so we need to remember and re-apply `selected_index`.
 	self.current.selected_index = current_selected_index
-	self:reset_navigation()
 end
 
 ---@param items MenuDataItem[]


tomasklaen:
This is not a sound fix. It's duct tape-y, and introduces a race condition (new selected_index might not get rendered if there is no render queued).

Lets continue in #282. I've also refactored it a bit in 85e5ac0, try it out, maybe it helps.

christoph-heinrich:
There is no way that introduced a race condition with rendering. If there was one, it already was there before. After all this whole thing runs in one thread and the current execution does not get interrupted by timers or whatever.

PR #289, opened 2022-10-03
feat: add `external` hint for tracklist menu

dyphire:


christoph-heinrich:
Looking at that code now, I'll rewrite that hint filtering thing once this is merged. It's a miracle that it even works and it working is [not guaranteed](https://www.lua.org/manual/5.1/manual.html#2.5.5). But this PR looks fine as is.

tomasklaen:
I'm doing some tweaks so I'll refactor this too.

PR #292, opened 2022-10-03
feat: add buffering indicator

christoph-heinrich:
This is merely a proof of concept. I've been wondering how it would look ever since #252.
Didn't know what texture to use, so I simply took the the same as in the timeline and volume slider.

Any ideas on how it should look are welcome.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -3757,8 +3757,8 @@ end
 
 ---@param id string
 function Curtain:register(id)
-	self.dependents[#self.dependents] = id
-	self:tween_property('opacity', self.opacity, 1)
+	self.dependents[#self.dependents + 1] = id
+	if #self.dependents == 1 then self:tween_property('opacity', self.opacity, 1) end
 end
 
 ---@param id string
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -437,6 +437,8 @@ local state = {
 	mouse_bindings_enabled = false,
 	uncached_ranges = nil,
 	cache = nil,
+	cache_buffering = 100,
+	cache_underrun = false,
 	render_delay = config.render_delay,
 	first_real_mouse_move_received = false,
 	playlist_count = 0,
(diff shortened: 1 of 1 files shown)


christoph-heinrich:
Buffering detection is better now, but still far from perfect.

dumblob:
Thanks for the efforts!

Could you also post a picture for others to ease their contribution on how it could look like?

tomasklaen:
This looks horrible. It'd be way better to do this with icons. Options:

1. Static `pending` icon with **Buffering** text below it.
2. Cycling `hourglass_empty` → `hourglass_bottom` → `hourglass_full` icons.
3. Implementing ass rotation and rotating `cached` icon.

It also needs a curtain while enabled, but I'll do that later as curtain and everything using it will need a refactor for this.

christoph-heinrich:
> This looks horrible.

Which is why it's a draft and asking for ideas :wink:

>     1. Static `pending` icon with **Buffering** text below it.
> 
>     2. Cycling `hourglass_empty` → `hourglass_bottom` → `hourglass_full` icons.
> 
>     3. Implementing ass rotation and rotating `cached` icon.

I also thought of a rotating icon (since most buffering indicators are basically that), but the shifting texture was easier for a poc. This was mostly about getting the basics down and getting the idea out there for suggestions.

tomasklaen:
I've refactored the curtain visibility handling. Any element that needs it can now register itself as a dependent with:

```lua
Elements.curtain.register('buffering')
```

And when it's done:

```lua
Elements.curtain.unregister('buffering')
```

christoph-heinrich:
https://user-images.githubusercontent.com/8932183/194105867-47a213b9-7111-48e4-a9fd-8cb2d6930c0e.mp4

tomasklaen:
I was hoping you'd first try the sand clock, as I was curious how'd that look like. Can you try that? :) If it looks good enough it'd be a better solution as we don't have to bloat `:txt` with text rotation.

christoph-heinrich:
Here is what the hourglass would look like.

https://user-images.githubusercontent.com/8932183/194114399-4a27d673-6fa1-4d1f-bf3f-1d62593d5455.mp4

I'm in favor of bloating `:txt` :laughing:

christoph-heinrich:
You sure about those 4 rotations / second ?

https://user-images.githubusercontent.com/8932183/194121781-f0bc5d1f-bf1d-4be1-9cfa-0541df2aea49.mp4

I've pushed a version with 2 rotations per second, which already looks quite fast.

https://user-images.githubusercontent.com/8932183/194122668-2b287097-87d9-4cd5-86ab-7dccbaa15823.mp4

Also that is the `autorenew` icon, because `cached` points in the other direction.

tomasklaen:
Yeah 2rps looks like I imagined 4rps would look in my head :)

hooke007:
Eh, is there anyone feel a little bit noisy when seeking and playing the local files?

hooke007:
If I understand correctly, this feature should be designed for streaming rather than local files.
Those who are using HDD could feel more noisy because of the lower R/W speed of disk.

PR #293, opened 2022-10-03
feat: add idle logo

christoph-heinrich:
We can't make the logo scale with window size, like it does in osc.lua, so I decided to make it bigger by a factor of 2.

Changing the size by factors of 2 is easy, but dynamically scaling it's size would require us to take the strings from the logo table and make lists with them where we then scale each number before converting to a string. This would also be required if we were to implement hidpi support that actually renders in native resolution.

osc.lua renders with a fixed vertical resolution of 360, so they get their scaling for free.

I've asked in the IRC if anyone has a problem with this, but no answer in the 6.5 hours since.

### DIFF
--- a/README.md
+++ b/README.md
@@ -541,6 +541,14 @@ This property can now be used as a control button badge by prefixing it with `@`
 controls=command:icon_name:command_name#@foo?My foo button
 ```
 
+### `osc-idlescreen <mode> <no_osd>`
+
+Controls the visibility of the mpv logo on idle.
+
+Valid arguments for ``mode`` are ``yes``, ``no``, and ``cycle`` to toggle between yes and no.
+
+Depending on the value of ``no_osd`` an osd message can be shown.
+
 ## Why _uosc_?
 
 It stood for micro osc as it used to render just a couple rectangles before it grew to what it is today. And now it means a minimalist UI design direction where everything is out of your way until needed.
--- a/script-opts/uosc.conf
+++ b/script-opts/uosc.conf
@@ -179,6 +179,10 @@ subtitle_types=aqt,ass,gsub,idx,jss,lrc,mks,pgs,pjs,psb,rt,slt,smi,sub,sup,srt,s
 font_height_to_letter_width_ratio=0.5
 # Default open-file menu directory
 default_directory=~/
+# Show the mpv logo and message when idle
+idlescreen=yes
+# Disable santa hat on idle logo in december
+greenandgrumpy=no
 
 # Convers some common chapter types into chapter range indicators.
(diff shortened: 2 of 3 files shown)


hooke007:
Maybe hiding logo&msg when the menu opening would be better.

dyphire:
I think we should also add an option to control the logo display, and allow other scripts to control the logo display through command messages.
Ref: https://github.com/mpv-player/mpv/issues/10201, https://github.com/mpv-player/mpv/commit/ec236f7a99754adb95fada8195f25c03c2218545

image

christoph-heinrich:
It now provides the same interface to other scripts as osc.lua.
Hopefully any scripts sending that script message broadcast it instead of targeting osc directly.

An option has also been added.

tomasklaen:
I'm strongly against this. I'm trying not to add anything non essential into a file that is already almost 5k lines, and this is a pretty useless decoration that won't be seen 99.9% of the time using uosc. And on top of it as mentioned above it introduces collisions with other stuff that now has to be handled with additional APIs. Overall the negatives far outweigh the positives here.

dumblob:
@christoph-heinrich could this functionality be split into a separate script to live on its own?

christoph-heinrich:
> @christoph-heinrich could this functionality be split into a separate script to live on its own?

Yes, then it also gets it's own overlay and can scale just like the one in osc. If I name it osc.lua it might even receive script messages that are intended for the original osc. I wonder if that causes any problems.

christoph-heinrich:
I've now made a stripped down version of osc.lua that shows the idle logo and listens to script messages just like the original.
There is only one thing missing... a project name. Ideas?

hooke007:
idle_show.lua
haha...

christoph-heinrich:
But it does more then that, it listens to `osc-message`, `osc-chapterlist`, `osc-playlist`, `osc-tracklist`, `osc-idlescreen` and shows the corresponding messages exactly like the original would.

Also the filename itself stays osc.lua, so no additional configuration or renaming is required. Simply put it in the scripts folder and it'll behave exactly like the original osc.lua would to the rest of the system, except that it doesn't show the actual timeline etc.

So it's like a UI less osc... sort of.

christoph-heinrich:
I ended up calling it [sosc](https://github.com/christoph-heinrich/sosc) because it supplements osc functionality, in case anyone cares.

hooke007:
And many options could be removed. (they did nothing in the lite-osc)

christoph-heinrich:
I actually removed them and then added them back in to prevent warnings caused by options in the osc.conf that are not available in the osc.lua.
``` lua
-- include all other options to suppress warnings when reading the config file
```

PR #298, opened 2022-10-05
feat: better buffering detection

christoph-heinrich:
This is a completely different approach to detecting buffering. It's primarily based on the `core-idle` property.

The good thing about this approach is that it shows the indicator immediately as soon as the video stops and continuously until the video continues. The old approach would already show the indicator a little bit before the video actually stops due to lack of data, and it would also intermittently stop showing while buffering. The down side is that it shows the indicator no matter what the cause of the video stopping playback is. So seeking, even when all the video data is cached, will also show the indicator for a little bit.

Maybe someone else has a good idea for how to handle that.

I've also added `uncached_ranges` as a metric because hard drives https://github.com/tomasklaen/uosc/pull/292#issuecomment-1268980178.

Maybe it would even make sense to use a different logic for enabling the indicator then for disabling it.


hooke007:
Another edge case is `cache=yes` for local files. Someone hopes to read the local file from RAM...

christoph-heinrich:
I've added the old stuff back in again and gave the "different logic for enabling the indicator then for disabling it" idea a shot. It's certainly an improvement, but not perfect. The code is also not great, but that can be cleaned up once the logic is right.

tomasklaen:
Glancing at it I don't see anything wrong, but I haven't played with it at all, so you'll definitely know better. I'll accept whatever you arrive at.

pintassilgo:
I was experiencing annoying buffering indicator flashing on the screen while seeking local videos, which I don't think should happen. This patch fixed it for me, thanks.

hooke007:
How about killing the anime when the menu is open?

tomasklaen:
@hooke007 That indicator below menu screenshot is quite jarring. You have to have a low menu opacity. This is why I don't like allowing all the customization options that I know will make UI unreadable/usable when changed, but I won't hear the end of it if I remove them... So, whatever floats your boat :)

@christoph-heinrich I think we can just give the spinner `0.1` opacity if `Elements.menu` exists.

christoph-heinrich:
Decided to check for `Elements.menu.is_closing` because `Elements.menu` alone took to long after closing the menu.

po5:
The buffer curtain is too prominent, it shouldn't cover my volume and timeline controls. It should absolutely not dim the window controls.

tomasklaen:
I was just about to test and merge, but that is true. Indicator should dim elements. Haven't thought of that 😞.

So we need to stop BufferingIndicator from using the curtain, and instead render its own as part of the `BufferingIndicator:render()`.

christoph-heinrich:
It draws it's own curtain now, just like the pause indicator.

PR #301, opened 2022-10-06
fix: use mouse-pos property instead of keybinds

### LINKED ISSUE #295: Mouse drag not propagated to other scripts
I did not study the code much but it seems mouse drag is being swallowed even if no action is bound to it in the given context/situation.

E.g. I would like `uosc` to work with `MBTN_LEFT  script-binding drag-to-pan
` in `input.conf` (drag-to-pan comes from https://github.com/occivink/mpv-image-viewer/blob/master/scripts/image-positioning.lua script). But the binding works only if I do not put `uosc` into `scripts/` folder.

Any hints how to propagate the binding instead of swallowing it?

christoph-heinrich:
Closes #295

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -4113,24 +4113,17 @@ function set_state(name, value)
 	Elements:trigger('prop_' .. name, value)
 end
 
-function update_cursor_position()
-	cursor.x, cursor.y = mp.get_mouse_pos()
-
+function update_cursor_position(x, y)
 	-- mpv reports initial mouse position on linux as (0, 0), which always
 	-- displays the top bar, so we hardcode cursor position as infinity until
 	-- we receive a first real mouse move event with coordinates other than 0,0.
 	if not state.first_real_mouse_move_received then
-		if cursor.x > 0 and cursor.y > 0 then
-			state.first_real_mouse_move_received = true
-		else
-			cursor.x = infinity
-			cursor.y = infinity
-		end
+		if x > 0 and y > 0 then state.first_real_mouse_move_received = true
+		else x, y = infinity, infinity end
 	end
 
 	-- add 0.5 to be in the middle of the pixel
-	cursor.x = (cursor.x + 0.5) / display.scale_x
-	cursor.y = (cursor.y + 0.5) / display.scale_y
+	cursor.x, cursor.y = (x + 0.5) / display.scale_x, (y + 0.5) / display.scale_y
 
 	Elements:update_proximities()
 	request_render()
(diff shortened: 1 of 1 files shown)


hooke007:
LGTM

christoph-heinrich:
Cleaned it up a bit.

tomasklaen:
Yup, looks good, just refactor away the `local mouse_keybinds = {}`, as we don't need it anymore.

Also while I'm looking at it, is there an observable property with window position? Because current `pause_on_click_shorter_than` implemented in there cancels the click only after the configured time. It'd be cool if it also canceled on any window movement, because currently if person drags and releases the window still within the configured time, the click (pause) happens, and it shouldn't. I didn't implement it as I couldn't find how to tap into window position in the docs. But maybe someone else knows about something?

christoph-heinrich:
`mouse-pos/hover` becomes false when dragging the window, so we can certainly do something with that.

hooke007:
I was confused about `pause_on_click_shorter_than`. If user's `MBTN_LEFT` is bound to another command such as `cycle mute` , they could not activate the command they want.

tomasklaen:
Well yeah, you can't have more than one function per button. That's to be expected, no? You'd want it to drag, pause AND do what's defined in `input.conf`?

hooke007:
The first time I used uosc, it took me a long time ti figure out why my MBTN_LEFT doesn't work...
There is no related content mentioned in the doc. My idea is making that command configurable.
```lua
local click_timer = mp.add_timeout(duration_seconds, function() mp.command(options.usr_command) end)
```

Natural-Harmonia-Gropius:
I think this function could be separated into another script, as this behavior also becomes sluggish when uosc is under high load.

christoph-heinrich:
Getting that to work correctly is hard because pressing the left mouse button triggers down and up events.
@Natural-Harmonia-Gropius will you make it work right? because then I'll stop tinkering with it.

christoph-heinrich:
I've made it work correctly now.

Natural-Harmonia-Gropius:
Here's a funny issue, just keep it before any real improvement. 🫠
https://github.com/tomasklaen/uosc/pull/301#issuecomment-1271809454

tomasklaen:
If I click without moving the mouse nothing happens.

christoph-heinrich:
While working on https://github.com/mpv-player/mpv/issues/10723#issuecomment-1271882790 I realized that it behaves differently in windows mode then in fullscreen/maximized, so I'll have to adjust it.

Or just drop the last commit, remove that "short click to pause" and use whatever @Natural-Harmonia-Gropius makes instead.

Natural-Harmonia-Gropius:
Mine is trying to be a replacement for input.conf, it won't be available so fast.
At least I think the uosc function works correctly now, so just keep it and make sure the code is not aggressive when the function is turned off.

christoph-heinrich:
Should work now both in windowed and non windowed mode.

tomasklaen:
This doesn't filter drags?

christoph-heinrich:
No, short drags get recognized as short clicks, and I don't see a way of filtering that out.

PR #302, opened 2022-10-06
feat: automatically choose between fast and exact seeking

christoph-heinrich:
ref. https://github.com/tomasklaen/uosc/discussions/300#discussioncomment-3818107

### DIFF
--- a/script-opts/uosc.conf
+++ b/script-opts/uosc.conf
@@ -24,8 +24,6 @@ timeline_border=1
 timeline_step=5
 # Opacity of chapter indicators in timeline, 0 to disable
 timeline_chapters_opacity=0.8
-# When dragging in timeline, use faster seeking to keyframes
-timeline_drag_seek_keyframes=no
 
 # A comma delimited list of items to construct the controls bar above the timeline. Set to `never` to disable.
 # Parameter spec: enclosed in `{}` means value, enclosed in `[]` means optional
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -160,7 +160,6 @@ local defaults = {
 	timeline_border = 1,
 	timeline_step = 5,
 	timeline_chapters_opacity = 0.8,
-	timeline_drag_seek_keyframes = false,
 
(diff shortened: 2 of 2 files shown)


tomasklaen:
Ok, but this should compliment the `timeline_drag_seek_keyframes` option, not remove it. I definitely don't want keyframe seeking to be the default and only option. Just look how bad it behaves:

[pr.webm](https://user-images.githubusercontent.com/47283320/194572795-0d80e974-d604-4d51-92b0-33183666f717.webm)

In comparison with current accurate seeking (`timeline_drag_seek_keyframes` disabled):

[current.webm](https://user-images.githubusercontent.com/47283320/194572907-673f7fa6-7589-4c9a-a017-cbd1bed99883.webm)

christoph-heinrich:
Hmm yeah, that's because it's a short video with few keyframes. I've tested it on ~20min videos and there it behaves much better then always doing exact seeks.

tomasklaen:
Can we detect distances between keyframes? If not, maybe we can just estimate the most common one, and turn on exact seeks if the distance between keyframes on timeline is more than a couple pixels.

christoph-heinrich:
I don't think there is a way for scripts to know the distance between keyframes, and the distance can also vary between each keyframe.
I think a common keyframe interval is 5-10 seconds (YouTube uses 5 second segments afaik)

christoph-heinrich:
I've asked on IRC and the only way to get the keyframe distance would be to to keyframe seeks and measure how much time changes.

How about keyframe seeks when `duration > 60`, exact seeks otherwise.

tomasklaen:
> Or I guess we could even make it dependent on how may pixels there are per second in the timeline, then it changes with window size.

Yup, that's what I've meant. I'd put it at >0.5 pixels per second = keyframes, and exact otherwise.

christoph-heinrich:
You in your video above you have a ratio of 80 pixels/s
0.5 seeks awfully low. That's a 20 min video in window of width 600 or less.

christoph-heinrich:
This seems reasonable to me from my testing now. Try it out, maybe put a print in there to see what the values are.

PR #312, opened 2022-10-09
Live update title

Hill-98:
If `title` uses some variable that is not available when the file is loaded, the title will be wrong.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -4275,7 +4275,10 @@ end)
 mp.observe_property('osc', 'bool', function(name, value) if value == true then mp.set_property('osc', 'no') end end)
 function update_title(title_template)
 	if title_template:sub(-6) == ' - mpv' then title_template = title_template:sub(1, -7) end
-	set_state('title', mp.command_native({'expand-text', title_template}))
+	local title = mp.command_native({'expand-text', title_template})
+	if state.title ~= title then
+		set_state('title', title)
+	end
 end
 mp.register_event('file-loaded', function()
 	set_state('path', normalize_path(mp.get_property_native('path')))
@@ -4288,9 +4291,9 @@ mp.register_event('end-file', function(event)
 		handle_file_end()
 	end
 end)
-mp.observe_property('title', 'string', function(_, title)
+mp.register_idle(function()
 	-- Don't change title if there is currently none
-	if state.title then update_title(title) end
+	if state.title then update_title(mp.get_property_native('title')) end
 end)
 mp.observe_property('playback-time', 'number', create_state_setter('time', function()
 	-- Create a file-end event that triggers right before file ends


tomasklaen:
With this the title won't update when something changes the template during runtime.

Just wrap the `observe_property` in `register_idle`, and remove the `if state.title ~= title then` condition. It's pointless.

Hill-98:
> With this the title won't update when something changes the template during runtime.

It can be changed at runtime, and in fact `register_idle` is always triggered during playback.

christoph-heinrich:
Using `register_idle` seems like a really bad idea for that. That gets triggered as soon as the lua script gets back to the event loop, which happens constantly.

tomasklaen:
Oh, I thought `reginster_idle` fires only once. Yeah, this is definitely not acceptable. It'd be firing at least once per frame...

Hill-98:
I know it's not very elegant, but I can't think of any other way to handle it, and there's a `property-change` event, but I don't think `register_idle` is any different.

Hill-98:
Unless we parse the `title` property to create observer for each property it uses, but that takes a lot of work.

tomasklaen:
```lua
function observe_title()
	mp.unregister_idle(observe_title)
	mp.observe_property('title', 'string', function(_, title)
		-- Don't change title if there is currently none
		if state.title then update_title(title) end
	end)
end
mp.register_idle(observe_title)
```

Hill-98:
That doesn't solve the problem very well.

Extreme examples: `title=${playback-time}`

christoph-heinrich:
We could just take the title property and expand it on each render, like osc.lua does.

Hill-98:
This would also work, although the render trigger frequency is also fast.

christoph-heinrich:
Periodic timer that expands the title every 500ms. It's not a nice solution, but without listening to all available properties, there is no way of knowing when something might change.

I guess it would also be possible to parse the title property and then observe the properties in there.

Hill-98:
I can't help much with this one, I'm not familiar with string handling in lua. if you guys decide you want to use this solution, I'll close this PR.

tomasklaen:
Commit above solves this like so:

```lua
function observe_title()
	-- Idle is needed as some template variables might not be present until everything else initialized
	mp.unregister_idle(observe_title)
	local hot_keywords = {'time'}
	local timer = mp.add_periodic_timer(0.9, function() print('updating') update_title(mp.get_property_native('title')) end)
	timer:kill()
	mp.observe_property('title', 'string', function(_, title)
		-- Don't change title if there is currently none
		if state.title then update_title(title) end
		-- Enable periodic updates for templates with hot variables
		local is_hot = itable_find(hot_keywords, function(var) return string.find(title or '', var) ~= nil end)
		if is_hot then timer:resume() else timer:kill() end
	end)
end
mp.register_idle(observe_title)
```

tl;dr: enables periodic updates only for titles where we detect hot variables. currently this is only titles with `time` keyword. Doesn't look like there are issues with it, but test it out.

Are there other keywords we should be detecting?

Hill-98:
`mp.register_idle(observe_title)` and `mp.unregister_idle(observe_title)` are somewhat redundant, because idle events will happen prematurely, they don't actually address some properties not being available at load time. 

test: `title=${audio-params/channels} - ${audio-params/samplerate}Hz - ${audio-codec-name}`

Hill-98:
Well, I tested again and if I play a file from my NAS's SMB share, it is sometimes unavailable, I guess because of the NAS's lag.

tomasklaen:
So this looks like mpv triggering `file-loaded` prematurely then. I don't see any other event to listen to, and can't think of other sane workaround, so oh well I guess. I'm definitely not going to make everyone's uosc periodically expand title template in the background because some people put weird stuff in there.

@christoph-heinrich do you remember why we need the _"Don't change title if there is currently none"_ condition?

christoph-heinrich:
Looking at my PR and corresponding issue where I introduced it, I don't see a reason why. From a quick test it seems to work fine without it, so we might as well remove it.

dyphire:
> Are there other keywords we should be detecting?

Maybe add `percent` keyword for `${percent-pos}%`.

dyphire:
> @christoph-heinrich do you remember why we need the _"Don't change title if there is currently none"_ condition?
>> Looking at my PR and corresponding issue where I introduced it, I don't see a reason why. From a quick test it seems to work fine without it, so we might as well remove it.

If without it, the title will be displayed as `No file` when mpv under idle state.
image
Maybe the `top_bar_title` should not be displayed when idle.

tomasklaen:
> If without it, the title will be displayed as No file when mpv under idle state.

Which is ... correct, no?

tomasklaen:
It's not meaningless. It's a correct window title for current state.

dyphire:
> It's not meaningless. It's a correct window title for current state.

That’s true. Is it possible not to display it?

tomasklaen:
Not without additional code, so no.

dyphire:
> Not without additional code, so no.

I see, it doesn't matter. I just thought that it may be handled through the automatic configurations.

christoph-heinrich:
Now I remembered why I added that condition. uosc used to not show the "No file", so I added that to make it behave like it did before my change.

dyphire:
> Now I remembered why I added that condition. uosc used to not show the "No file", so I added that to make it behave like it did before my change.

An auto profile to mimics old behavior:
```ini
[idle-title]
profile-cond=idle_active
title=''
```
But it's still not perfect, because it still shows the `top_bar_title`. 
This will overwrite other osd displays:
image
I still prefer the old behavior of your pr implementation.

christoph-heinrich:
Letting a playlist finish with `--keep-open=no --idle=yes` results in this :upside_down_face: 
playlist_idle
Btw I just wrote and tested a function that extracts the properties that are used when expanding that string, so we can observe those properties. Made 15 test cases out of the syntax description in the man page. It's time to integrate that into uosc and solve that whole title thing properly.

christoph-heinrich:
@dyphire You might want to restore the title again when not idle anymore :wink: 
```
[idle-title]
profile-cond=idle_active
title=''
profile-restore=copy-equal
```

PR #318, opened 2022-10-11
fix: properly update title

christoph-heinrich:
Get the properties relevant to the title, observe and update based on their changes.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -654,6 +654,24 @@ function wrap_text(text, target_line_length)
 	return table.concat(lines, '\n'), max_length, #lines
 end
 
+---Extracts the properties used by property expansion of that string.
+---@param str string
+---@param res { [string] : boolean } | nil
+---@return { [string] : boolean }
+local function get_expansion_props(str, res)
+	res = res or {}
+	for str in str:gmatch('%$(%b{})') do
+		local name, str = str:match('^{[?!]?=?([^:]+):?(.*)}$')
+		if name then
+			local s = name:find('==') or nil
+			if s then name = name:sub(0, s - 1) end
+			res[name] = true
+			if str and str ~= '' then get_expansion_props(str, res) end
+		end
+	end
+	return res
+end
+
 -- Escape a string for verbatim display on the OSD
 ---@param str string
 function ass_escape(str)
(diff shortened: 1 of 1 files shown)


christoph-heinrich:
osc.lua removes newlines and trailing slashes. Seems like a reasonable thing to do and while we're at it, we might as well trim whitespace.

po5:
Property expansion can be nested, it's certainly an edge case that almost no one will run into but it exists. Have not checked if it works or not with your current code.  
Here's my lua parser with nesting support if it helps.  
https://github.com/ekisu/mpv-webm/blob/874d444ea5950d982ebb1b9babc17638991e7ed1/src/util.moon#L39-L80

christoph-heinrich:
It already does recursive expansion :wink:

dyphire:
In my test, I found an issue. When the following auto profiles are used, and the files stops playing and returns to the idle state, the wrong title will be displayed.
```ini
[idle-title]
profile-cond=idle_active
profile-restore=copy-equal
title=''

[uosc_filename]
profile-cond=not idle_active and path:match('://')==nil
profile-restore=copy
title=${filename}
```
image

This issue does not exist in the master branch.

christoph-heinrich:
> In my test, I found an issue. When the following auto profiles are used, and the files stops playing and returns to the idle state, the wrong title will be displayed.

I've tried `mpv --keep-open=no --idle=yes` with your auto profiles, dragging in files and letting them run out to return to idle. Works without any problems  for me.

dyphire:
> I've tried `mpv --keep-open=no --idle=yes` with your auto profiles, dragging in files and letting them run out to return to idle. Works without any problems for me.

yes, The latest push fixes it.

christoph-heinrich:
Ah ok, it didn't handle the case where no properties are need for the title, thus no update happened.

dyphire:
> Ah ok, it didn't handle the case where no properties are need for the title, thus no update happened.

I can't reproduce it anyway again, just fine.

christoph-heinrich:
> a closing brace will close every open brace

the default title is `${?media-title:${media-title}}${!media-title:No file}` which does have two successive closing braces. if the first one closed both scopes, we'd always have a "}" in our titles, but we don't.

PR #319, opened 2022-10-11
fix: don't render empty title

christoph-heinrich:
Prevents the empty rectangle seen in https://github.com/tomasklaen/uosc/pull/312#issuecomment-1275013070

The playlist indicator that's seen in the comment below that one is still shown when in idle after a playlist runs out.
That's technically correct as there still is a playlist, but maybe it would be better to hide it in order to have an empty corner again, just like when the player gets opened without files? I'm not sure...

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -3221,8 +3221,8 @@ function TopBar:render()
 		end
 
 		-- Title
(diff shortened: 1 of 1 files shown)


tomasklaen:
If there's a playlist, there should be playlist indicator. I'm fine with how it is now.

PR #320, opened 2022-10-12
fix: improve text width estimation

tomasklaen:
I'm working on improving text width estimation by creating a map of character widths relative to font size. Would love to hear some feedback and help/ideas on how to fix the current issues:

image

1. Some european characters are definitely missing.
2. It works well for most stuff, except some asian text (chinese characters on the screenshot).
3. I also don't like the idea of putting all drawing characters such as `└` into the table.

I guess we need a way to identify character ranges with similar widths, but I'm not sure about the best way how to do that here.

This is just a WIP PR that includes tests in the `render()` function.

### DIFF
--- a/script-opts/uosc.conf
+++ b/script-opts/uosc.conf
@@ -176,10 +176,11 @@ stream_quality_options=4320,2160,1440,1080,720,480,360,240,144
 media_types=3g2,3gp,aac,aiff,ape,apng,asf,au,avi,avif,bmp,dsf,f4v,flac,flv,gif,h264,h265,j2k,jp2,jfif,jpeg,jpg,jxl,m2ts,m4a,m4v,mid,midi,mj2,mka,mkv,mov,mp3,mp4,mp4a,mp4v,mpeg,mpg,oga,ogg,ogm,ogv,opus,png,rm,rmvb,spx,svg,tak,tga,tta,tif,tiff,ts,vob,wav,weba,webm,webp,wma,wmv,wv,y4m
 # File types to look for when loading external subtitles
 subtitle_types=aqt,ass,gsub,idx,jss,lrc,mks,pgs,pjs,psb,rt,slt,smi,sub,sup,srt,ssa,ssf,ttxt,txt,usf,vt,vtt
-# Used to approximate text width
-# If you are using some wide font and see a lot of right side clipping in menus,
-# try bumping this up.
-font_height_to_letter_width_ratio=0.5
+# Text width approximation options
+# - enable if you're using a monospace font
+monospace=no
+# - tweak for fonts that are wider or narrower
+font_width_scale=1
 # Default open-file menu directory
 default_directory=~/
 
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -225,6 +225,8 @@ local defaults = {
 	stream_quality_options = '4320,2160,1440,1080,720,480,360,240,144',
 	media_types = '3g2,3gp,aac,aiff,ape,apng,asf,au,avi,avif,bmp,dsf,f4v,flac,flv,gif,h264,h265,j2k,jp2,jfif,jpeg,jpg,jxl,m2ts,m4a,m4v,mid,midi,mj2,mka,mkv,mov,mp3,mp4,mp4a,mp4v,mpeg,mpg,oga,ogg,ogm,ogv,opus,png,rm,rmvb,spx,svg,tak,tga,tta,tif,tiff,ts,vob,wav,weba,webm,webp,wma,wmv,wv,y4m',
 	subtitle_types = 'aqt,ass,gsub,idx,jss,lrc,mks,pgs,pjs,psb,rt,slt,smi,sub,sup,srt,ssa,ssf,ttxt,txt,usf,vt,vtt',
+	monospace = false,
+	font_width_scale = 1,
 	font_height_to_letter_width_ratio = 0.5,
 	default_directory = '~/',
 	chapter_ranges = 'openings:30abf964,endings:30abf964,ads:c54e4e80',
@@ -564,7 +566,7 @@ end
 function text_length(text)
 	if not text or text == '' then return 0 end
 	local text_length = 0
-	for _, _, length in utf8_iter(tostring(text)) do text_length = text_length + length end
+	for char, length in utf8_iter(tostring(text)) do text_length = text_length + length end
 	return text_length
 end
 
@@ -587,8 +589,9 @@ function utf8_iter(string)
 
 		local start = byte_start
 		byte_start = byte_start + byte_count
+		local byte_end, length = start + byte_count - 1, (byte_count > 2 and 2 or 1)
 
-		return start, byte_count, (byte_count > 2 and 2 or 1)
+		return string:sub(start, byte_end), length, start, byte_end, byte_count
 	end
 end
 
@@ -603,9 +606,7 @@ function wrap_text(text, target_line_length)
 	local before_line_start = 0
 	local before_removed_length = 0
 	local max_length = 0
-	for char_start, count, char_length in utf8_iter(text) do
-		local char_end = char_start + count - 1
-		local char = text.sub(text, char_start, char_end)
+	for char, char_length, char_start, char_end in utf8_iter(text) do
 		local can_wrap = false
 		for _, c in ipairs(wrap_at_chars) do
 			if char == c then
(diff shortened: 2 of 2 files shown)


Natural-Harmonia-Gropius:
the aspect ratio of character depends on the font.

Segoe UI
image

hooke007:
I think it could be out of control. I don't know how libass did that but [`--sub-back-color`](https://mpv.io/manual/master/#options-sub-back-color) & [`--sub-shadow-offset`](https://mpv.io/manual/master/#options-sub-shadow-offset) work well.

tomasklaen:
Yes, that's what `font_width_scale` is there for. But looking at your screenshot, the asian characters in segoe have different width ratios 😕 hm...

tomasklaen:
I wonder if we should just ship a good and consistent font and ignore `osd-font` altogether. This would eliminate all of the issues and inaccuracies with text width estimation.

We could still add `use_osd_font` option for people who like to suffer.

dyphire:
> I wonder if we should just ship a good and consistent font and ignore `osd-font` altogether. This would eliminate all of the issues and inaccuracies with text width estimation.
>
> We could still add `use_osd_font` option for people who like to suffer.

I don't think we can find a good and consistent font. 
Especially for Asian characters, if language specific fonts are not used, incorrect glyphs rendering will occur.

The following is an example of Asian glyph differentiation:
When use `Source Han Mono`
image


When use `Source Han Mono SC`
image

Note that  different language specific fonts render different results for this character of `入`.
This phenomenon is common in Asian characters.

Therefore, people who want to specify a specific font are not like to suffer.

christoph-heinrich:
We could have one table with character ranges and their rough widths, and another table for exceptions.
The only problem is that going through all characters exhaustively is next to impossible. Ideas for some way of quickly getting character widths?

The only thing I can think of is having some slider like the volume slider and printing a bunch of the same char. the slider adjusts the estimated width for that char, and when the background lines up with the real width, we have a good estimation. But good luck doing that for all characters...
One improvement would be to not just print one, but rows with different characters at the same time, then a good general estimation and outliers become clear.

tomasklaen:
So I'm pretty set on including a font in the installation archive and calibrating the text width estimation for it. The questions are now:

1. What font to use? It has to be some good looking font that covers 99 percentile use case of latin, asian, arabic, and cyrilic
2. Should we just use a monospace font which would make this absolutely trivial? This would also fix stuff like these hints looking pretty bad:
    image
4. If not and we need tables, how to identify character ranges here? I'm not the author of the utf8 iterator we're using, so not exactly sure what's going on there :)

I'm mainly interested in your font suggestions as I'm only familiar with fonts that cover latin. I'll try them out later and determine where to move from here based on what's available and how it looks like.

> printing a bunch of the same char ... and when the background lines up with the real width, we have a good estimation

That's how I got the current table.

Natural-Harmonia-Gropius:
> What font to use? & Should we just use a monospace？

Others say this is a modern looking and monospace font, but I haven't used it in UI.
https://github.com/be5invis/Sarasa-Gothic

> how to identify character ranges here?

It's super hard to maintain.
https://en.wikipedia.org/wiki/Unicode_block

> I'm only familiar with fonts that cover latin.

If we use an mono font, We could just use "0" for latin and "水" for CJK.
https://developer.mozilla.org/en-US/docs/Glossary/advance_measure

dyphire:
> What font to use? It has to be some good looking font that covers 99 percentile use case of latin, asian, arabic, and cyrilic

Maybe consider [Noto Sans Mono](https://fonts.google.com/noto/specimen/Noto+Sans+Mono)

christoph-heinrich:
I've had an idea for width calculation a while ago, but totally forgot about it again.

With [compute_bounds](https://mpv.io/manual/master/#command-interface-compute-bounds) we can find out how wide a character is relative to it's font size, so whenever we encounter a character that we don't know the width yet, draw an overlay and measure it's size. With that we dynamically build up a table of known widths, independent of the font used.

Problem with that is that there are characters whose width depends on the context. To work around that we could measure each whole string instead and build up a table of known string widths.

I don't know how bad that would be for performance, but I'll try implementing that later and make a PR so you can try it.

christoph-heinrich:
Just made a proof of concept script for the `compute_bounds` idea and it comes with a few complications.
1. the width/height ratio varies with font size
2. the width/font size ratio varies with font size
3. bold and italic influence size (probably borders too, but didn't test)
4. some letters seem to have more space at the sides then others?!?

Here is the script, in case anyone wants to play with it.


<details>
  <summary>Click me</summary>

  ```lua
  local osd = mp.create_osd_overlay("ass-events")

  local config = { font = mp.get_property('options/osd-font') }
  local options = {}
  local bgt = 'ffffff'
  local bg = '000000'

  local assdraw = require("mp.assdraw")

  local ass_mt = getmetatable(assdraw.ass_new())

  function opacity_to_alpha(opacity)
    return 255 - math.ceil(255 * opacity)
  end

  function ass_mt:txt(x, y, align, value, opts)
    local border_size = opts.border or 0
    local shadow_size = opts.shadow or 0
    local tags = '\\pos(' .. x .. ',' .. y .. ')\\rDefault\\an' .. align .. '\\blur0'
    -- font
    tags = tags .. '\\fn' .. (opts.font or config.font)
    -- font size
    tags = tags .. '\\fs' .. opts.size
    -- bold
    if opts.bold or (opts.bold == nil and options.font_bold) then tags = tags .. '\\b1' end
    -- italic
    if opts.italic then tags = tags .. '\\i1' end
    -- rotate
    if opts.rotate then tags = tags .. '\\frz' .. opts.rotate end
    -- wrap
    if opts.wrap then tags = tags .. '\\q' .. opts.wrap end
    -- border
    tags = tags .. '\\bord' .. border_size
    -- shadow
    tags = tags .. '\\shad' .. shadow_size
    -- colors
    tags = tags .. '\\1c&H' .. (opts.color or bgt)
    if border_size > 0 then tags = tags .. '\\3c&H' .. (opts.border_color or bg) end
    if shadow_size > 0 then tags = tags .. '\\4c&H' .. (opts.shadow_color or bg) end
    -- opacity
    if opts.opacity then tags = tags .. string.format('\\alpha&H%X&', opacity_to_alpha(opts.opacity)) end
    -- clip
    if opts.clip then tags = tags .. opts.clip end
    -- render
    self:new_event()
    self.text = self.text .. '{' .. tags .. '}' .. value
  end

  mp.observe_property("osd-dimensions", "native", function(name, val)
    local w, h, a = mp.get_osd_size()

    local fs = w/4
    local text = "fffffffff"

    local ass = assdraw.ass_new()
    
    ass:txt(0, 0, 7, text, {size = fs, bold = false, opacity = 0.1})

    osd.res_x = w
    osd.res_y = h
    osd.data = ass.text
    osd.z = 0
    osd.compute_bounds = true

    local res = osd:update()
    local x0, x1, y0, y1 = res.x0, res.x1, res.y0, res.y1
    w, h = x1 - x0, y1 - y0

    print(x0, x1, y0, y1)
    print("w h", w, h)
    print("w/h", w/h)
    print("w/fs", w/fs)
    print("fs", fs)
  end)
  ```
</details>

dyphire:
>I wonder if we should just ship a good and consistent font and ignore altogether
> We could still add `use_osd_font` option for people who like to suffer.

This patch may be considered when implementing the plan, just specify a default font.
```patch
From 61a9d66515fe421ba57d0d920bba39c7055b2b77 Mon Sep 17 00:00:00 2001
From: dyphire <qimoge@gmail.com>
Date: Thu, 13 Oct 2022 11:04:18 +0800
Subject: [PATCH] feat: add font options to custom UI font

---
 scripts/uosc.lua      | 3 ++-
 1 files changed, 2 insertions(+), 1 deletion(-)

index 7740431..1ce01d4 100644
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -203,6 +203,7 @@ local defaults = {
 	shuffle = false,
 
 	ui_scale = 1,
+	font = '',
 	font_scale = 1,
 	text_border = 1.2,
 	pause_on_click_shorter_than = 0, -- deprecated by below
@@ -284,7 +285,7 @@ local config = {
 	-- sets max rendering frequency in case the
 	-- native rendering frequency could not be detected
 	render_delay = 1 / 60,
-	font = mp.get_property('options/osd-font'),
+	font = options.font ~= '' and options.font or mp.get_property('options/osd-font'),
 	media_types = split(options.media_types, ' *, *'),
 	subtitle_types = split(options.subtitle_types, ' *, *'),
 	stream_quality_options = split(options.stream_quality_options, ' *, *'),
-- 
```

tomasklaen:
After some research, this are the walls I've hit:

- Can't use monospace fonts, they just look bad as a UI font. So it'd have to be a normal font + char width lookup table.
- There is no free/open font that provides a single file with all common currently used characters inside. Noto is divided into tons of small font files each per some region/language group, and afaik there's no way to make that work in ass tags. **But**, I've found this [Go Noto Universal](https://github.com/satbyy/go-noto-universal) fork that does just that. Unfortunately it only comes in regular weight, which is bad. We need at least medium for it to look good and be readable, and also need to provide bold.
- Even if the fork would provide required weights, Noto on its own has other issues, such as characters being too tiny and not correctly vertically aligned. I was able to fix these in FontForge, but when I then try to generate the font I get some unicode range error that says font file might not be usable. While testing it worked fine, but it's likely I broke some characters.

The roadblock as of now is that we just don't have a font to use... The closest one is Go Noto Universal but that only provides regular weight.

I'll look into Go Noto build scripts tomorrow, maybe I can use them to generate missing weights.

christoph-heinrich:
What if we continue using our current width estimation by default, but provide the user the option to generate width estimation for their exact setup.

If we limit it to the first 65535 code points (all [basic multilingual planes](https://en.wikipedia.org/wiki/Plane_(Unicode)#BMP) ) the generation would take 55 minutes (based on the 50ms per measurement I observed in #322).

Any characters outside of that range will continue to use our current width estimation.

By putting the generation process into its own lua file, we can start it using a script message, receive progress reports with script messages with that even have a progress bar and an eta.

By setting the opacity to 0 it can all happen in the background while the user is using mpv as usual.

We could have an entry in the default menu that then opens up some sort of dialog box describing what it is, that it takes roughly an hour (I actually suspect that it takes 3 refresh cycles, so higher refresh displays might be done quicker) and if they accept, the generation happens in the background.

tomasklaen:
Absolutely no installation scripts or background scripts. It's already way too annoying to set uosc up with all the configs, script bindings, messages, and menu comments. We need to move towards simplifying it not making it even more complicated. Trying out a UI plugin shouldn't feel like you're installing an operating system.

If we manage to create a font we can use, a simple lookup table with most common stuff, and average widths for other code ranges would be a good enough solution.

christoph-heinrich:
It would be purely optional and not required for "trying out a UI plugin".

Independent of if it gets exposed to the user or not, the code that generates character widths and calculates string widths based on that needs to be written. I'm currently digging through Unicode stuff to see what needs to be handled specially. #324
It wont be perfect, but I'd like it to be as good as possible without implementing the entire Unicode spec.

tomasklaen:
I've played with it for a couple hours, and we need to use the [go noto universal](https://github.com/satbyy/go-noto-universal) tools as they have scripts to generate subsets of asian fonts that contain only the currently used characters. This is necessary to fit into the 65k char limit for font files.

Unfortunately, I was unable to make it work. I hate python so much...

If anyone has a bit more experience with duct taping environments to make python stuff work, it'd be awesome.

We need to modify their `create_cjk_subset()`, `create_korean_hangul_subset()`, and `create_japanese_kana_subset()` helpers to generate medium and bold fonts.

Then, modify and use their `GoNotoCurrent` category to generate medium and bold fonts, but this also needs to be stripped a bit, as even though it is supposed to generate fonts with 61k chars, I'm still running into some overflow errors when trying to modify it in FontForge, and that is necessary to fix size and vertical alignment issues.

So we need to go through it and drop anything that looks excessive, such as `Cherokee`, etc..

christoph-heinrich:
I think I've unlocked the holy gail of text width calculation by removing one line of code from #322.
Look at those bounds measure times. #322 is now actually usable.

image

Noticed it when writing the code for measuring (almost) all of the first 65535 unicode characters.

I'll quickly rebase on master to get the latest fixes.

PR #321, opened 2022-10-12
fix: menu title/hint clipping

christoph-heinrich:
Having a (mostly) accurate text widths has revealed a mistake in the cut coordinate calculation.

before:
title_clipping
after:
title_no_clipping


dyphire:
I found that this pr introduced a new hint clipping mistake.
before
image
after
image

dyphire:
> What's the title in those items? They don't get cut off at the same position, indicating that the title is different. I have tested this PR with the text width measurement from #322 before submitting and didn't notice any problem with it.>

I have updated the screenshot to show the full information.
> I don't see anything wrong with the calculation of the cut off position, but we can try removing the spacing from the cutting to give us a little bit more leeway

The patch fixes it for me.
image

PR #322, opened 2022-10-12
feat: accurate text width

christoph-heinrich:
Implementation of the `compute_bounds` idea from https://github.com/tomasklaen/uosc/pull/320#issuecomment-1276348405.
Each bounds calculation takes ~50ms, but subsequent lookups are very fast.

Opening the menu for the first time is awfully slow.

Moving over the timeline was also unbearably slow because the time always changes, so I let it use the old width calculation because it works pretty well for the time anyway.


tomasklaen:
Yeah this is way too slow. It'll have to be a lookup table calibrated to a bundled in font.

christoph-heinrich:
Massive performance improvement by removing `text_osd:remove()`.
Bounds measurement time went from ~50ms down to ~200us.

christoph-heinrich:
I think it's pretty much as good as it gets now. Requires #325 to not have clipping in the menu.

Technically it could be a bit better by incorporating parts of #324 in the initial width estimation, but that would be huge amounts of code for only marginal gain in some situations. It's already pretty darn good as is.

I left a debug message in there for testing. Ideally the loop will never repeat, but depending on the text, font and the initial width estimation it might sometimes. In my testing it doesn't anymore, but I didn't test cyrillic, cjk or any other exotic stuff.

Ideally #325 gets merged, I'll rebase and then there shouldn't be any clipping or superfluous spacing anywhere anymore.

tomasklaen:
Damn this looks good. Though I've tried opening a menu with a couple thousand items and `Menu:update_content_dimensions()` took almost a second to calculate. But I think we can make it faster!

Here are my suggestions:

### Caching

Instead of caching whole texts, we cache only the width/height ratio of individual characters, and then compose the text width of combined ratios of all characters times font size. The compromise here is that this would ignore kerning, but I absolutely don't mind that. Stripping kerning from estimation will make it only a couple pixels bigger, so still great estimation, and no risk of clipping.

### Measuring

When measuring each character, we do that by multiplying it lets say 10 times (`string.rep(char, 10)`), and setting the measured font size to 20 pixels so it's big enough to give us an accurate estimation (it's as if we're measuring 1 character 200 pixels big) and should never clip out as what window will ever be smaller than 200 pixels, but this is where the `repeat` loop might kick in.

The reason why I want to cap the measured size to 20 pixels is that rendering too many pixels with libass is slower, so we can squeeze a bit of performance here too.

christoph-heinrich:
> Damn this looks good.

Yup :smile: 

>Though I've tried opening a menu with a couple thousand items and `Menu:update_content_dimensions()` took almost a second to calculate. But I think we can make it faster!

Damn, didn't even think of opening folders with thousands of files in them.

> The reason why I want to cap the measured size to 20 pixels is that rendering too many pixels with libass is slower, so we can squeeze a bit of performance here too.

I used to render at a fixed size at the beginning, but that would lead to somethings being waaay wider then they're supposed to.
image
Measured at size 20.
But with repeating characters enough times, that might not be a problem anymore.

> We really don't need to be pixel perfect. All I'm trying to do is have a good enough elimination to prevent these 2 issues:

I think that will still be possible, because of context based widths and what not. Unicode has lots of weird stuff that we just avoid but using the whole string, otherwise I would have already tried a character based approach.

I would still like to retain the option for very accurate but potentially slow width calculation. Now that I've seen that it's possible, I don't want to go back :laughing:

christoph-heinrich:
I've now implemented a basic character based width calculation, but the character based widths are too far off from the real widths and I don't know why. Increasing the character size or font size makes the problem worse instead of better.
There has to be something obvious that I'm missing. Could you have a look please?

Calculating string widths of strings that we didn't have before, but where we already have the characters of is in the single digit microseconds with the character based approach.

Exceptional Unicode ranges will be added once the calculation itself is right.

christoph-heinrich:
Enclosing characters in `|` did help (space did indeed get measured as 0 width), but now it's consistently over instead of sometimes over and sometimes under.
image

The spacing tag is weird. The whole text measurement barely increases with it, but character based measurement becomes more accurate???

christoph-heinrich:
Choosing too large font sizes for character size calculation leads to under estimation of text widths :confused:.

I ended up going with a font size cap of 50 and a character count of 20, which was  a good compromise between accuracy and speed in my testing. Going smaller increases error but the calculation time barely changes. Bigger/longer increases calculation time (bigger in particular), but the accuracy barely gets better (+ the issue mentioned above).

Spaces get measured too big for some reason. That doesn't really seem to be a problem in practice, but it's noticeable when comparing the measuring methods. quality-menu often has a bunch of spaces at the end (all lines are the same length), and there some lines get measured ~30% too big, but with other lines actually having characters instead of spaces there, the total menu width is still good.

I'm not all that happy with how the enclosing character thing is handled, but I couldn't think of a better way of writing it because it's a circular dependency.

I've added an option `text_width_estimation` for switching between whole text measurement and character measurement.

@dyphire @hooke007 Can you please test how well it works for you? Also check in the console if you get any "repeat" messages. If there is only sometimes one or two it's fine, but if they are too frequent, we'll have to do something about that.

scripts/uosc.lua
```diff
@@ -205,6 +205,7 @@ local defaults = {
 	ui_scale = 1,
 	font_scale = 1,
 	text_border = 1.2,
+	text_width_estimation=false,
```
hooke007, line 208: diff with conf `text_width_estimation=yes`

christoph-heinrich: conf files use yes/no for booleans

christoph-heinrich:
There is some problem with italic text and whole text measurements that used to not exist. I'll have to investigate that.

scripts/uosc.lua
hooke007: I mean uosc.conf 's value doesn't match script's default value.

hooke007: Accroding to the previous commits and prs, tomasklaen always use the same value.

dyphire:
I didn't see any menu clipping in my tests. But when I try to open some folders, I get obviously delayed loading.

In addition, I noticed that the menu text space behavior of the new menu is not free enough.
before:
image image

after:
image image

This makes the menu seem too compact.

christoph-heinrich:
> I didn't see any menu clipping in my tests. But when I try to open some folders, I get obviously delayed loading whichever `text_width_estimation` is yes/no.

That's because of the large variety of chinese characters. seems like I'll have to implement the optimization that treats all characters in certain unicode ranges as if they were the same character.

> In addition, I noticed that the text space behavior of the new menu is not free enough
> This makes the menu seem too compact.

`menu_min_width` and `menu_min_width_fullscreen`

Thanks for testing :)

christoph-heinrich:
There is something super weird going on, idk if it's a libass bug or something in uosc but I've been looking for it for the past hour and I just can't make sense of it.

I've added a whole text width calculation that calculates every string without bold/italic and also with and then calculates the ratio and average to get good values for bold/italic factors. However, even tho those shouldn't influence the result, the chapter background in the top bar gets smaller when I use `bold = false` `italic = true`. doesn't happen when bold is true.

What's even more intriguing, is that when i modify `txt()` to always append the italic tag, the text in the top bar chapter gets smaller and then the width estimation matches again. wtf....

christoph-heinrich:
Literally just changing the chapter in `TopBar:render()` to not be italic changes the font size.
image
image

And then here is what it looks like if italic gets added to everything unconditionally in `txt()`
image

what is going on x.x

christoph-heinrich:
I've made two minimal patches to reproduce this madness. They work both on master branch and on this PR, which means that problem is independent of what I'm doing here. There must be a problem in the `txt()` function or whatever, I have no idea.
I'll open an issue about it.

scripts/uosc.lua
christoph-heinrich: Ah yes, didn't even notice. Thanks

christoph-heinrich:
Text gets shorter on average when italic, who would have thought. I made screenshots to compare and it's actually true.

christoph-heinrich:
@dyphire `text_width_estimation=yes` should be much faster now for you.

The function `character_based_width` is ugly af now.

dyphire:
> > In addition, I noticed that the text space behavior of the new menu is not free enough
> > This makes the menu seem too compact.
> 
> `menu_min_width` and `menu_min_width_fullscreen`

It's helpful.

> @dyphire `text_width_estimation=yes` should be much faster now for you.

Yes, I see some improvement. But the issue still exists, sometimes the calculate even more than a second.

christoph-heinrich:
> Yes, I see some improvement. But the issue still exists. sometimes the calculate even more than a second, even though there are only dozens of files in the folder.

Hmm weird, I now made a folder with files that have chinese names (random chinese text I found on the internet) and it's pretty quick. Everything in the "CJK Unified Ideographs" unicode block and extensions should be very fast now.

Can you send me a list of filenames or something to test with? Maybe I missed an important Unicode block.

dyphire:
> > Yes, I see some improvement. But the issue still exists. sometimes the calculate even more than a second, even though there are only dozens of files in the folder.
> 
> Hmm weird, I now made a folder with files that have chinese names (random chinese text I found on the internet) and it's pretty quick. Everything in the "CJK Unified Ideographs" unicode block and extensions should be very fast now.
> 
> Can you send me a list of filenames or something to test with? Maybe I missed an important Unicode block.
> 
> Edit: I've taken a bunch of text from your repositories and made 94 files with it. opening that folder is pretty quick, < 100ms I'm sure

It does not only appear on chinese characters, but also under english characters.

https://user-images.githubusercontent.com/61936050/196017828-22fad1e6-b846-4f18-9390-0ccfea22f03a.mp4

christoph-heinrich:
Now after I've turned your config into files (almost 1700 files) it's still not super slow. And I know most of that delay actually doesn't come from the width estimation, because I first forgot the file extension, so the files didn't even appear and it still took a little bit to open.

No idea why some folders in particular are that slow for you while others seem pretty quick. That happens with `text_width_estimation` `yes` and `no`, but not without this PR?

dyphire:
> No idea why some folders in particular are that slow for you while others seem pretty quick. That happens with `text_width_estimation` `yes` and `no`, but not without this PR?

Whatever `text_width_estimation` is `yes` or `no`.

There will be a delay in loading when I open some folders under the master branch, but only about 1 second.

tomasklaen:
I think the slowdown @dyphire is seeing might be file list retrieving interface, and not the actual menu rendering.

@christoph-heinrich try adding time logging to `Menu:update_content_dimensions()`, and then @dyphire should test it and report if there are bigger values in the console. If not, it's the file listing and not text width estimation.

christoph-heinrich:
Added time logging to the menu. Opening that folder with 1700 files with mixed filenames takes 3.8s to open the menu with whole text measurement, and 177ms when using the character based approach.

@dyphire please test :)

dyphire:
well, there are the test results:
<details>
<summary>text_width_estimation=yes </summary>
<pre><code>
[   5.392][i][uosc] Menu:update_content_dimensions 0.0004470000000012
[   9.458][i][uosc] Menu:update_content_dimensions 0.0074500000000022
[  12.577][i][uosc] Menu:update_content_dimensions 0.016719000000002
[  16.775][i][uosc] Menu:update_content_dimensions 8.399999999753e-05 
[  18.117][i][uosc] Menu:update_content_dimensions 0.00051000000000201
[  25.841][i][uosc] Menu:update_content_dimensions 0.0001620000000031
[  26.660][i][uosc] Menu:update_content_dimensions 2.699999999578e-05
[  27.391][i][uosc] Menu:update_content_dimensions 9.6999999996683e-05
[  34.129][i][uosc] Menu:update_content_dimensions 0.00014099999999928
[  34.730][i][uosc] Menu:update_content_dimensions 6.7000000001372e-05 
[  36.247][i][uosc] Menu:update_content_dimensions 5.0999999999135e-05
[  37.157][i][uosc] Menu:update_content_dimensions 6.1000000002309e-05 
[  38.654][i][uosc] Menu:update_content_dimensions 3.9999999998486e-05
[  39.303][i][uosc] Menu:update_content_dimensions 5.9999999997729e-05 
[  40.508][i][uosc] Menu:update_content_dimensions 6.0999999995204e-05
[  41.112][i][uosc] Menu:update_content_dimensions 6.9000000003427e-05
[  45.403][i][uosc] Menu:update_content_dimensions 7.8999999999496e-05
[  51.178][i][uosc] Menu:update_content_dimensions 8.7000000000614e-05 
</code></pre>
</details>

<details>
<summary>text_width_estimation=no </summary>
<pre><code>
[  11.026][i][uosc] Menu:update_content_dimensions 0.029176999999997 
[  16.307][i][uosc] Menu:update_content_dimensions 0.024207000000001  
[  19.838][i][uosc] Menu:update_content_dimensions 8.9999999996593e-05 
[  20.519][i][uosc] Menu:update_content_dimensions 0.004249999999999 
[  21.352][i][uosc] Menu:update_content_dimensions 5.4999999999694e-05 
[  22.336][i][uosc] Menu:update_content_dimensions 0.013148999999999 
[  23.187][i][uosc] Menu:update_content_dimensions 9.4000000004257e-05 
[  37.836][i][uosc] Menu:update_content_dimensions 0.0053650000000047 
[  38.512][i][uosc] Menu:update_content_dimensions 3.3999999999423e-05 
[  39.167][i][uosc] Menu:update_content_dimensions 4.7999999999604e-05 
[  45.578][i][uosc] Menu:update_content_dimensions 0.016463000000002 
[  46.500][i][uosc] Menu:update_content_dimensions 5.9000000000253e-05 
[  46.936][i][uosc] Menu:update_content_dimensions 3.5999999994374e-05 
[  48.476][i][uosc] Menu:update_content_dimensions 0.014741999999998 
[  53.298][i][uosc] Menu:update_content_dimensions 3.9999999998486e-05 
[  55.119][i][uosc] Menu:update_content_dimensions 5.5000000003247e-05 
[  58.531][i][uosc] Menu:update_content_dimensions 0.0036769999999962 
[  59.422][i][uosc] Menu:update_content_dimensions 4.3000000005122e-05 
[  59.907][i][uosc] Menu:update_content_dimensions 3.1000000006998e-05 
[  62.040][i][uosc] Menu:update_content_dimensions 4.2999999990911e-05 
[  65.766][i][uosc] Menu:update_content_dimensions 0.0026860000000113 
[  66.790][i][uosc] Menu:update_content_dimensions 0.0050689999999918 
[  68.363][i][uosc] Menu:update_content_dimensions 4.7000000009234e-05 
[  70.978][i][uosc] Menu:update_content_dimensions 0.0012569999999954 
[  73.145][i][uosc] Menu:update_content_dimensions 3.4999999996899e-05 
</code></pre>
</details>

christoph-heinrich:
Text wrapping now uses the character based width calculation, which is needed for getting the text width of wrapped text correct when using `text_width_estimation=yes`.

There has been a lot of refactoring in the process, and at the end I timed `Menu:update_content_dimensions()` with that one folder again to make sure there isn't a regression and it actually got faster by a surprising amount, from 177ms to 126ms.

Less work is being done when the character is already cached, but since in exchange there is more work done for strings with multiple uncached characters, I thought it would be more of a toss up.

christoph-heinrich:
I think the code is fine as is now.

There is still one problem left though (aside from the italic problem). The global functions in the do block with the text stuff don't get recognized as functions, and instead as global variables with unknown type. Doesn't lead to any warnings or anything, but it's weird.

christoph-heinrich:
It now automatically draws text vertically when the window is more vertical resolution then horizontal.
I also decreased the character count from 20 to 10 because that ended up working better for vertical videos.

Overall both width measuring methods are now more accurate with vertical videos.

christoph-heinrich:
I removed the huge Unicode blocks table in favor of only containing what's necessary.

tomasklaen:
It's still waay too much. Just have a single variable:

```lua
local uniform_character_blocks = {{start, end}, {start, end}, ...}
```

for all blocks such as simplified chinese with characters that are usually similar/same width. No need to keep their names as a string, or have them separated and then merge during runtime.

---

Remove zero width detection completely. Zero width characters are incredibly rare, and if we encounter some, they're just going to be measured as zero. No need to bloat the code with so many lines to prevent this.

---

I don't want to have 2 text measuring methods in the codebase. It's honestly bloat. You'd be pretty much the only person using the more accurate/slow one. Until you hit a situation where it's slow, so you temporarily disable it, and then again when it happens again, and then you just go ah screw it and leave it off completely, and then none will be using it :)

I don't want to provide a feature that we know is going to cause issues. Even optionally.

---

And any other low hanging fruit line saving measures you can thing of. I really don't want 20% of the file be dedicated to measuring text widths.

christoph-heinrich:
> Just have a single variable

It will be harder to maintain if we ever decide to update the blocks in the future but ok.

> Remove zero width detection completely.

No, because e.g. [Combining Diatrical Marks](https://en.wikipedia.org/wiki/Combining_Diacritical_Marks) will not measure zero, even though they usually are in practice with letters like `é`. 

> I don't want to have 2 text measuring methods in the codebase. It's honestly bloat. You'd be pretty much the only person using the more accurate/slow one. Until you hit a situation where it's slow, so you temporarily disable it, and then again when it happens again, and then you just go ah screw it and leave it off completely, and then none will be using it :)

It's not that slow. I don't know how big your folders are, but for me even folders that reach 100 items aren't that common and estimating widths for such a folder takes, what, 200ms? Not enough to be annoying that's for sure.
Removing it only frees up 24 lines of code. Complexity is what makes things hard, not line count and with almost all the code being shared between the two methods, having this barely adds any.

> I don't want to provide a feature that we know is going to cause issues. Even optionally.

It's not like character based estimation is without issues. Whole text measurement might be the only way to get good width estimation for some people using weird alphabets where combining multiple characters into one is very common. We'll never be able to account for everything with the character based approach, but whole text estimation by it's nature takes everything into account.

> And any other low hanging fruit line saving measures you can thing of. I really don't want 20% of the file be dedicated to measuring text widths.

There is always the option of splitting it up into a module and then loading it with dofile or require.

tomasklaen:
> It will be harder to maintain

Keep the comments next to the blocks of course:

```lua
local foo = {
  {a, b}, -- Block name
}
```

Or did you mean something else that makes it harder to maintain? I mainly didn't like how it's split into multiple variables that you're combining during runtime, and how they also include their names as strings which we're not using.

> No, because e.g. [Combining Diatrical Marks](https://en.wikipedia.org/wiki/Combining_Diacritical_Marks)

But characters like `é` are one glyph no? You can write `´` separately, but writing both `e´` doesn't combine them together, and any keyboard/writing app outputs the combined character as a single codepoint, no? Or are there some languages where texts have some invisible characters that modify characters before them?

> _accurate text width estimation_

All right, keep it.

> splitting it up into a module 

Yeah I guess we'll have to rip this bandaid sooner or later. I'll do it when current PRs are resolved.

christoph-heinrich:
>I mainly didn't like how it's split into multiple variables that you're combining during runtime, and how they also include their names as strings which we're not using.

Blocks are now pre-combined and the individual character arrays have been turned into blocks as well. They are combined into short ranges where possible and a few are ranges with just a single code point.

> But characters like `é` are one glyph no? You can write `´` separately, but writing both `e´` doesn't combine them together, and any keyboard/writing app outputs the combined character as a single codepoint, no? Or are there some languages where texts have some invisible characters that modify characters before them?

You're right é is indeed a single code point U+00E9, I should have checked before giving that example. But the point still stands that it is valid Unicode to have an `e` with a `´` diatric separately.

From a quick check the [Hebrew example](https://en.wikipedia.org/wiki/Diacritic#/media/File:Example_of_biblical_Hebrew_trope.svg) on the [diatric Wikipedia page](https://en.wikipedia.org/wiki/Diacritic#Hebrew) contains a bunch of characters with diatrics that can neither be found in the "Hebrew" Unicode block nor in the "Alphabetic Presentation Forms" Unicode block, which are the only blocks mentioning hebrew, so they seem to be unrepresentable by a single code point.

Also there is the extreme example of [Zalgo text](https://en.wikipedia.org/wiki/Zalgo_text).

> do you know about some go-to lua bundler?

Nope and I don't think shipping a merged file would be a good idea. Otherwise people will post errors with line numbers that don't line up with the actual source file. Sure one can still  check the release file and then find the same line in the actual source file, but it's additional work for little benefit.

tomasklaen:
I'm getting a weird issue where around 60-80% of the time when I launch the mpv with test text on screen, I get this:

dunes 00_00_08 920 0001

And the rest I get this (correct measurement I assume):

dunes 00_00_08 920 0002

I thought it might something to do with initial window size during measurement on load, and sure enough, in the clipped example, `fit_on_screen()` returns size always as `160`, while in the correct one as `213.33333333333`.

We might want to skip cache when we know the window size might increase (stuff still loading and size smaller than lets say 400).

---

The bold clip at the 2nd screenshot is due to that bold scale factor being below `1.1`.

scripts/uosc.lua
```diff
@@ -550,108 +550,388 @@ function get_point_to_rectangle_proximity(point, rect)
...
+			local shift = 2 ^ 6
+			local after_shift = unicode
+			for _ = byte_count, 2, -1 do
+				local before_shift = after_shift
+				after_shift = math.floor(before_shift / shift)
+				table.insert(res, 1, before_shift - after_shift * shift + 0x80)
+			end
+			shift = 2 ^ (8 - byte_count)
+			table.insert(res, 1, after_shift + math.floor(0xFF / shift) * shift)
+			return string.char(unpack(res))
```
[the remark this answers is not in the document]
christoph-heinrich: That already happens in [defaults.lua](https://github.com/mpv-player/mpv/blob/master/player/lua/defaults.lua#L2)

christoph-heinrich:
> The bold text clipping is due to the bold scale factor being below `1.1`

Those numbers were calculated by measuring a ton of text in normal, bold, and italic and then taking the average factor.
But I've now done some more testing and it turns out that factor varies greatly based on the character and window size.
E.g. This is with text `'fffffffffffffffff'` and a small change in window size, each line is a new render due to a change in window size. (bold_factor italic_factor)
image

I'll look into it further later, but how is that even possible...

scripts/uosc.lua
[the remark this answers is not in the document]
tomasklaen: If the `defaults.lua` was a thing for long enough so that we don't have to be afraid that someone's mpv doesn't have it, than sure keep it at `unpack()` only, but at least add the `@diagnostic` line so that the editor doesn't yell at me.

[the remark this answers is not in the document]
po5: Blame shows it's been there for 8 years. Will that be enough? :^)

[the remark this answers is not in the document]
tomasklaen: Yeah it is :) I was just trying to be consistent with #155. I guess it wasn't needed after all. But we still need the diagnostic thingy or this turns into error/warning.

christoph-heinrich:
ass is weird... look how the measured size jumps for the normal measurement. It's usually the normal one that jumps. but I've also seen the italic one jumping. Bold also jumps when measuring bold only.
image

Interestingly the text itself doesn't jump in size, it's only the measured size that jumps around
Measuring at a constant font size with changing resolution does not result in jumps, but rounding the font size does result in jumps again, but less frequent because it stays at the same size for longer.

I have no idea what we could do about that. Maybe there is a problem in how mpv gets those bounds.

I'll increase the bold factor to 1.1 and the italic factor to 1.0 because overestimation is better then underestimation and those values match up with your screenshot

> I'm talking only during the initial couple milliseconds when stuff is still loading and window has tiny dimensions

Yes I got that, but what if the size never reaches "big enough"? We could empty the cache once the size reaches "big enough" but still cache stuff before that. Ideally we'd then have two caches, one with "big enough" results and one with things measured below that, but I don't know if that complexity is worth it. We could also store what size something got measured at, and then remeasure when the last measurement was too small and there is currently more space available.

christoph-heinrich:
FYI I made a minimal example that demonstrates the jumping x1 coordinate and asked on the mpv IRC about it, where I was advised to ask the libass IRC. Currently waiting for a response.

<details>
<summary>Minimal example</summary>

```lua
local assdraw = require('mp.assdraw')
local text_osd = mp.create_osd_overlay("ass-events")

local function measure_bounds(ass_text, width, height)
   text_osd.res_x, text_osd.res_y = width, height
   text_osd.data = ass_text
   local res = text_osd:update()
   return res.x0, res.y0, res.x1, res.y1
end

mp.observe_property('osd-dimensions', 'native', function (_, dim)
   local width, height = dim.w, dim.h
   local text = 'abcdefghijklmnopqrstuvwxyz'
   local estimated_width = #text * 0.5
   local font_size = math.min(width / estimated_width, height) * 0.9

   local ass = assdraw.ass_new()
   ass.text = '{\\pos(0,0)\\rDefault\\an7\\fs' .. font_size .. '}' .. text
   local ass_text = ass.text
   text_osd.compute_bounds = true
   x0, y0, x1, y1 = measure_bounds(ass.text, width, height)

   ass:new_event()
   ass.text = ass.text .. '{\\pos(0,0)\\rDefault\\an7\\alpha&H33&\\1c&H0000ff&}'
   ass:draw_start()
   ass:rect_cw(x0, y0, x1, y1)
   ass:draw_stop()
   text_osd.compute_bounds = false
   measure_bounds(ass.text, width, height)

   print('w='..width, 'h='..height, 'x0='..x0, 'x1='..x1, ass_text)
end)
```
</details>

christoph-heinrich:
The answer to the jumping coordinate was "rasterizer works on 16x16 tiles, so up to ~16 pixel border (plus some other padding) is possible". And in order to increase the accuracy of the bounds measurement from mpv it would have to "analyze bitmap data and calculate boundaries from it".
There is a related feature request https://github.com/libass/libass/issues/348.

tomasklaen:
If I understand it correctly, measuring has a 16px precision? So the measured sizes might overshoot a bit, but not cause clipping, which isn't that big of a problem, no?

Though it might lead to big overshoots for very thin characters. I think we need to check if measured size is way smaller than anticipated, and multiply the number of measured characters by some appropriate amount and measure again.

christoph-heinrich:
This PR is cursed...

I've implemented the loop for increasing character count when the measured size is too small, which then revealed a race condition where if the window size changes while character sizes are being measured, we get complete bogus values for width, and they are always the same independent of character and character count, which inevitably leads to an endless loop. To remedy this it now always gets the actual window size when updating the overlay and also in `fit_on_screen`.

Because of the text size cap of 50, this loop now ended up measuring every character twice and to get around that it now tries to figure out how many characters will fit on the screen at the calculated size instead of a fixed count.

tomasklaen:
I've tried a couple things and it looks good now. `Menu:update_content_dimensions()` with 2000 items takes ~30ms, and width estimation is good enough, _except_ bold text:

image

Since the estimation is pretty fast now, I think we can split cache for normal and bold text, and not use a static scale factor, since from the screenshot above it's obvious it just won't ever work right.

christoph-heinrich:
Bold characters are now measured separately and italic text gets scaled by 1.01.

christoph-heinrich:
It now keeps track of what got measured at which resolution and updates it if there is more then 10% more space available now and the measured resolution is below 800.

PR #324, opened 2022-10-12
UTF-8 to Unicode and back + Unicode based width estimation preparation

christoph-heinrich:
Took the Unicode blocks from https://en.wikipedia.org/wiki/Unicode_block and converted them to decimal.

Implemented Unicode code point calculation and the beginnings of a function that maps that code point to a width.
It already finds which range that code point belongs to, but doesn't do any width calculation.

I'll just park that here for now, should be a good starting point for anyone who wants to attempt width estimation based on Unicode blocks.

ref https://github.com/tomasklaen/uosc/pull/320

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -568,30 +568,553 @@ function text_length(text)
 	return text_length
 end
 
-function utf8_iter(string)
-	local byte_start, byte_count = 1, 1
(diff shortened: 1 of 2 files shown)


christoph-heinrich:
With recent improvements in #322 I see little reason to continue working on this, so I'll just park the current state here in case it becomes useful in the future.

PR #325, opened 2022-10-14
fix: spacing and clipping in menu

christoph-heinrich:
This gives us a little bit more leeway to prevent clipping from inaccurate width estimation.

ref. https://github.com/tomasklaen/uosc/pull/321#issuecomment-1278600099


christoph-heinrich:
Testing #322 revealed a shortcoming that required #326 to be included here in order to fix properly.

PR #326, opened 2022-10-14
Add spacing only for elements that take up width

christoph-heinrich:
A menu item without title would still add a spacing between hint and the left content edge.
Empty strings (`width == 0`) also don't need spacing.


christoph-heinrich:
Closed because it got included in #325

PR #331, opened 2022-10-17
feat: drag to scroll for menu

christoph-heinrich:
With this touchscreen support is pretty decent, and while it is very much a WIP, it already works pretty well.

Supporting scrolling by dragging necessitates that menu item selection now happens on mouse up rather then mouse down, but that is how applications usually behave anyway.

I don't like how many variables it introduces, especially all the `*_last_x` ones, but updates come in such irregular intervals that it requires some filtering to get good results. Sometimes there isn't even a milliseconds between position updates, and other times it's >30ms, which then leads to huge variance in calculated velocities. Maybe the updates would be more reliable if we switch to `mouse_move` while the cursor is over elements.

Also the math for scroll position updates via momentum is not great. It does work pretty well, but I'll have a look at how such things are usually implemented to see how it should be done.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -143,6 +143,10 @@ function table_shallow_copy(table)
 	return result
 end
 
+--[[ EASING FUNCTIONS ]]
+
+function ease_out_quart(x) return 1 - ((1 - x) ^ 4) end
+
 --[[ OPTIONS ]]
 
 local defaults = {
@@ -1698,6 +1702,11 @@ function Menu:init(data, callback, opts)
 	self.key_bindings = {}
 	self.is_being_replaced = false
 	self.is_closing = false
+	---@type {y: integer, time: number}[]
+	self.drag_data = nil
+	self.is_dragging = false
+	---@type {y: number, distance: number, time: number, menu: MenuStack}
+	self.fling = nil
 
 	self:update(data)
 
@@ -1892,6 +1901,13 @@ function Menu:scroll_to(pos, menu)
 	request_render()
 end
 
+---@param delta? number
+---@param menu? MenuStack
+function Menu:scroll_by(delta, menu)
+	menu = menu or self.current
+	self:scroll_to(menu.scroll_y + delta, menu)
+end
+
 ---@param index? integer
 ---@param menu? MenuStack
 function Menu:scroll_to_index(index, menu)
@@ -2044,29 +2060,64 @@ function Menu:on_prop_fullormaxed() self:update_content_dimensions() end
 
 function Menu:on_global_mbtn_left_down()
 	if self.proximity_raw == 0 then
-		self:select_item_below_cursor()
-		self:open_selected_item({preselect_submenu_item = false})
+		self.drag_data = {{y = cursor.y, time = mp.get_time()}}
+		self.fling = nil
 	else
 		if cursor.x < self.ax then self:back()
 		else self:close() end
 	end
 end
 
+function Menu:fling_distance()
+	local first, last = self.drag_data[1], self.drag_data[#self.drag_data]
+	if mp.get_time() - last.time > 0.05 then return 0 end
+	for i = #self.drag_data - 1, 1, -1 do
+		local drag = self.drag_data[i]
+		if last.time - drag.time > 0.03 then return ((drag.y - last.y) / ((last.time - drag.time) / 0.03)) * 10 end
+	end
+	return #self.drag_data < 2 and 0 or ((first.y - last.y) / ((first.time - last.time) / 0.03)) * 10
+end
+
+function Menu:on_global_mbtn_left_up()
+	if self.proximity_raw == 0 and not self.is_dragging then
+		self:select_item_below_cursor()
+		self:open_selected_item({preselect_submenu_item = false})
+	end
+	if self.is_dragging then
+		local distance = self:fling_distance()
+		if math.abs(distance) > 50 then
+			self.fling = {
+				y = self.current.scroll_y, distance = distance,
+				time = self.drag_data[#self.drag_data].time, menu = self.current,
+			}
+		end
+	end
+	self.is_dragging = false
+	self.drag_data = nil
+end
+
+
 function Menu:on_global_mouse_move()
 	self.mouse_nav = true
-	if self.proximity_raw == 0 then self:select_item_below_cursor()
+	if self.drag_data then
+		self.is_dragging = self.is_dragging or math.abs(cursor.y - self.drag_data[1].y) >= 10
+		local distance = self.drag_data[#self.drag_data].y - cursor.y
+		if distance ~= 0 then self:scroll_by(distance) end
+		self.drag_data[#self.drag_data + 1] = {y = cursor.y, time = mp.get_time()}
+	end
+	if self.proximity_raw == 0 or self.is_dragging then self:select_item_below_cursor()
 	else self.current.selected_index = nil end
 	request_render()
 end
 
 function Menu:on_wheel_up()
-	self:scroll_to(self.current.scroll_y - self.scroll_step * 3)
+	self:scroll_by(self.scroll_step * -3)
 	self:on_global_mouse_move() -- selects item below cursor
 	request_render()
 end
 
 function Menu:on_wheel_down()
-	self:scroll_to(self.current.scroll_y + self.scroll_step * 3)
+	self:scroll_by(self.scroll_step * 3)
 	self:on_global_mouse_move() -- selects item below cursor
 	request_render()
 end
@@ -2137,6 +2188,13 @@ function Menu:create_key_action(name)
 end
 
 function Menu:render()
+	if self.fling then
+		local time_delta = state.render_last_time - self.fling.time
+		local progress = ease_out_quart(math.min(time_delta / 0.5, 1))
+		self:scroll_to(round(self.fling.y + self.fling.distance * progress), self.fling.menu)
+		if progress < 1 then request_render() else self.fling = nil end
+	end
+
 	local ass = assdraw.ass_new()
 	local opacity = options.menu_opacity * self.opacity
 	local spacing = self.item_padding


christoph-heinrich:
I've implemented your suggestions and it works fine without any filtering. Based on the generated numbers I would have expected to need filtering, but after trying it out it seems fine without.

I also tried `mouse_move` but the improvement over `mouse-pos` is barely existent if at all (I didn't calculate the variance).

> You also need to filter out tiny velocities. When releasing a finger the position often moves slightly, and it's annoying to get a tiny scroll because of that.

I couldn't reproduce that issue with mouse or on a touch device, but velocities below 50 now get filtered out. Let me know if the threshold is fine like that.

christoph-heinrich:
`pos_time` has been renamed to `drag_data`, velocity is set to 0 on mouse down and velocity decays twice as fast now.

With that flings scroll far less then on android now, but it's still usable.

tomasklaen:
Sorry for being nitpicky, but also make the `velocity` a single table with:

```lua
velocity = {value: number, updated: number}
```

properties.

And is there anything else preventing merge here?

christoph-heinrich:
> Sorry for being nitpicky

No worries :wink:

`velocity` is now a table, and `drag_data` also has value names instead of just `[1]` and `[2]`.

The one thing I'd still like to do is maybe change the math. Currently it's just basic cobbled together decay that adjusts with frame rate and that works well enough that we could merge it as is, but something actually physics based would be nice (and more accurate).

tomasklaen:
I always use a time based exponential decay with friction coefficient for stuff like this. It's simple and feels nice and snappy.

```lua
local friction = 0.5 -- velocity decay per 100ms
velocity = velocity * math.pow(1 - friction, time_delta / 0.1)
```

christoph-heinrich:
I tried going down the physics route, but the result with a constant deceleration is pretty bad, and I couldn't get something working where the deceleration depends on the speed (even when I let wolfram alpha work out the equation I didn't get something usable).

So I ended up going with the good old exponential decay similar to how you suggested, but with the euler number because that makes integration easier (yours would have been `v_start * (0.1 * math.pow(1 - f, t / 0.1) - 0.1) / math.log(1 - f)`). With a lambda of 7 the decay is pretty much identical to to your 0.5 friction. There is also a small constant deceleration because otherwise it never reaches 0.

tomasklaen:
Looking at it I felt it was a bit too complex, and noticed a lot of potential simplifications. Also, the equation really doesn't need to be this complex. All it needs is just a dumb time based fractional decay, so that it's easier to maintain and tweak in the future. It also feels more responsive now imo 😳, I always prefer faster snappy animations so that I don't have to wait while stuff slowly animates in before I can start reading again.

Here's my patch:

`fling.distance` is a distance that dragging would cover if it'd continue dragging at current speed for next 300ms.

The animation than applies this distance at a speed of 60% per 100ms.

```diff
diff --git a/scripts/uosc.lua b/scripts/uosc.lua
index fb449cf..b8edf63 100644
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -1699,9 +1699,8 @@ function Menu:init(data, callback, opts)
 	self.is_being_replaced = false
 	self.is_closing = false
 	---@type {y: integer, time: number}[]
-	self.drag_data = nil
-	self.is_dragging = false
-	---@type {y: number, velocity: number, time: number}
+	self.dragging = nil
+	---@type {distance: number, updated: number}
 	self.fling = nil
 
 	self:update(data)
@@ -1897,6 +1896,13 @@ function Menu:scroll_to(pos, menu)
 	request_render()
 end
 
+---@param delta? number
+---@param menu? MenuStack
+function Menu:scroll_by(delta, menu)
+	menu = menu or self.current
+	self:scroll_to(menu.scroll_y + delta, menu)
+end
+
 ---@param index? integer
 ---@param menu? MenuStack
 function Menu:scroll_to_index(index, menu)
@@ -2049,7 +2055,7 @@ function Menu:on_prop_fullormaxed() self:update_content_dimensions() end
 
 function Menu:on_global_mbtn_left_down()
 	if self.proximity_raw == 0 then
-		self.drag_data = {{y = cursor.y, time = mp.get_time()}}
+		self.dragging = {{y = cursor.y, time = mp.get_time()}}
 		self.fling = nil
 	else
 		if cursor.x < self.ax then self:back()
@@ -2057,65 +2063,54 @@ function Menu:on_global_mbtn_left_down()
 	end
 end
 
-function Menu:fling_velocity()
-	local last_drag_data = self.drag_data[#self.drag_data]
-	local last_y = last_drag_data.y
-	local last_time = last_drag_data.time
-	if mp.get_time() - last_time > 0.05 then return 0 end
-	for i = #self.drag_data - 1, 1, -1 do
-		local drag_data = self.drag_data[i]
-		local time = drag_data.time
-		if last_time - time > 0.03 then
-			return (last_y - drag_data.y) / (time - last_time)
-		end
+function Menu:fling_distance()
+	local first, last = self.dragging[1], self.dragging[#self.dragging]
+	if mp.get_time() - last.time > 0.05 then return 0 end
+	for i = #self.dragging - 1, 1, -1 do
+		local drag = self.dragging[i]
+		if last.time - drag.time > 0.03 then return ((drag.y - last.y) / ((last.time - drag.time) / 0.03)) * 10 end
 	end
-	return #self.drag_data < 2 and 0 or (last_y - self.drag_data[1].y) / (self.drag_data[1].time - last_time)
+	return #self.dragging < 2 and 0 or ((first.y - last.y) / ((first.time - last.time) / 0.03)) * 10
 end
 
 function Menu:on_global_mbtn_left_up()
-	if self.proximity_raw == 0 and not self.is_dragging then
+	local was_dragging = self.dragging and #self.dragging > 10
+
+	if self.proximity_raw == 0 and not was_dragging then
 		self:select_item_below_cursor()
 		self:open_selected_item({preselect_submenu_item = false})
 	end
-	if self.is_dragging then
-		local velocity = self:fling_velocity()
-		if math.abs(velocity) > 50 then
-			self.fling = {
-				y = self.current.scroll_y, velocity = velocity,
-				time = self.drag_data[#self.drag_data].time
-			}
-		else self.fling = nil end
+
+	if was_dragging then
+		local distance = self:fling_distance()
+		if math.abs(distance) < 50 then self.fling = nil
+		else self.fling = {distance = distance, updated = self.dragging[#self.dragging].time} end
 	else self.fling = nil end
-	self.is_dragging = false
-	self.drag_data = nil
+
+	self.dragging = nil
 end
 
 
 function Menu:on_global_mouse_move()
 	self.mouse_nav = true
-	if self.drag_data then
-		if self.is_dragging or math.abs(cursor.y - self.drag_data[1].y) >= 5 then
-			self.is_dragging = true
-			local distance = cursor.y - self.drag_data[#self.drag_data].y
-			if distance ~= 0 then
-				self:scroll_to(self.current.scroll_y - distance)
-				self.drag_data[#self.drag_data + 1] = {y = cursor.y, time = mp.get_time()}
-			end
-		end
+	if self.dragging then
+		local distance = self.dragging[#self.dragging].y - cursor.y
+		if distance ~= 0 then self:scroll_by(distance) end
+		self.dragging[#self.dragging + 1] = {y = cursor.y, time = mp.get_time()}
 	end
-	if self.proximity_raw == 0 or self.is_dragging then self:select_item_below_cursor()
+	if self.proximity_raw == 0 or self.dragging then self:select_item_below_cursor()
 	else self.current.selected_index = nil end
 	request_render()
 end
 
 function Menu:on_wheel_up()
-	self:scroll_to(self.current.scroll_y - self.scroll_step * 3)
+	self:scroll_by(self.scroll_step * -3)
 	self:on_global_mouse_move() -- selects item below cursor
 	request_render()
 end
 
 function Menu:on_wheel_down()
-	self:scroll_to(self.current.scroll_y + self.scroll_step * 3)
+	self:scroll_by(self.scroll_step * 3)
 	self:on_global_mouse_move() -- selects item below cursor
 	request_render()
 end
@@ -2187,13 +2182,12 @@ end
 
 function Menu:render()
 	if self.fling then
-		local v_start = self.fling.velocity
-		local t = state.render_last_time - self.fling.time
-		local lambda, deceleration = 7, v_start >= 0 and -2 or 2
-		local v = v_start * math.exp(-lambda * t) + deceleration * t
-		local y = self.fling.y + v_start / lambda * (1 - math.exp(-lambda * t)) + deceleration * t * t / 2
-		self:scroll_to(y)
-		if (v_start >= 0 and v >= 0) or (v_start < 0 and v < 0) then request_render()
+		local time_delta, distance = state.render_last_time - self.fling.updated, self.fling.distance
+		local new_distance = math[distance > 0 and 'floor' or 'ceil'](distance * (0.4 ^ (time_delta / 0.1)))
+		if math.abs(new_distance) > 0 then
+			self:scroll_by(self.fling.distance - new_distance)
+			self.fling.distance, self.fling.updated = new_distance, state.render_last_time
+			request_render()
 		else self.fling = nil end
 	end
 

```

tomasklaen:
I've noticed a couple issues with my patch, so I've updated it in case you've started doing something with it in between.

christoph-heinrich:
It starts off faster then I'm dragging by a factor of roughly 3, which is probably where that "responsiveness" comes from.

Detecting a fling based on the number of move updates seems very error prone. On a high dpi display with high touch sampling rate it's easy to get a bunch of tiny position changes while taping. Going off of the total change in position seems more robust to me.

I've incorporated most of your changes, but it's still velocity based. However I've removed the velocity calculation in `Menu:render()` and instead use the change of the current position for stopping. It now also checks for scrolling bounds and stops early if those are reached. Lambda has been adjusted to behave pretty much identically to what you have.

tomasklaen:
> It starts off faster then I'm dragging by a factor of roughly 3

YES, that's by design. I know that your implementation is _physically accurate_, but it feels like I'm throwing rocks underwater. In my patch, I can use fast flicks as a navigation method to scroll further, while yours feels more like just an animation effect to smoothly end dragging. I have to go **really** fast for the flick to cover any significant distance, which is frustrating.

Mine might not be accurate, but it just feels better and is more usable. Just as it also doesn't make sense for holding a jump button to make your character jump higher, but it makes better games.

> Lambda has been adjusted to behave pretty much identically to what you have.

Nope, it's still like throwing rocks underwater. Just compare fast flicks on some long menu with this: [uosc.zip](https://github.com/tomasklaen/uosc/files/9842288/uosc.zip)

tomasklaen:
Oh, and it seems like you forgot to round the new position.

christoph-heinrich:
Are you serious? You want the speed to fall off really quickly, but also scroll far.
My initial intent was to make it feel similar to android, where you can scroll very far with a fast flick because it doesn't fall off super fast, but that didn't fall fast enough for you.... ffs.

> Oh, and it seems like you forgot to round the new position.

That was intentional, I tried rounding and it leads to a more abrupt stop at the end.

christoph-heinrich:
There, those numbers behave very similar to how it is on android from my testing. A good fling can reach a little over 100 items.

There is however one critical flaw... we don't know the exact time when velocity reaches 0 and I can't find an analytical way of calculating it. An iterative approach would work, but it doesn't really seem to be a problem in practice so I'm not sure it's worth the extra code.

tomasklaen:
> You want the speed to fall off really quickly, but also scroll far.

Yes. You achieve that by higher initial velocity, and fast decay. Just like in my patch.

Right now we are here:

The_most_INCREDIBLE_Curling_Shots_at_Beijing_2022- t0gjtnXlEOk

tomasklaen:
All right, an executive call: I have some other plans for this, and I need it to be a simple tweakable and maintainable distance based animation with dumb easing implementation. Just apply this patch so I can merge it and leave this stunlock:

```diff
diff --git a/scripts/uosc.lua b/scripts/uosc.lua
index c900bd4..c56e190 100644
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -143,6 +143,10 @@ function table_shallow_copy(table)
 	return result
 end
 
+--[[ EASING FUNCTIONS ]]
+
+function ease_out_quart(x) return 1 - ((1 - x) ^ 4) end
+
 --[[ OPTIONS ]]
 
 local defaults = {
@@ -1701,7 +1705,7 @@ function Menu:init(data, callback, opts)
 	---@type {y: integer, time: number}[]
 	self.drag_data = nil
 	self.is_dragging = false
-	---@type {y: number, velocity: number, time: number, menu: MenuStack}
+	---@type {y: number, distance: number, time: number, menu: MenuStack}
 	self.fling = nil
 
 	self:update(data)
@@ -2064,14 +2068,14 @@ function Menu:on_global_mbtn_left_down()
 	end
 end
 
-function Menu:fling_velocity()
+function Menu:fling_distance()
 	local first, last = self.drag_data[1], self.drag_data[#self.drag_data]
 	if mp.get_time() - last.time > 0.05 then return 0 end
 	for i = #self.drag_data - 1, 1, -1 do
 		local drag = self.drag_data[i]
-		if last.time - drag.time > 0.03 then return (last.y - drag.y) / (drag.time - last.time) end
+		if last.time - drag.time > 0.03 then return ((drag.y - last.y) / ((last.time - drag.time) / 0.03)) * 10 end
 	end
-	return #self.drag_data < 2 and 0 or (last.y - first.y) / (first.time - last.time)
+	return #self.drag_data < 2 and 0 or ((first.y - last.y) / ((first.time - last.time) / 0.03)) * 10
 end
 
 function Menu:on_global_mbtn_left_up()
@@ -2080,11 +2084,11 @@ function Menu:on_global_mbtn_left_up()
 		self:open_selected_item({preselect_submenu_item = false})
 	end
 	if self.is_dragging then
-		local velocity = self:fling_velocity()
-		if math.abs(velocity) > 50 then
+		local distance = self:fling_distance()
+		if math.abs(distance) > 50 then
 			self.fling = {
-				y = self.current.scroll_y, velocity = velocity,
-				time = self.drag_data[#self.drag_data].time, menu = self.current
+				y = self.current.scroll_y, distance = distance,
+				time = self.drag_data[#self.drag_data].time, menu = self.current,
 			}
 		end
 	end
@@ -2185,16 +2189,10 @@ end
 
 function Menu:render()
 	if self.fling then
-		local menu, v_start = self.fling.menu, self.fling.velocity
 		local time_delta = state.render_last_time - self.fling.time
-		local decel, lambda = v_start >= 0 and -300 or 300, 1 -- half life period = -ln(0.5) / lambda
-		local p1, p2 =  math.exp(-lambda * time_delta), decel * time_delta
-		local y = self.fling.y + v_start / lambda * (1 - p1) + p2 * time_delta / 2
-		local v = v_start * p1 + p2
-		if v * v_start >= 0 and y >= 0 and y <= menu.scroll_height then
-			self:scroll_to(y, menu)
-			request_render()
-		else self.fling = nil end
+		local progress = ease_out_quart(math.min(time_delta / 0.5, 1))
+		self:scroll_to(round(self.fling.y + self.fling.distance * progress), self.fling.menu)
+		if progress < 1 then request_render() else self.fling = nil end
 	end
 
 	local ass = assdraw.ass_new()

```

christoph-heinrich:
> Right now we are here:

Just like every other interface with touch support.

> Just apply this patch so I can merge it and leave this stunlock

I really don't like how it starts off with higher speed, but here you go your majesty.

I'll still try that jerk thing. Maybe a velocity curve that looks something like this would be more to your liking.
image

Would be pretty easy to provide an interface where you tell it how long the animation should last and how sharp the drop off should be and it calculates the necessary jerk and acceleration on it's own.

christoph-heinrich:
Here, in case you want to try it.
`a_stop` is the acceleration when velocity hits 0.
`t_stop` is how long the animation lasts.

Everything before `t_delta =` could be done at the initialization of the fling.
```diff
diff --git a/scripts/uosc.lua b/scripts/uosc.lua
index c56e190..a4d6a7a 100644
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -1705,7 +1705,7 @@ function Menu:init(data, callback, opts)
 	---@type {y: integer, time: number}[]
 	self.drag_data = nil
 	self.is_dragging = false
-	---@type {y: number, distance: number, time: number, menu: MenuStack}
+	---@type {y: number, velocity: number, time: number, menu: MenuStack}
 	self.fling = nil
 
 	self:update(data)
@@ -2078,17 +2078,27 @@ function Menu:fling_distance()
 	return #self.drag_data < 2 and 0 or ((first.y - last.y) / ((first.time - last.time) / 0.03)) * 10
 end
 
+function Menu:fling_velocity()
+	local first, last = self.drag_data[1], self.drag_data[#self.drag_data]
+	if mp.get_time() - last.time > 0.05 then return 0 end
+	for i = #self.drag_data - 1, 1, -1 do
+		local drag = self.drag_data[i]
+		if last.time - drag.time > 0.03 then return (last.y - drag.y) / (drag.time - last.time) end
+	end
+	return #self.drag_data < 2 and 0 or (last.y - first.y) / (first.time - last.time)
+end
+
 function Menu:on_global_mbtn_left_up()
 	if self.proximity_raw == 0 and not self.is_dragging then
 		self:select_item_below_cursor()
 		self:open_selected_item({preselect_submenu_item = false})
 	end
 	if self.is_dragging then
-		local distance = self:fling_distance()
-		if math.abs(distance) > 50 then
+		local velocity = self:fling_velocity()
+		if math.abs(velocity) > 50 then
 			self.fling = {
-				y = self.current.scroll_y, distance = distance,
-				time = self.drag_data[#self.drag_data].time, menu = self.current,
+				y = self.current.scroll_y, velocity = velocity,
+				time = self.drag_data[#self.drag_data].time, menu = self.current
 			}
 		end
 	end
@@ -2189,10 +2199,16 @@ end
 
 function Menu:render()
 	if self.fling then
-		local time_delta = state.render_last_time - self.fling.time
-		local progress = ease_out_quart(math.min(time_delta / 0.5, 1))
-		self:scroll_to(round(self.fling.y + self.fling.distance * progress), self.fling.menu)
-		if progress < 1 then request_render() else self.fling = nil end
+		local menu, v_start = self.fling.menu, self.fling.velocity
+		local a_stop, t_stop = v_start >= 0 and -6000 or 6000, math.min(math.abs(v_start) / 3000, 0.5)
+		local jerk = 2 * (a_stop * t_stop + v_start) / (t_stop ^ 2)
+		local a = (-v_start - jerk * t_stop ^ 2 / 2) / t_stop
+		local t_delta = math.min(state.render_last_time - self.fling.time, t_stop)
+		local y = self.fling.y + v_start * t_delta + a * t_delta ^ 2 / 2 + jerk * t_delta ^ 3 / 6
+		self:scroll_to(y, menu)
+		if t_delta < t_stop then
+			request_render()
+		else self.fling = nil end
 	end
 
 	local ass = assdraw.ass_new()
```

PR #335, opened 2022-10-18
fix: pause while dragging the timeline

Natural-Harmonia-Gropius:
When you slowly move the timeline to left, it automatically moves to right.

For a better experience, I would like to temporally disable the pause indicator. How can I do this?

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -2670,6 +2670,7 @@ function PauseIndicator:init()
 	self.opacity = 0
 
 	mp.observe_property('pause', 'bool', function(_, paused)
+		if Elements.timeline.pressed then return end
 		if options.pause_indicator == 'flash' then
 			if self.paused == paused then return end
 			self:flash()
@@ -2857,6 +2858,8 @@ function Timeline:clear_thumbnail() mp.commandv('script-message-to', 'thumbfast'
 
 function Timeline:on_mbtn_left_down()
 	self.pressed = true
+	self.pressed_pause = state.pause
+	mp.set_property_native('pause', true)
 	self:set_from_cursor()
 end
 function Timeline:on_prop_duration() self:decide_enabled() end
@@ -2866,7 +2869,10 @@ function Timeline:on_prop_fullormaxed() self:update_dimensions() end
 function Timeline:on_display() self:update_dimensions() end
 function Timeline:on_mouse_leave() self:clear_thumbnail() end
 function Timeline:on_global_mbtn_left_up()
-	self.pressed = false
+	if self.pressed then
+		mp.set_property_native('pause', self.pressed_pause)
+		self.pressed = false
+	end
 	self:clear_thumbnail()
 end
 function Timeline:on_global_mouse_leave()


hooke007:
It would break user's expection.
i.e. Dragging the timeline in paused stated.

hooke007:
> It's how YouTube does it

I checked youtube, it would never force to play after dragging.

tomasklaen:
I've implemented it once and then removed it because it produced an annoying UX that was optimizing for 1% use case:

You need to pause on mouse down, which makes quick seeks by simply clicking the timeline annoying as they always cycle the pause. And quick seeks are the majority of seeks. Not many people drag the timeline.

And there's no way around it. Pausing in move handler is too late, as video will progress between mouse down and move which always feels wrong/buggy.

I'm afraid you just have to pause the video before dragging, as there's no way to have a good quick seeking and a good drag seeking at the same time (it is impossible to know what the user is intending to do when they mouse down on the timeline). And since quick seeking is vastly more common, that has the priority.

Natural-Harmonia-Gropius:
> Dragging the timeline in paused stated.

Will fix.

> Not many people drag the timeline.

Indeed, especially now with thumbnails.

> which makes quick seeks by simply clicking the timeline annoying as they always cycle the pause.

Youtube is work same as this, the pause button in the bottom left corner will have a shape change.
I personally don't find fast pause/play annoying, but the indicator is, so I want to disable it temporally.

Natural-Harmonia-Gropius:
The current code is not elegant, but I am satisfied with the functionality.

tomasklaen:
All right I've tried it and it actually feels nice. Dunno why I decided to not do it when I tried it before.

Anyway, here's a cleaner solution:

```diff
diff --git a/scripts/uosc.lua b/scripts/uosc.lua
index c6b98bc..1d645d4 100644
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -2670,6 +2670,7 @@ function PauseIndicator:init()
        self.opacity = 0

        mp.observe_property('pause', 'bool', function(_, paused)
+               if Elements.timeline.pressed then return end
                if options.pause_indicator == 'flash' then
                        if self.paused == paused then return end
                        self:flash()
@@ -2855,14 +2856,10 @@ function Timeline:set_from_cursor(fast)
 end
 function Timeline:clear_thumbnail() mp.commandv('script-message-to', 'thumbfast',
'clear') end

-local temp_pause = true
 function Timeline:on_mbtn_left_down()
-       local temp_pause_indicator = options.pause_indicator
-       options.pause_indicator = 'manual'
-       temp_pause = mp.get_property_native('pause')
-       mp.set_property_native('pause', true)
-       mp.add_timeout(0.05, function() options.pause_indicator = temp_pause_indicator end)
        self.pressed = true
+       self.pressed_pause = state.pause
+       mp.set_property_native('pause', true)
        self:set_from_cursor()
 end
 function Timeline:on_prop_duration() self:decide_enabled() end
@@ -2872,11 +2869,10 @@ function Timeline:on_prop_fullormaxed() self:update_dimensions() end
 function Timeline:on_display() self:update_dimensions() end
 function Timeline:on_mouse_leave() self:clear_thumbnail() end
 function Timeline:on_global_mbtn_left_up()
-       local temp_pause_indicator = options.pause_indicator
-       options.pause_indicator = 'manual'
-       mp.set_property_native('pause', temp_pause)
-       mp.add_timeout(0.05, function() options.pause_indicator = temp_pause_indicator end)
-       self.pressed = false
+       if self.pressed then
+               mp.set_property_native('pause', self.pressed_pause)
+               self.pressed = false
+       end
        self:clear_thumbnail()
 end
 function Timeline:on_global_mouse_leave()
```

Apply it and if you don't see any issues I'll merge.

Natural-Harmonia-Gropius:
Applied, works fine.

PR #336, opened 2022-10-18
fix: support checking URL without slashes

Hill-98:
It now supports checking URLs like this: `magnet:?`


tomasklaen:
You have an operator precedence error there. You need to wrap the ORed matches in parentheses or we get a "path:match is nil" error when path is not a string.

Hill-98:
fixed

PR #338, opened 2022-10-25
feat: better file sorting

christoph-heinrich:
There is currently this PR for mpv https://github.com/mpv-player/mpv/pull/10779 and that algorithm would be a good improvement here too.

This maintains the current behavior + better sorting of numbers.

The for loop always only ran once, so I've removed the loop and just let it check the first character.

I actually tried fixing up the loop to iterate over each character (until a difference is found), but that led to a directory with lots of `N.mp3` files having `11.mp3` before `9.mp3` with all other numbers being correctly sorted, and the function always returned that 11 is not smaller then 9, so I have no idea how it ended up like that and I didn't want to spend hours trying to debug this.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -486,25 +486,27 @@ local file_order_comparator = (function()
 		default_order = 21
 	end
 
+	-- Alphanum sorting for humans in Lua
+	-- http://notebook.kulchenko.com/algorithms/alphanumeric-natural-sorting-for-humans-in-lua
+	local function padnum(d)
+		local dec, n = d:match("(%.?)0*(.+)")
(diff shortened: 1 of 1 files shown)


tomasklaen:
The reason why I didn't go for natural sorting from the start is because how slow it is. But I agree that overall it'd be a net positive as the need for better sorting of normal folders is way more common than fast sorting of big ones.

But, I have a test folder with 15k files and its sorting went from 65ms to 700ms. So it's slightly more than 10 times slower. Do you see any way to make it faster?

tomasklaen:
Maybe some other time. I want to split the file already.

PR #341, opened 2022-10-25
perf: faster filename sorting

christoph-heinrich:
I measured a ~15 times speedup on that 1700 files folder I made out of dyphires config.

Numbers with leading zeros are now sorted after identical numbers without leading zeros.

ref. https://github.com/tomasklaen/uosc/pull/338#issuecomment-1290137738

### DIFF
--- a/scripts/uosc/lib/menus.lua
+++ b/scripts/uosc/lib/menus.lua
@@ -167,7 +167,7 @@ function open_file_navigation_menu(directory_path, handle_select, opts)
 	end
 
 	-- Files are already sorted
-	table.sort(directories, file_order_comparator)
+	sort_filenames(directories)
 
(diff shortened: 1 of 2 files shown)


christoph-heinrich:
Oops, didn't notice the `"`. I've run the formatter now.

dyphire:
> Numbers with leading zeros are now sorted after identical numbers without leading zeros.

This sort does not seem to be correct. It is different from the sort in File Explorer and `ls` command printed.

christoph-heinrich:
That is because it used to append the length of the other string, but that's not possible when creating the formatted string before the sorting... Maybe we could remove appending the length and then add an equality and length check in the comparison. I'll give that a try.

PR #342, opened 2022-10-25
fix: restore leading zero sorting

christoph-heinrich:
Restore the sorting behavior from before #341

Also comes with a small performance improvement, but hardly worth mentioning. (~5%)

Ref. https://github.com/tomasklaen/uosc/pull/341#issuecomment-1290954070


dyphire:
Looks like it still needs improvement.
image
image

hooke007:
The explorer
Snipaste_2022-10-25_19-42-07

christoph-heinrich:
> Looks like it still needs improvement.

You're right, had the wrong index and it just happened to work in my test case. Should be fine now.

dyphire:
Now it works well.

PR #344, opened 2022-10-25
perf: menu directory opening speedup

christoph-heinrich:
I've been digging into why opening large directories takes so long, because I know it's not because the width estimation takes that long, and sorting is now also pretty quick. It turns out there is a lot of work being done that really doesn't have to be done.

The vast majority of the time is spent in [this line](https://github.com/tomasklaen/uosc/blob/614ccf6b3ae5af35e2ed0b0a2884ba43c02755d3/scripts/uosc/lib/utils.lua#L204).

The time this one loop takes when opening my test folder went from ~880ms down to ~35ms.
From a quick look it seems that  `filename` and `extension` from that table is never actually used, and getting rid of those and `dot_split` lowers it even further to ~15ms.

I don't know how robust what I have here actually is, I haven't done a lot of testing and I doubt it will ever get into a mergeable state. The main point here is to show how much faster it could be. This only touches files and not directories, so directories containing lots of directories will still be slow.

You know a lot better then me which kinds of paths can come from where and what's actually needed.

### DIFF
--- a/scripts/uosc/lib/menus.lua
+++ b/scripts/uosc/lib/menus.lua
@@ -197,8 +197,16 @@ function open_file_navigation_menu(directory_path, handle_select, opts)
 	end
 
 	for _, file in ipairs(files) do
-		local serialized = serialize_path(utils.join_path(directory.path, file))
-		if serialized then
+		if not is_protocol(file) then
+			local dot_split = split(file, '%.')
+			local serialized = {
+				path = utils.join_path(directory.path, file),
+				is_root = false,
+				dirname = directory.path,
+				basename = file,
+				filename = #dot_split > 1 and table.concat(itable_slice(dot_split, 1, #dot_split - 1), '.') or file,
+				extension = #dot_split > 1 and dot_split[#dot_split] or nil,
+			}
 			serialized.is_file = true
 			items[#items + 1] = {title = serialized.basename, value = serialized}
 		end


tomasklaen:
From my testing, serialization and sorting is always a fraction of file list retrieval time:

```
[uosc] read dirs(0) 0.22972
[uosc] read files(5000) 0.233056
[uosc] files sorted 0.016261
[uosc] dirs sorted 0
[uosc] dirs serialized 0
[uosc] files serialized 0.052353999999999
```

So the most speedup we can get here would be to figure out a faster interface, or at the very least to not do it in 2 calls. Currently I'm doing `utils.readdir(path, 'dirs')` and `utils.readdir(path, 'files')`, because I don't know about a different API that allows me to retrieve the list AND differentiate between dirs and files.

But if serialization is slow, we might as well just save full paths as strings, and only create utils to get directory/file name from them as needed.

The extension is needed for `allowed_types` filter to work, but I see that is done with external `get_extension()` utility. That can also be sped up by converting it into `has_extension(path, extension)` that only checks if the end of the path matches extension.

christoph-heinrich:
> The extension is needed for allowed_types filter to work

Those are already filtered long before the serialization.

Without this change it looks like this for me.
```
[uosc] read dirs 0 0.018008000000002 
[uosc] read files 1700 0.019266000000002 
[uosc] files sorted 0.022632000000002 
[uosc] dirs sorted 2.0000000020559e-06 
[uosc] dirs serialized 5.7000000005303e-05 
[uosc] files serialized 0.893081
```

With this change (including getting rid of `dot_split`, `filename` and `extension`)
```
[uosc] read dirs 0 0.017938999999998 
[uosc] read files 1700 0.019305000000003 
[uosc] files sorted 0.024330000000006 
[uosc] dirs sorted 3.999999989901e-06 
[uosc] dirs serialized 3.0000000066366e-06 
[uosc] files serialized 0.016493000000011
```

tomasklaen:
In the commit above I've managed to drop the time to read directories by half. I think this is now as fast as it gets.

There's still a bug (introduced I think by your recent PRs, or maybe it was there before, not sure) where traversing directories makes the menu loose it's title, which should be `{current_directory_name}/`, and instead it only displays `/`. But don't have time to look into it today anymore.

dyphire:
@tomasklaen was right. I can reproduce it.
This is introduced by https://github.com/tomasklaen/uosc/pull/346. We carelessly ignored it in the test.

PR #346, opened 2022-10-26
perf: optimize serialize_path

christoph-heinrich:
This is not just a performance improvement, but a bug fix as well. I used to not be possible to get to the root directory on linux, but now that works.

The loop I measured in #344 went from ~880ms down to ~40ms with this. There is still a lot of room for improvement, but this is already huge.

Splitting `working_path` with `last_index_of()` had identical performance to `utils.split_path()`, but using the built-in function is less code.

Creating `filename` and `extension` now only contributes ~1ms to the total time instead of ~20ms.

### DIFF
--- a/scripts/uosc/lib/std.lua
+++ b/scripts/uosc/lib/std.lua
@@ -37,6 +37,20 @@ function split(str, pattern)
 	return list
 end
 
+---Get index of the last appearance of `sub` in `str`
+---@param str string
+---@param sub string
+---@return integer|nil
+function string.last_index_of(str, sub)
+	local sub_length = #sub
+	for i = #str, 1, -1 do
+		for j = 1, sub_length do
+			if str:byte(i + j - 1) ~= sub:byte(j) then break end
+			if j == sub_length then return i end
+		end
+	end
+end
+
 ---@param itable table
 ---@param value any
 ---@return integer|nil
--- a/scripts/uosc/lib/utils.lua
+++ b/scripts/uosc/lib/utils.lua
@@ -198,23 +198,17 @@ function serialize_path(path)
 	if not path or is_protocol(path) then return end
 
 	local normal_path = normalize_path(path)
-	-- normalize_path() already strips slashes, but leaves trailing backslash
-	-- for windows drive letters, but we don't need it here.
-	local working_path = normal_path:sub(#normal_path) == '\\' and normal_path:sub(1, #normal_path - 1) or normal_path
-	local parts = split(working_path, '[\\/]+')
-	local basename = parts and parts[#parts] or working_path
-	local dirname = #parts > 1
-		and table.concat(itable_slice(parts, 1, #parts - 1), state.os == 'windows' and '\\' or '/')
-		or nil
-	local dot_split = split(basename, '%.')
+	local dirname, basename = utils.split_path(normal_path)
+	if basename == '' then dirname = nil end
+	local dot_i = basename:last_index_of('.')
 
 	return {
 		path = normal_path,
 		is_root = dirname == nil,
 		dirname = dirname,
 		basename = basename,
-		filename = #dot_split > 1 and table.concat(itable_slice(dot_split, 1, #dot_split - 1), '.') or basename,
-		extension = #dot_split > 1 and dot_split[#dot_split] or nil,
+		filename = dot_i and basename:sub(1, dot_i - 1) or basename,
+		extension = dot_i and basename:sub(dot_i + 1) or nil,
 	}
 end
 


dyphire:
There is a bug: the root directory to drive cannot be returned on Windows.

christoph-heinrich:
I used to not be able to go to the root directory, but it works now.
I don't have a windows installation to test with.
Maybe removing `working_path` and just using `normal_path` will work?
```diff
diff --git a/scripts/uosc/lib/utils.lua b/scripts/uosc/lib/utils.lua
index 821ec8e..b0ae977 100644
--- a/scripts/uosc/lib/utils.lua
+++ b/scripts/uosc/lib/utils.lua
@@ -198,10 +198,7 @@ function serialize_path(path)
 	if not path or is_protocol(path) then return end
 
 	local normal_path = normalize_path(path)
-	-- normalize_path() already strips slashes, but leaves trailing backslash
-	-- for windows drive letters, but we don't need it here.
-	local working_path = normal_path:sub(#normal_path) == '\\' and normal_path:sub(1, #normal_path - 1) or normal_path
-	local dirname, basename = utils.split_path(working_path)
+	local dirname, basename = utils.split_path(normal_path)
 	local dot_i = last_index_of(basename, '.')
 
 	return {

```

dyphire:
> I used to not be able to go to the root directory, but it works now. I don't have a windows installation to test with. Maybe removing `working_path` and just using `normal_path` will work?

No, the patch does not work.

christoph-heinrich:
Then I can only suggest adding in some `print` statements to see where things go wrong. Maybe `utils.split_path()` doesn't handle that properly?

dyphire:
> Then I can only suggest adding in some `print` statements to see where things go wrong. Maybe `utils.split_path()` doesn't handle that properly?

only need to apply the following handle to work normally:
```lua
local dirname, basename = utils.split_path(working_path)
if dirname:match('^[A-Z]+:$') then dirname = nil end
```

christoph-heinrich:
So it's not possible to switch drives on windows? Didn't know that.
I removed `working_path` and added a backslash to the match instead.

dyphire:
>  added a backslash to the match instead.

This is wrong. Matching the backslash will cause failure. The match I give can work normally.

dyphire:
The `working_path` seems necessary. This match does not work properly after you delete it. That's how it works.

christoph-heinrich:
Did you test the version that had the backslash in the match? Because I don't understand how the backslash removal is necessary but also matching on backslash breaks things because there is no backslash to match...

dyphire:
> Did you test the version that had the backslash in the match? Because I don't understand how the backslash removal is necessary but also matching on backslash breaks things because there is no backslash to match...

Why do you think it works?

https://user-images.githubusercontent.com/61936050/198110769-35a579e5-9d78-4147-9f8c-bbe346298706.mp4

As I said at the beginning, matching backslashes doesn't work as expected.

christoph-heinrich:
Oh that's what you mean. I encountered what I think was basically the same failure on linux, so I solved it in the same way. Does that work?

dyphire:
> Oh that's what you mean. I encountered what I think was basically the same failure on linux, so I solved it in the same way. Does that work?

It works.

christoph-heinrich:
Now I wonder if that will also work, because I noticed that basename is an empty string when serializing root. Would be a bit simpler.
``` diff
diff --git a/scripts/uosc/lib/utils.lua b/scripts/uosc/lib/utils.lua
index 208f73f..c78334f 100644
--- a/scripts/uosc/lib/utils.lua
+++ b/scripts/uosc/lib/utils.lua
@@ -199,8 +199,7 @@ function serialize_path(path)
 
 	local normal_path = normalize_path(path)
 	local dirname, basename = utils.split_path(normal_path)
-	if state.os == 'windows' and normal_path:match('^[A-Z]:\\$') then dirname = nil
-	elseif normal_path == '/' then dirname = nil end
+	if basename == '' then dirname = nil end
 	local dot_i = last_index_of(basename, '.')
 
 	return {
```

dyphire:
> Now I wonder if that will also work, because I noticed that basename is an empty string when serializing root. Would be a bit simpler.

Ah, it really works. Well done.

scripts/uosc/lib/std.lua
```diff
@@ -37,6 +37,20 @@ function split(str, pattern)
 	return list
 end
 
+---Get index of the last appearance of `sub` in `str`
+---@param str string
+---@param sub string
+---@return integer|nil
+function string.last_index_of(str, sub)
```
tomasklaen, line 44: Don't extend `string`, this might cause issues. Make it `string_last_index_of*()`.

tomasklaen: Nvm, I'll do it alter, I want to refactor the directory structure to support mpv 0.32 now.

PR #347, opened 2022-10-27
perf: optimize normalize_path

christoph-heinrich:
Does not convert backslashes to slashes or vice versa anymore.
Backslash is a valid character on Unix filesystems, so converting it to slash can actually break things.
I don't think it's possible to encounter a slash in a path on windows, except maybe from the `default_directory` option. If that's really a problem, we should sanitize that option at startup instead of always substituting in `normalize_path`.

This together with #346 brings the measured time on that file loop on my test directory down to ~18ms, getting surprisingly close to the time from #344.

### DIFF
--- a/scripts/uosc/lib/std.lua
+++ b/scripts/uosc/lib/std.lua
@@ -17,6 +17,21 @@ function serialize_rgba(rgba)
 	}
 end
 
+---Trim any `char` from the end of the string
+---@param str string
+---@param char string
+---@return string
+function string.trim_end(str, char)
+	local char, end_i = char:byte(), nil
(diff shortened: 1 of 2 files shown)


scripts/uosc/lib/std.lua
```diff
@@ -17,6 +17,21 @@ function serialize_rgba(rgba)
 	}
 end
 
+---Trim any `char` from the end of the string
+---@param str string
+---@param char string
+---@return string
+function string.trim_end(str, char)
```
tomasklaen, line 24: Don't extend `string`, this might cause issues. Just make it `trim_end`.

PR #349, opened 2022-10-27
fix: directory name when browsing files

christoph-heinrich:
This also addresses displaying `/` as path separator on windows.

@dyphire can you please test that it works fine on windows?

ref. https://github.com/tomasklaen/uosc/pull/344#issuecomment-1293393638

### DIFF
--- a/scripts/uosc_shared/lib/menus.lua
+++ b/scripts/uosc_shared/lib/menus.lua
@@ -149,7 +149,7 @@ end
 ---@param handle_select fun(path: string): nil
 ---@param opts NavigationMenuOptions
 function open_file_navigation_menu(directory_path, handle_select, opts)
-	directory = serialize_path(directory_path)
+	directory = serialize_path(normalize_path(directory_path))
 	opts = opts or {}
 
 	if not directory then
@@ -184,11 +184,12 @@ function open_file_navigation_menu(directory_path, handle_select, opts)
 
 	local items_start_index = #items + 1
 
+	local path_separator = path_separator(directory.path)
 	for _, dir in ipairs(directories) do
 		local serialized = serialize_path(utils.join_path(directory.path, dir))
 		if serialized then
 			serialized.is_directory = true
-			items[#items + 1] = {title = serialized.basename, value = serialized, hint = '/'}
+			items[#items + 1] = {title = serialized.basename, value = serialized, hint = path_separator}
 		end
 	end
 
@@ -214,7 +215,7 @@ function open_file_navigation_menu(directory_path, handle_select, opts)
 	end
 
 	local menu_data = {
-		type = opts.type, title = opts.title or directory.basename .. '/', items = items,
+		type = opts.type, title = opts.title or directory.basename .. path_separator, items = items,
 		on_open = opts.on_open, on_close = opts.on_close,
 	}
 
--- a/scripts/uosc_shared/lib/utils.lua
+++ b/scripts/uosc_shared/lib/utils.lua
@@ -145,33 +145,90 @@ function opacity_to_alpha(opacity)
 	return 255 - math.ceil(255 * opacity)
 end
 
--- Ensures path is absolute and remove trailing slashes/backslashes.
+do
+	local os_separator = state.os == 'windows' and '\\' or '/'
+
+	-- Get appropriate path separator for the given path.
+	---@param path string
+	---@return string
+	function path_separator(path)
+		return path:sub(1, 2) == '\\\\' and '\\' or os_separator
+	end
+
+	-- Joins paths with the OS aware path separator or UNC separator.
+	---@param p1 string
+	---@param p2 string
+	---@return string
+	function utils.join_path(p1, p2)
+		return p1 .. path_separator(p1) .. p2
+	end
+end
+
+-- Check if path is absolute.
 ---@param path string
+---@return boolean
+function is_absolute(path)
+	if path:sub(1, 2) == '\\\\' then return true
+	elseif state.os == 'windows' then return path:find('^%a+:') ~= nil
+	else return path:sub(1, 1) == '/' end
+end
+
+-- Ensure path is absolute.
+---@param path string
+---@return string
+function ensure_absolute(path)
+	if is_absolute(path) then return path end
+	return utils.join_path(state.cwd, path)
+end
+
+-- Remove trailing slashes/backslashes.
+---@param path string
+---@return string
+function trim_trailing_separator(path)
+	path = trim_end(path, path_separator(path))
+	if state.os == 'windows' then
+		-- Drive letters on windows need trailing backslash
+		if path:sub(#path) == ':' then return path .. '\\' end
+		return path
+	else
+		if path == '' then return '/' end
+		return path
+	end
+end
+
+-- Ensures path is absolute, remove trailing slashes/backslashes.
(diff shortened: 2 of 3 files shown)


dyphire:
It works for me.

tomasklaen:
It now displays the whole directory path, it should only display parent directory name.

dyphire:
I agree that it should only show the parent directory name.

christoph-heinrich:
> It now displays the whole directory path, it should only display parent directory name.

It does for meI I would have never committed it if I noticed an unintended change in behavior. Intentional changes would be mentioned in the OP.
image

As far as I can replicate what happens on windows without modifying mpv itself, everything seems to be working fine.

dyphire:
> It does for me. I would have never committed it if I noticed an unintended change in behavior. Intentional changes would be mentioned in the OP.

The whole directory path will be displayed in my test.

https://user-images.githubusercontent.com/61936050/198337786-4fcb058b-2629-4c2c-9732-7f69bb407c43.mp4

christoph-heinrich:
There is a `/` at the end of the path. That's weird and it makes sens that it would show the whole path then. I'll have to find out where that comes from and then it should work.

tomasklaen:
The `/` is I think added manually when creating the menu title to signal that this is a directory name.

christoph-heinrich:
> The `/` is I think added manually when creating the menu title to signal that this is a directory name.

But I've added a windows check to make it a `\` instead.

No matter what I try, I can't reproduce this. I even modified `mp_basename()` to make `utils.split_path` behave the same as on windows, but everything seems fine and I can't find where a `/` would be added on windows.
```
[uosc] normal_path 'C:\intel\bla'
[uosc] dirname 'C:\intel\'
[uosc] basename 'bla'
```
```
[uosc] normal_path 'C:\intel'
[uosc] dirname 'C:\'
[uosc] basename 'intel'
```
```
[uosc] normal_path 'C:\'
[uosc] dirname nil
[uosc] basename 'C:'
```

christoph-heinrich:
> The clue might be that it only happens when navigating up the the directory tree via the pre-inserted first `..` item. See video above.

Yes I noticed that, but I don't see anything special happening there. It calls `serialize_path()` on the `dirname` of directory and then has the result as value for that item, but when I use any of the generated `dirname`s as input, I still get good outputs.

tomasklaen:
The issue is that `utils.join_path()` always joins with `/`.

Here's a log of traversing to `foo` subdirectory and back:

```
[uosc] directory_path F:\Downloads\test\
[uosc] directory.basename test
[uosc] directory_path F:\Downloads\test/foo
[uosc] directory.basename foo
[uosc] directory_path F:\Downloads\test/
[uosc] directory.basename F:\Downloads\test
```

The initial `directory_path` also shouldn't have the `\` at the end. Probably directory path not being normalized inside `open-file` script biding anymore.

christoph-heinrich:
Omg you're right, I've looked at the implementation and while it differentiates between OS on some things, it always joins with `/`.
I've asked on the IRC and it's by design, so we'll have to make windows work with both `/` and `\`.
AMM also pointed me to his path utils ported from python https://github.com/TheAMM/mpv_script_libs/blob/master/helpers.lua#L28-L368

tomasklaen:
No, we need to normalize to current OS separator, and ensure stripped trailing slashes. I remember there being a lot of gotchas that the old code was working around, and we'll have to stumble into them again I guess.

One of them I can recall is that the paths we are using to open files absolutely have to match how that path would look like when system opens the file (doubleclick/enter) using mpv as a default player, otherwise the `write-watch-later-config` and  `--save-position-on-quit` won't work, as they use the hash of the path as an ID. Meaning, only single slashes of current platform, no mixed ones, or trailing ones.

christoph-heinrich:
`trim_end()` now supports multiple characters and windows trims both backslash and slash.

tomasklaen:
Or we could have our own `join_path` that uses current platform separator, and then have a separate expensive `super_normalize_path()` that converts slashes to current platform, de-duplicates them, and ensures no trailing ones. This `super_normalize_path()` would then be used only when passing a path to some file opening command.

christoph-heinrich:
Overwriting the utils function with a dumb joiner made it way faster lol.

Does this work fine on windows?

I haven't yet tackled deduplication. I think we'll have to be extra careful because of `'^\\\\'` paths.

christoph-heinrich:
Deduplication has now also been implemented. Still unsure about joining UNC paths.

christoph-heinrich:
If this works on windows for you on windows, I'd say it's good enough.
Supporting everything specified in https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats would be madness. (also I'm pretty sure we used to not support all of that)

scripts/uosc_shared/lib/utils.lua
```diff
@@ -164,10 +181,31 @@ function normalize_path(path)
 		end
 		return path
 	else
-		return trim_end(path, '/')
+		path = trim_end(path, path:match('^\\\\') and '\\' or '/')
+		if path == '' then path = '/' end
+		return path
 	end
```
tomasklaen, line 187: Extract this if/else into `trim_trailing_separator()` function so that it can be used in `super_normalize_path()` instead of calling `normalize_path()` there, which does a lot of duplicitous work.

Also, wouldn't `path:sub(1, 2) == '\\\\'` be faster than `:match('^\\\\')`?

tomasklaen (commented):
Use `super_normalize_path` instead of `normalize_path` everywhere in the codebase that isn't speed critical.

tomasklaen:
Also a better nomenclature would be to rename these as:

- `normalize_path` -> `normalize_path_lite`
- `super_normalize_path` -> `normalize_path`

And document that `_lite` is needed because it's faster.

scripts/uosc_shared/lib/utils.lua
christoph-heinrich: I've timed it and `:find()` is faster then `:match()` in theory because it doesn't have to create a string for returning (when there are no captures), but I couldn't really measure a difference in our case.
`:sub()` did bring a small but measurable improvement.

scripts/uosc_shared/lib/menus.lua
```diff
@@ -184,11 +184,12 @@ function open_file_navigation_menu(directory_path, handle_select, opts)
 
 	local items_start_index = #items + 1
 
+	local path_seperator = state.os == 'windows' and '\\' or '/'
```
tomasklaen, line 187: `seperator` -> `separator`

scripts/uosc_shared/lib/utils.lua
```diff
@@ -145,33 +145,90 @@ function opacity_to_alpha(opacity)
...
 end
 
--- Ensures path is absolute and remove trailing slashes/backslashes.
----@param path string
-function normalize_path(path)
-	if not path or is_protocol(path) then return path end
+do
+	local os_separator = state.os == 'windows' and '\\' or '/'
+
+	-- Get appropriate path separator for the given path
```
tomasklaen, line 151: This and others are function descriptions, so use dots at the end of sentences.

tomasklaen (commented):
Just a couple typos.

christoph-heinrich:
The path separator displayed at the end of the current directory and as a hint now use the `path_separator()` function.

PR #358, opened 2022-10-31
feat: disable previewing while dragging the timeline

hooke007:


tomasklaen:
This will wait until #353 is resolved.

PR #362, opened 2022-11-02
fix: infinite margin on initialization

### LINKED ISSUE #361: Margins crash console.lua
(on the latest mpv commit)

`update_margins()` sets `osc-margins`, and console.lua will crash if any of them are not numbers (by performing math with them).

https://github.com/mpv-player/mpv/blob/fd91776207cd6b8d82e09765e0c6e2ec5e23d881/player/lua/console.lua#L85-L89

This should be fixed on the console.lua side too, the crash potential has been there forever and only doesn't happen because scripts don't set the bottom margin to an invalid value, even before https://github.com/mpv-player/mpv/commit/fd91776207cd6b8d82e09765e0c6e2ec5e23d881.  
uosc should add guards for this.

This can (and always does on init for me) return 'inf' for bottom. `tonumber()` in console.lua turns it into nil.

https://github.com/tomasklaen/uosc/blob/5f5bf4b842b7d3d3f1887780c8e5085d88ef100c/scripts/uosc.lua#L385

```
[console] stack traceback:
[console]       @console.lua:178: in function 'update'
[console]       @console.lua:224: in function 'set_active'
[console]       @console.lua:792: in function 'fn'
[console]       mp.defaults:230: in function 'fn'
[console]       mp.defaults:65: in function 'handler'
[console]       mp.defaults:380: in function 'handler'
[console]       mp.defaults:510: in function 'call_event_handlers'
[console]       mp.defaults:552: in function 'dispatch_events'
[console]       mp.defaults:503: in function <mp.defaults:502>
[console]       [C]: in ?
[console]       [C]: in ?
[console] Lua error: @console.lua:178: attempt to perform arithmetic on upvalue 'global_margin_bottom' (a nil value)
```

christoph-heinrich:
Closes #361

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -379,6 +379,8 @@ end
 
 -- Notifies other scripts such as console about where the unoccupied parts of the screen are.
 function update_margins()
+	if display.height == 0 then return end
+
 	-- margins are normalized to window size
 	local timeline, top_bar, controls = Elements.timeline, Elements.top_bar, Elements.controls
 	local bottom_y = controls and controls.enabled and controls.ay or timeline.ay


po5:
Can confirm it fixes the issue. Strangely, after more testing it would only happen on certain files.

PR #363, opened 2022-11-02
fix: wait for dimensions to properly initialize

po5:
On some video files with chapters, on some launches, uosc is stuck in a busy loop.  
It happens even with `text_width_estimation=no`.  
After much trial and error, I tracked it down to the initial `serialize_chapters()`, and this block of code.

https://github.com/tomasklaen/uosc/blob/d9e5853542f017d2cd092e199ec2e03f919e5bf1/scripts/uosc_shared/lib/text.lua#L264-L271

These are the values of different variables in the loop.

```
[uosc] char_count -nan 
[uosc] width -nan 
[uosc] width_ratio -nan 
[uosc] next_char_count -nan
```

Displaying values of the `fit_on_screen()` call just before the loop reveals the issue.

```
[uosc] max_size 0 
[uosc] horizontal true
```

`osd_width` and `osd_height` are `0`. `osd-dimensions` is initially reported as 0,0.

This was the core issue behind #361. It no longer happens, and #362 is not needed.  
It seems like mpv now always broadcasts an `osd-dimensions` update on launch.

```
[uosc] dim 0 0 
[auto_profiles] Applying auto profile: Anime 
 (+) Video --vid=1 (*) (h264 640x480 23.976fps)
 (+) Audio --aid=1 --alang=jpn (*) (vorbis 2ch 48000Hz)
 (+) Subs  --sid=1 --slang=eng (*) 'English Subtitles' (ass)
     Subs  --sid=2 --slang=eng 'English Subtitles (No Songs)' (ass)
AO: [pipewire] 48000Hz stereo 2ch floatp
VO: [gpu] 640x480 yuv420p
AV: 00:00:00 / 00:24:15 (0%) A-V:  0.000
[uosc] dim 958 719
```

Wait until the values actually get initialized.


po5:
The same issue was happening in `measure_bounds()`, and could be triggered by registering a custom menu on initialization as described in the README.

christoph-heinrich:
What are the values measure_bounds() got when things were 0?
The way it is now it will simply return nil 5 times (because of the loop in `normalized_text_width`) and then say the text has 0 width.

po5:
`osd_width, osd_height = mp.get_osd_size()` would set both of those "global" variables back to 0, leading to the same loop issue next time we enter `character_width()`.

christoph-heinrich:
Looks like mpv uses [1280x720](https://github.com/mpv-player/mpv/blob/806dddc7d9273c4f520f03d0e87ef7992cf563f9/sub/osd_libass.c#L607-L611) as fallback, which contrary to the comment does not entirely line up with the [command description](https://mpv.io/manual/master/#command-interface-osd-overlay). Maybe we should also fall back to that resolution?

po5:
If you want to change the fallback, only change the initial value of both variables.  
Keeping the current value when we get something unexpected is better than forcing a bogus one.

christoph-heinrich:
> If you want to change the fallback, only change the initial value of both variables.

Changing the initial values won't help getting around that `return nil`.

https://github.com/tomasklaen/uosc/pull/363/files#diff-9799130f0b2d705fa9fe5b5d2661af25411f9d58ee6967952c3ad82c97c8847cR129-R132

po5:
Does the loop in `normalized_text_width()` actually serve any purpose as it is now? Would make more sense to have some delay between tries, which would be enough to catch the real dimensions.

christoph-heinrich:
> Does the loop in `normalized_text_width()` actually serve any purpose as it is now? Would make more sense to have some delay between tries, which would be enough to catch the real dimensions.

It's purpose is to catch cases where the text is too long and gets clipped by libass, so it reduces the text size and tries again.

tomasklaen:
I actually don't understand how this prevents #361 😶.

Also, doesn't this produce bad text measurements (and catches them) during the time the osd-dimensions are 0/nil?

Shouldn't we just defer the whole uosc initialization until we have some valid osd-dimensions?

po5:
Looking at it again I'm not sure how it fixes it either.

We have another call to `mp.get_osd_size()` in `update_display_dimensions()` which may cause the same issue?  
And that *does* update `display.width` & `display.height`.

tomasklaen:
Is this still necessary? Didn't it get fixed by recent fix of osd_* variable scoping issues?

tomasklaen:
Seems like this is not needed anymore. Let me know otherwise.

Issue #49, opened 2021-04-25
Crash when using ab loop (playing livestream)


TheLongLife:
When I'm watching a live stream with streamlink and I press "l" to create the ab loop it crashes the UI.

```
[  32.652][d][cplayer] Run command: ab-loop, flags=73, args=[]
[  32.652][v][cplayer] Set property: ab-loop-a -> 1
[  32.658][w][uosc] 
[  32.658][w][uosc] stack traceback:
[  32.658][w][uosc] 	[string "C:/ProgramData/chocolatey/lib/mpv.install/too..."]:1576: in function 'maybe'
[  32.658][w][uosc] 	[string "C:/ProgramData/chocolatey/lib/mpv.install/too..."]:2129: in function 'cb'
[  32.658][w][uosc] 	mp.defaults:351: in function 'process_timers'
[  32.658][w][uosc] 	mp.defaults:520: in function 'dispatch_events'
[  32.658][w][uosc] 	mp.defaults:493: in function <mp.defaults:492>
[  32.658][w][uosc] 	[C]: at 0x00a92de0
[  32.658][w][uosc] 	[C]: at 0x00a917f0
[  32.658][f][uosc] Lua error: [string "C:/ProgramData/chocolatey/lib/mpv.install/too..."]:1576: bad argument #1 to 'ipairs' (table expected, got nil)
[  32.658][d][uosc] Exiting...
```

vitormathews:
This issue has been solved? Because it's still happening to me, with local files.

tomasklaen:
Do you have anything to share? Like uosc version (top of the uosc.lua file) or error logs?

vitormathews:
uosc 4.5.0 - 2022-Dec-07

This is what the terminal shows when I press l (set/clear A-B loop points):
```
AO: [wasapi] 48000Hz stereo 2ch float
VO: [gpu] 1920x1080 yuv420p10

Id     Name            PSJobTypeName   State         HasMoreData     Location             Command
--     ----            -------------   -----         -----------     --------             -------
1      mpvminloop      BackgroundJob   Running       True            localhost            ...
(Paused) AV: 00:00:33.867 / 00:25:57.077 (2%) A-V:  0.000
[uosc]
[uosc] stack traceback:
[uosc]  ...ortable_config/scripts/uosc_shared/elements/Timeline.lua:301: in function 'maybe'
[uosc]  ...207236/portable_config/scripts/uosc_shared/lib/utils.lua:513: in function 'cb'
[...]

PR #389, opened 2022-12-13
fix: A-B loop crash

christoph-heinrich:
Fixes https://github.com/tomasklaen/uosc/issues/49#issuecomment-1348022899

This made me realize that a-b loop indicators aren't clickable even though they look the same as chapter indicators. Do we want that?

### DIFF
--- a/scripts/uosc_shared/elements/Timeline.lua
+++ b/scripts/uosc_shared/elements/Timeline.lua
@@ -298,8 +298,8 @@ function Timeline:render()
 				end
 			end
 
-			if state.ab_loop_a and state.ab_loop_a > 0 then draw_chapter(state.ab_loop_a) end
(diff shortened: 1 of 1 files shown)


tomasklaen:
Yeah, I think anyone would expect them to be clickable.

I think we should just insert A-B points as chapters into chapters table during serialization/normalization. Can you think of any issues with that?

christoph-heinrich:
It could interfere with chapter ranges, ending some ranges sooner then they are supposed to. Also the shown chapter when hovering would be influenced by that.

It's probably best if the indicators become their own elements, then A-B points can simply be added as such elements and automatically be handled correctly. It would also remove the special click handling from timeline and controls.
Rendering order is important, so they can't simply have a `render()` method themselves and instead need to be rendered in chronological order except for the hovered one.

tomasklaen:
Sounds right. But this is gonna be a bit more work, so dunno when I'll have time for it.

Issue #354, opened 2022-10-30
Sometimes menu text is not rendered


NaiveInvestigator:
![image](https://user-images.githubusercontent.com/72654596/198879043-30caa4e0-5c4c-4085-82af-118a6518a007.png)

This doesn't always happen, only sometimes randomly but also kinda frequently. Also the elements that dissappear are also random at times.

This is my menu:
```
### UOSC MENU

menu				script-binding uosc/menu
MOUSE_BTN2			script-binding uosc/menu
o				script-binding open-file-dialog					#! Open File
#				script-binding SmartCopyPaste/paste				#! Open From Clipboard
CTRL+Y				script-binding youtube-search					#! Search In YouTube
#				script-binding uosc/chapters					#! Navigation > Chapters
#				script-binding seek_to/toggle-seeker				#! Navigation > Seek To Time
#				script-binding SmartCopyPaste/copy-specific			#! Navigation > Copy Timestamp
#				script-binding SmartCopyPaste/paste-specific			#! Navigation > Paste Timestamp
CTRL+n				add audio-delay +0.10						#! Audio > Delay > Increase Audio Delay
CTRL+N				add audio-delay -0.10						#! Audio > Delay > Decrease Audio Delay
#				set audio-delay 0						#! Audio > Delay > Reset Audio Delay
#				script-binding uosc/audio-device				#! Audio > Devices
#				script-binding uosc/audio					#! Audio > Tracks
#				script-binding uosc/video					#! Video > Tracks
D				cycle deinterlace						#! Video > Toggle Deinterlace
#				set window-scale 0.5						#! Video > Window Scale > 50%
#				set window-scale 0.8						#! Video > Window Scale > 80%
#				set window-scale 1.0						#! Video > Window Scale > 100%
#				set window-scale 1.2						#! Video > Window Scale > 120%
#				set window-scale 1.5						#! Video > Window Scale > 150%
#				set window-scale 2.0						#! Video > Window Scale > 200%
# 				vf toggle vflip							#! Video > Flip > Vertical
#				vf toggle hflip							#! Video > Flip > Horizontal
n				script-binding autosubsync/autosubsync-menu			#! Subtitles > Automatic Synchronisation
#				script-binding autosub/download_subs				#! Subtitles > Download For Current Video > Language 1 (English)
#				script-binding autosub/download_subs2				#! Subtitles > Download For Current Video > Language 2 (Bengali)
[...]

PR #393, opened 2022-12-17
fix text width measurement

dyphire:
https://github.com/tomasklaen/uosc/pull/383 introduced a strange text width measurement bug: ref https://github.com/tomasklaen/uosc/issues/354#issuecomment-1336307900, https://github.com/tomasklaen/uosc/issues/354#issuecomment-1336351521, this can fix it.

### DIFF
--- a/scripts/uosc_shared/lib/text.lua
+++ b/scripts/uosc_shared/lib/text.lua
@@ -127,7 +127,7 @@ text_osd.compute_bounds, text_osd.hidden = true, true
 ---@type integer, integer
 local osd_width, osd_height = 100, 100
 mp.observe_property('osd-dimensions', 'native', function (_, dim)
-	if dim then update_osd_resolution(dim.w, dim.h) end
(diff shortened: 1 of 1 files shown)


christoph-heinrich:
Didn't you effectively just inline that function? I don't see how that results in different behavior. I mean it's great if it actually fixes things, I just don't understand how.

dyphire:
> Didn't you effectively just inline that function? I don't see how that results in different behavior. I mean it's great if it actually fixes things, I just don't understand how.

I also don't understand why the text width measurement behavior of the two is different, but it does fix it for me.

tomasklaen:
Maybe because the function is defined before the variables, it's setting them to global? 🤷‍♂️

Try to move all variable declarations to the top.

christoph-heinrich:
> Maybe because the function is defined before the variables, it's setting them to global? :man_shrugging: 

Makes sense, that must be it.

tomasklaen:
closed by #396

Issue #390, opened 2022-12-14
[Feature Requests] Delete file to Recycle Bin (Windows 10) | Option to change font and timeline bar colors | Show play/pause Indicator in timeline | Hide hour count for files less than 1 hour | Option to show media title, author, description, etc. in the UI


vitormathews:
About the last one: I miss being able to see file metadata in MPV like in MPC-HC/MPC-BE:
![image](https://user-images.githubusercontent.com/51209382/207672195-cfb1b10a-f617-405b-b745-ba1018272a2f.png)
      
They could be shown inside the timeline, or below the file name in the top bar, I don't know.

And about the "hide hour count" request, I have the milliseconds enabled, so I found it very visual polluted to have a count like this: 00:00:00.000
![image](https://user-images.githubusercontent.com/51209382/207677801-4b5c37ad-b75e-4322-9c62-16e5214d91a7.png)

- Windows 10 21H2 19044.2251
- uosc 4.5.0 - 2022-Dec-07
- mpv 0.35.0-30-g4574dd5dc6

dyphire:
> Delete file to Recycle Bin

This request is possible, refer to [delete-current-file.lua](https://github.com/stax76/mpv-scripts/blob/main/delete-current-file.lua)

> Option to change font and timeline bar colors

dup https://github.com/tomasklaen/uosc/issues/294

> Show play/pause Indicator in timeline

You can add `cycle:play_arrow:pause:no=pause/yes=play_arrow?play/pause` to the options of `controls` in your `script-opts/uosc.conf` to implement it.

> Hide hour count for files less than 1 hour

I haven't seen any other players behave like this. It's just your personal preference.

> Option to show media title, author, description, etc. in the UI

The media-title is already displayed in the UI, see: https://github.com/tomasklaen/uosc/blob/bfaccc8f66874d04339e9b1e515b61450b157aec/script-opts/uosc.conf#L119-L121

As for author and description, mpv does not seem to provide this metadata. I also can't see the benefits of further swelling the UI.

vitormathews:
> This request is possible, refer to [delete-current-file.lua](https://github.com/stax76/mpv-scripts/blob/main/delete-current-file.lua)

[...]

PR #394, opened 2022-12-17
feat: add option to delete file to Recycle Bin

dyphire:
ref: https://github.com/tomasklaen/uosc/issues/390

### DIFF
--- a/script-opts/uosc.conf
+++ b/script-opts/uosc.conf
@@ -183,6 +183,10 @@ media_types=3g2,3gp,aac,aiff,ape,apng,asf,au,avi,avif,bmp,dsf,f4v,flac,flv,gif,h
 subtitle_types=aqt,ass,gsub,idx,jss,lrc,mks,pgs,pjs,psb,rt,slt,smi,sub,sup,srt,ssa,ssf,ttxt,txt,usf,vt,vtt
 # Default open-file menu directory
 default_directory=~/
+# Move files to trash (recycle bin) when deleting files. Dependencies:
+# - Linux: `sudo apt install trash-cli`
+# - MacOS: `brew install trash`
+use_trash=no
 
 # Convers some common chapter types into chapter range indicators.
 # Instead of displaying the start of the chapter as a diamond icon on top of the
--- a/scripts/uosc_shared/lib/utils.lua
+++ b/scripts/uosc_shared/lib/utils.lua
@@ -380,7 +380,29 @@ end
 -- `status:number(<0=error), stdout, stderr, error_string, killed_by_us:boolean`
 ---@param path string
 function delete_file(path)
-	local args = state.os == 'windows' and {'cmd', '/C', 'del', path} or {'rm', path}
+	if state.os == 'windows' then
+		if options.use_trash then
+            local ps_code = [[
+				Add-Type -AssemblyName Microsoft.VisualBasic
+				[Microsoft.VisualBasic.FileIO.FileSystem]::DeleteFile('__path__', 'OnlyErrorDialogs', 'SendToRecycleBin')
+			]]
+
+            local escaped_path = string.gsub(path, "'", "''")
+            escaped_path = string.gsub(escaped_path, "’", "’’")
+            escaped_path = string.gsub(escaped_path, "%%", "%%%%")
+            ps_code = string.gsub(ps_code, "__path__", escaped_path)
+		    args = { 'powershell', '-NoProfile', '-Command', ps_code }
+		else
+			args = { 'cmd', '/C', 'del', path }
+		end
+	else
+		if options.use_trash then
+			--On Linux and Macos the app trash-cli/trash must be installed first.
+		    args = { 'trash', path }
+	    else
+		    args = { 'rm', path }
+		end
+	end
 	return mp.command_native({
 		name = 'subprocess',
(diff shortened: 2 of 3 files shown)


christoph-heinrich:
With this code being very similar to [delete-current-file.lua](https://github.com/stax76/mpv-scripts/blame/main/delete-current-file.lua#L59-L68) and GPL-2 technically not being compatible with GPL-3, we should probably ask for permission to make sure we're not stepping on anyone's toes.

@stax76 Are you fine with this?

dyphire:
> With this code being very similar to [delete-current-file.lua](https://github.com/stax76/mpv-scripts/blame/main/delete-current-file.lua#L59-L68) and GPL-2 technically not being compatible with GPL-3, we should probably ask for permission to make sure we're not stepping on anyone's toes.

It's my mistake, @stax76 Can you tell me your opinion?

tomasklaen:
1. Rename the option to `use_trash` so it's more universal, as only windows calls it a recycle bin.
2. I don't see any condition branch for macOS, so it's going to use `{ 'trash', path }`. It doesn't seem like mac has a `trash` command.
3. If there are some tools/packages required to be installed, it should be documented in the config file. But I'd be surprised if this couldn't be done without that. Can't we just `mv` files to some standard trash directory on linux/mac?
4. If the option is enabled, it should never fall back to deletion. If we are trying to delete on a platform where we can't implement it, it should log error message instead.

And make the condition branching easier to follow. So instead of:

```lua
if win and use_trash then
elseif win then
elseif not win and ...
```

do:

```lua
if win then
    if use_trash then
        -- trashing
    else
        -- deleting
    end
else if mac then
    -- mac if it differs from linux
else
    -- linux
end
```

christoph-heinrich:
> 3\. If there are some tools/packages required to be installed, it should be documented in the config file. But I'd be surprised if this couldn't be done without that. Can't we just `mv` files to some standard trash directory on linux/mac?

`mv` doesn't create the necessary metadata file. We could create it ourselves, the format looks pretty simple.

stax76:
It's fine, no problem.

dyphire:
> 3\. If there are some tools/packages required to be installed, it should be documented in the config file. But I'd be surprised if this couldn't be done without that. Can't we just `mv` files to some standard trash directory on linux/mac?

I'm not familiar with linux and macos, so I'm not sure how to implement it.

christoph-heinrich:
The spec for linux can be found here: https://specifications.freedesktop.org/trash-spec/trashspec-latest.html.
Before moving we'd have to also check for potential file conflicts in trash in addition to creating the metadata file. It's certainly easier to rely on trash-cli instead of implementing it ourselves.

tomasklaen:
All right, 3rd party it is. So we need to find the macos alternative to `trash-cli`, use it in its branch, and document it in default config file.

Or if there is no alternative (I'd be surprised), document that the option is not available on macos.

dyphire:
> All right, 3rd party it is. So we need to find the macos alternative to `trash-cli`, use it in its branch, and document it in default config file.
> 
> Or if there is no alternative (I'd be surprised), document that the option is not available on macos.

In fact, I found that there are third-party commands of `trash` on the macos. ref https://formulae.brew.sh/formula/trash

Done.

script-opts/uosc.conf
```diff
@@ -183,6 +183,9 @@ media_types=3g2,3gp,aac,aiff,ape,apng,asf,au,avi,avif,bmp,dsf,f4v,flac,flv,gif,h
 subtitle_types=aqt,ass,gsub,idx,jss,lrc,mks,pgs,pjs,psb,rt,slt,smi,sub,sup,srt,ssa,ssf,ttxt,txt,usf,vt,vtt
 # Default open-file menu directory
 default_directory=~/
+# Move to the trash(recycle bin) when deleting file
+# On Linux and Macos the app trash-cli/trash must be installed first.
```
tomasklaen, line 187: Change the comment to:

```
# Move files to trash (recycle bin) when deleting files. Dependencies:
# - Linux: `sudo apt install trash-cli`
# - MacOS: `brew install trash`
```

scripts/uosc_shared/lib/utils.lua
```diff
@@ -380,7 +380,29 @@ end
...
+
+            local escaped_path = string.gsub(path, "'", "''")
+            escaped_path = string.gsub(escaped_path, "’", "’’")
+            escaped_path = string.gsub(escaped_path, "%%", "%%%%")
+            ps_code = string.gsub(ps_code, "__path__", escaped_path)
+		    args = { 'powershell', '-NoProfile', '-Command', ps_code }
+		else
+			args = { 'cmd', '/C', 'del', path }
+		end
+	elseif state.os ~= 'windows' then
```
tomasklaen, line 398: You don't need `elseif` here. This is just an `else`.

dyphire: Done

script-opts/uosc.conf
dyphire: Done

tomasklaen:
If you enable `use_trash` in your `script-opts/uosc.conf`, all deletion done by uosc will go to trash.

On linux/mac this depends on trash-cli/trash being installed.

PR #396, opened 2022-12-22
fix: declare variable before function using it

christoph-heinrich:
Moved the declaration above it's usage as noted in https://github.com/tomasklaen/uosc/pull/393#issuecomment-1356759785

This did improve width estimation for me in some situations where I thought that's just some inaccuracy from estimating it with individual characters, but after this change its better now. Based on that I'd expect it to also improve https://github.com/tomasklaen/uosc/issues/354#issuecomment-1336351521.

I've checked all other variables in the file and they are all fine.

### DIFF
--- a/scripts/uosc_shared/lib/text.lua
+++ b/scripts/uosc_shared/lib/text.lua
@@ -40,6 +40,11 @@ local same_width_blocks = {
 	{0x31350, 0x323AF}, -- CJK Unified Ideographs Extension H	 3 TIP	Han
 }
 
+local width_length_ratio = 0.5
+
+---@type integer, integer
+local osd_width, osd_height = 100, 100
+
 ---Get byte count of utf-8 character at index i in str
 ---@param str string
 ---@param i integer?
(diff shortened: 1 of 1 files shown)


tomasklaen:
Any reason to not just move all variable declarations to the top of the file and have functions below? Or at the very least to initialize them at the top if they can't be set there.

As it is we are setting up a timer for someone to add a new function above one of the variables it needs and cause the same issue.

christoph-heinrich:
I like declaring variables close to where they are used. `text_osd` and `bounds_opts` are only used right below their declaration. `osd_width` and `osd_height` ended up being used in a bunch of places, so moving those to the top might make sense. I can also move all of them to the top if you would prefer that.

tomasklaen:
I'd move `osd_*` to the top, and scope the variables that are needed only in one function to that context. For example:

```lua
local foo
do
  local bar = true
  foo = function()
    -- use bar
  end
end
```

or maybe:

```lua
local foo = (function()
  local bar = true
  return function()
    -- use bar
  end
end)()
```

christoph-heinrich:
Good idea with the scoping.

scripts/uosc_shared/lib/text.lua
```diff
@@ -320,37 +332,40 @@ local function opts_scale_factor(opts)
...
+text_width = nil
+do
+	---@type {[boolean]: {[boolean]: {[string|number]: {[1]: number, [2]: integer}}}} | {[boolean]: {[string|number]: {[1]: number, [2]: integer}}}
+	local width_cache = {}
+
+	---Calculate width of text with the given opts
+	---@param text string|number
+	---@return number
+	---@param opts {size: number; bold?: boolean; italic?: boolean}
+	text_width = function(text, opts)
```
tomasklaen, line 344: You can just do

```lua
function text_width(text, opts)
```

here, it'll also make a global function, and you can delete the `text_width = nil` above.

christoph-heinrich: done

PR #399, opened 2022-12-25
feat: add destination_time option and deprecate total_time

### LINKED ISSUE #397: Add option to show time remaining (as opposed to playtime remaining)
Right now time in the right corner is shown as a playtime remaining. Meaning, it is divided by speed.

I think it would be good improvement to add an option to show time remaining not divided by speed.

If you agree, I am willing to open a PR.

sarmong:
Resolves #397

I used `playtime-remaining` and `time-remaining` because this is how the are described in mpv documentation.

### DIFF
--- a/script-opts/uosc.conf
+++ b/script-opts/uosc.conf
@@ -163,7 +163,7 @@ background=000000
 background_text=ffffff
 # Use only bold font weight throughout the whole UI
 font_bold=no
-# One of 'total', 'playtime-remaining' (scaled by the current speed), 'time-remaining' (remaining length of file)
+# One of `total`, `playtime-remaining` (scaled by the current speed), `time-remaining` (remaining length of file)
 destination_time=playtime-remaining
 # Display sub second fraction in timestamps up to this precision
 time_precision=0
--- a/script-opts/uosc.conf
+++ b/script-opts/uosc.conf
@@ -163,8 +163,8 @@ background=000000
 background_text=ffffff
 # Use only bold font weight throughout the whole UI
 font_bold=no
-# Show total time instead of time remaining
-total_time=no
+# One of 'total', 'playtime-remaining' (scaled by the current speed), 'time-remaining' (remaining length of file)
+destination_time=playtime-remaining
 # Display sub second fraction in timestamps up to this precision
 time_precision=0
 # Display stream's buffered time in timeline if it's lower than this amount of seconds, 0 to disable
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -113,6 +113,10 @@ if options.pause_on_click_shorter_than > 0 and options.click_threshold == 0 then
 	msg.warn('`pause_on_click_shorter_than` is deprecated. Use `click_threshold` and `click_command` instead.')
 	options.click_threshold = options.pause_on_click_shorter_than
 end
+if options.total_time and options.destination_time == 'playtime-remaining' then
+	msg.warn('`total_time` is deprecated. Use `destination_time` instead.')
+	options.destination_time = 'total'
+end
 -- Ensure required environment configuration
 if options.autoload then mp.commandv('set', 'keep-open-pause', 'no') end
 -- Color shorthands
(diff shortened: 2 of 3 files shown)


scripts/uosc.lua
```diff
@@ -1,5 +1,5 @@
---[[ uosc 4.5.0 - 2022-Dec-07 | https://github.com/tomasklaen/uosc ]]
-local uosc_version = '4.5.0'
+--[[ uosc 4.6.0 - 2022-Dec-25 | https://github.com/tomasklaen/uosc ]]
+local uosc_version = '4.6.0'
```
tomasklaen, line 2: Don't bump version.

scripts/uosc.lua
```diff
@@ -369,9 +370,13 @@ function update_human_times()
 		state.time_human = format_time(state.time)
 		if state.duration then
 			local speed = state.speed or 1
-			state.duration_or_remaining_time_human = format_time(
-				options.total_time and state.duration or ((state.time - state.duration) / speed)
-			)
+      if options.total_time or options.destination_time == "total" then
```
tomasklaen, line 373: Since `total_time` is deprecated, code should only work with `destination_time`. Put this:

```lua
if options.total_time and options.destination_time == 'playtime-remaining' then
	options.destination_time = 'total'
end
```

into `scripts/uosc.lua` among other option normalizations (starts at line 108).

And here only check for `options.destination_time`. Start the elseif conditions by checking for the most common `playtime-remaining`, then `total`, and fall to `time-remaining`.

And use single quotes when defining strings.

scripts/uosc.lua
```diff
@@ -369,9 +370,13 @@ function update_human_times()
...
+        state.duration_or_remaining_time_human = format_time(state.duration)
```
tomasklaen, line 374: Rename `state.duration_or_remaining_time_human` everywhere in the codebase to `state.destination_time_human`.

script-opts/uosc.conf
```diff
@@ -163,8 +163,8 @@ background=000000
 background_text=ffffff
 # Use only bold font weight throughout the whole UI
 font_bold=no
-# Show total time instead of time remaining
-total_time=no
+# One of 'total', 'playtime-remaining' (scaled by the current speed), 'time-remaining' (remaining length of file)
```
tomasklaen, line 166: Use \` to wrap values for consistency with the rest of the document.

tomasklaen (changes requested):

scripts/uosc.lua
sarmong: Should I add deprecation warning?

sarmong: >  Start the elseif conditions by checking for the most common playtime-remaining, then total, and fall to time-remaining.

My thinking was that if a user sets invalid value, it will set to a default value of `playtime-remaining`

tomasklaen: > Should I add deprecation warning?

Sure.

> My thinking was that if a user sets invalid value, it will set to a default value of playtime-remaining

This would be better handled during normalization. So:

```lua
if options.total_time and options.destination_time == 'playtime-remaining' then
	msg.warn('`total_time` is deprecated. Use `destination_time` instead.')
	options.destination_time = 'total'
elseif not itable_index_of({'total', 'playtime-remaining', 'time-remaining'}, options.destination_time) then
	options.destination_time = 'playtime-remaining'
end
```

sarmong: done

scripts/uosc.lua
```diff
@@ -369,11 +376,15 @@ function update_human_times()
 		state.time_human = format_time(state.time)
 		if state.duration then
 			local speed = state.speed or 1
-			state.duration_or_remaining_time_human = format_time(
-				options.total_time and state.duration or ((state.time - state.duration) / speed)
-			)
+			if options.destination_time == 'playtime_remaining' then
```
tomasklaen, line 379: It's `playtime-remaining`, not `playtime_remaining`.

tomasklaen (approved):

Issue #421, opened 2023-01-10
Make Buffered Time change according to playback speed


NaiveInvestigator:
Buffered Time does not change according to playback speed, but time remaining and others change with according to speed tho.

Minor issue, just thought you'd want to know.

christoph-heinrich:
I don't know if this should change or not, because it's just a matter of perspective, but I've always understood the `buffered_time_threshold` option to be "the largest number that should be shown".
If it's actually intended as such, then we should scale the time before the `buffered_time < options.buffered_time_threshold` check, because otherwise speeds > 1 won't show numbers up to the configured `buffered_time_threshold` and speeds < 1 will show numbers bigger then that.

christoph-heinrich:
Since I'm not the only one to interpret it this way, I've made a PR to try it out.

PR #423, opened 2023-01-10
tweak: buffered time threshold based on displayed value

christoph-heinrich:
This adjusts `buffered_time_threshold` to behave the way described in https://github.com/tomasklaen/uosc/issues/421#issuecomment-1377084759.


tomasklaen:
Makes sense. But since this is the only use for this variable, let's instead of creating a new `buffered_time_scaled`, just rename the original to `buffered_playtime`, and scale it as we're defining it at line 229.

christoph-heinrich:
done

PR #424, opened 2023-01-10
fix: buffered time respects minus in time remaining

christoph-heinrich:
Without this the buffered time would get too close to the remaining time.


tomasklaen:
I've noticed that too some time ago but forgot about it :x Thx!

PR #429, opened 2023-01-14
Add integration with mpv-quality-menu

microraptor:
Integrates @christoph-heinrich excellent [mpv-quality-menu](https://github.com/christoph-heinrich/mpv-quality-menu) script.

I added one if-statement that delegates `script-binding uosc/stream-quality` to `script-binding quality_menu/video_formats_toggle`, if the special value `stream_quality_options=script` is set for the already existing config option.

This simplifies the setup for the script a lot, because the user doesn't have to manually modify the control button and menu entry for the built-in quality-menu.

There is some context at https://github.com/tomasklaen/uosc/pull/102

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,9 @@ Editions menu. Editions are different video cuts available in some mkv files.
 
 #### `stream-quality`
 
-Switch stream quality. This is just a basic re-assignment of `ytdl-format` mpv property from predefined options (configurable with `stream_quality_options`) and video reload, there is no fetching of available formats going on.
+Switch stream quality.
+
+This is just a basic re-assignment of `ytdl-format` mpv property from predefined options (configurable with `stream_quality_options`) and video reload. There is no fetching of available formats going on. However, if the special value `stream_quality_options=script` is set, uosc will try to open the menu of quality-menu.lua instead, which fetches the available formats. For installation see [mpv-quality-menu](https://github.com/christoph-heinrich/mpv-quality-menu).
 
 #### `open-file`
 
--- a/script-opts/uosc.conf
+++ b/script-opts/uosc.conf
@@ -175,7 +175,9 @@ autohide=no
 pause_indicator=flash
 # Screen dim when stuff like menu is open, 0 to disable
 curtain_opacity=0.5
-# Sizes to list in stream quality menu
+# Sizes (pixel height) to list in stream quality menu.
+# The special value "stream_quality_options=script" will try to open the menu of quality-menu.lua, which fetches the available formats.
+# "script" as a value requires this script to be installed: https://github.com/christoph-heinrich/mpv-quality-menu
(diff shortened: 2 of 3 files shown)


tomasklaen:
This doesn't make sense to me. mpv-quality-menu already [has uosc integration](https://github.com/christoph-heinrich/mpv-quality-menu#installation) built in.

If user wants to display it's quality menu, they should just use mpv-quality-menu script bindings, not uosc's. If they want buttons in control bar, they can just copy & paste the ones from its readme instructions into uosc `controls` option. It makes no sense for uosc to re-route its `stream-quality` binding to a 3rd party script.

For a user, there's no difference between configuring `stream_quality_options=script` in uosc (what does `script` even mean? too much implicity), or configuring their `inputs.conf` and `uosc.conf:controls` to just use mpv-quality-menu commands instead. Just a minor difference in what to configure when adding mpv-quality-menu to uosc.

But from our point of view, we now have to maintain integration with a proprietary script we chose. And if christoph goes awol, or another script pops up, what then? More branching and maintaining integrations. uosc has to be agnostic to 3rd party scripts. The only exception being thumbnailing, but that's more of a library for UIs than a user script.

FichteFoll:
> If they want buttons in control bar, they can just copy & paste the ones from its readme instructions into uosc `controls` option.

There is actually an argument for this, that being modifying controls option just to replace one button means you'd not get updates to the button bar configuration from the default options and having to check whether that has been updated in every update and then merging the changes can be rather tedious. Perhaps uosc could provice a simpler generic approach for replacing or overriding one button in particular?

tomasklaen:
I think anything we would come up with would be more complicated, from both dev time/maintenance and user configuration fatigue perspectives, than to just update your `controls` option. If there's going to be a default `controls` change, it'll be in the changelog.

Just look how it sounds: We'd have `controls` option to construct controls, and then another sets of options to control how those controls operate? That's too over-complicated. And the reason for it being so that people don't have to change their `controls` option from default? But then what's the point of that option in the first place?

---

The only alternative I'd be fine with would be to add an API that would allow 3rd party scripts to reroute any uosc binding to a different command. For example:

```lua
mp.commandv('script-message-to', 'uosc', 'reroute', 'stream-quality', 'custom mpv command')
```

That way the script provider just sets it once, uosc doesn't care about any 3rd party integrations, no new options have to be added, and user doesn't have to configure anything.

I'm gonna make a PR for it to discuss further.

FichteFoll:
Allowing third-party userscripts to reroute commands definitely sounds like a good idea that does not involve manual user configuration and should be simple to implement. 

It would not be able to set the number of streams available to be shown in the tip, but that's hardly something I needed in the first place. If it's a youtube video, I just know that there are multiple resolutions available and don't care about how many exactly.

PR #430, opened 2023-01-14
feat: added an API to reroute uosc key bindings

tomasklaen:
This allows 3rd party scripts to replace the handling of built in uosc key bindings with their own logic.

Example that replaces uosc's basic stream quality menu with christoph-heinrich/mpv-quality-menu:

```lua
mp.commandv('script-message-to', 'uosc', 'reroute-binding', 'stream-quality', 'script-binding quality_menu/video_formats_toggle')
```

Any objections or suggestions? Maybe `reroute` is not the best name to describe this?

---

ref #429

### DIFF
--- a/README.md
+++ b/README.md
@@ -547,6 +547,18 @@ External properties can also be used as control button badges:
 controls=command:icon_name:command_name#foo@script_name?My foo button
 ```
 
+### `overwrite-binding <name> <command>`
+
+Allows a overwriting handling of uosc built in bindings. Useful for 3rd party scripts that specialize in a specific domain to replace built in menus or behaviors provided by existing bindings.
+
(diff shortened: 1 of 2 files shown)


christoph-heinrich:
From a user perspective this overwrites the integrated binding, so `overwrite-binding` might be a good fit? `set-binding` might also be a good choice.

I would have expected something simpler like
``` lua
mp.register_script_message('reroute-binding', function(name, command)
    mp.add_key_binding(nil, name, function() mp.command(command) end)
end)
```
but I think your approach with the map also allows disabling the custom binding again, which wouldn't be possible with this snippet.

Overall this seems like a good idea, allowing scripts to overwrite functionality they specialize in without requiring user configuration. Although this might also lead to unwanted behavior. Installing a script and then having a button in the UI behave differently without configuring anything might be confusing. It's a double edged sword, but as long as scripts are well behaved that shouldn't be a problem.

PR #435, opened 2023-01-18
tweak: prefix below alt title with `└`

christoph-heinrich:
This is done for chapters and it looks more cohesive if it's also done for the alt title.


tomasklaen:
I didn't add it on purpose because it doesn't make sense to me semantically. `└` denotes a child item, or a subcategory, while alt title is just alternative title, not a child or subcategory.

christoph-heinrich:
True, semantically it doesn't really make sense, but it didn't look right to me without it.
Maybe it's just because I'm so used to how the chapters look.

PR #443, opened 2023-01-30
fix: buffered time indicator should respect the trimmed time width

christoph-heinrich:


tomasklaen:
Would it be too much of a premature optimization to cache the gsub per timestamp char count? 😳

This:

```lua
do
	local cache = {}
	function timestamp_width(timestamp, opts)
		local substitute = cache[#timestamp]
		if not substitute then
			substitute = timestamp:gsub('%d', '0')
			cache[#timestamp] = substitute
		end
		return text_width(substitute, opts)
	end
end
```

And use it to calculate widths of all timestamps where needed.

christoph-heinrich:
I wasn't sure if it should go in `utils.lua` or `text.lua`.

The commit from #444 got added here so that the cache can be applied there too.

PR #444, opened 2023-01-30
fix: constant width for hovered time

christoph-heinrich:
Without this the text jumps around a bit on the edges where the position gets clamped.


christoph-heinrich:
Commit got added to the other PR.

PR #454, opened 2023-02-15
Chapter range patterns

christoph-heinrich:
I've recently encountered a few videos with false positives for `endings`. Turns out it's because chapters ending with e.g. 'trending' would be detected as an ending. To fix that I've now applied the same pattern for 'ending' as was already used for 'ed'.

While doing that I've noticed that 'opening' and 'closing' were also the same way, so I've applied the same thing to them too, but since I've never actually noticed any problems with them, I've made it a separate commit. Also I don't know if having all three variants even makes sense for all of them.

This is a draft because I'm not sure about the change and it could certainly benefit from testing and discussion.

A collection of examples of what should and shouldn't be detected for each kind would take a lot of the guesswork out of such changes. I've prepared  a table in a test file (not included here) to collect examples I encounter in the wild, let's see if that will ever become good enough to be useful.

### DIFF
--- a/scripts/uosc_shared/lib/utils.lua
+++ b/scripts/uosc_shared/lib/utils.lua
@@ -430,7 +430,7 @@ function serialize_chapter_ranges(normalized_chapters)
 	local simple_ranges = {
 		{name = 'openings', patterns = {'^op ', '^op$', ' op$', 'opening$'}, requires_next_chapter = true},
 		{name = 'intros', patterns = {'^intro$'}, requires_next_chapter = true},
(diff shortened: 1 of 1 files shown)


christoph-heinrich:
I've gathered some examples from old discussions and local files and made some adjustments to make them pass.
It's debatable which of those should be hard coded or loaded from the config file, but I'll leave it hard coded for now (makes testing easier).

tomasklaen:
Makes sense. No objections merging this as is.

christoph-heinrich:
I've opened it up for merging since you're fine with it as is, but I suspect that won't be the last time those patterns get touched.

If anyone runs into false positives/negatives, please let me know so I can expand my collection and (hopefully) fix it :wink:

PR #481, opened 2023-03-27
feat: open playlist when clicking playlist indicator in the top bar

### LINKED ISSUE #480: Open the playlist menu when clicking top bar playlist indicator
As the title suggest I was wondering if it would be possible to make the Playlist Position in the Tob Bar open the playlist menu when clicked upon.

christoph-heinrich:
I'm not particularly happy with that implementation, arguably any coordinate calculation from the render function should be moved into `update_dimensions()` to avoid the duplication, but that should be a separate change.

Closes #480

### DIFF
--- a/scripts/uosc_shared/elements/TopBar.lua
+++ b/scripts/uosc_shared/elements/TopBar.lua
@@ -53,6 +53,7 @@ function TopBar:init()
 	self.top_border = options.timeline_border
 	self.show_alt_title = false
 	self.main_title, self.alt_title = nil, nil
+	self.playlist_bx = 0
 
 	local function get_maximized_command()
 		return state.border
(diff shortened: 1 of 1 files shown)


tomasklaen:
Damn, sorry you wasted your time, but this is getting a bit out of hand so I want to implement this with a new API for registering click zones during render loop, and then use it also for chapter buttons and similar stuff. I'm also playing with an idea of moving proximity calculations into each individual element, and maybe other pointer<>elements interaction changes as well. Not sure yet.

christoph-heinrich:
That sounds like a good idea (although not so sure about doing it during rendering), but unless you get to do that soon, this could still land as a temporary solution.

PR #483, opened 2023-03-27
refactor: cursor events consolidated into one interface

tomasklaen:
PR for testing and feedback if anyone is interested. I wanted a bit different and "prettier" API initially, but I've realized this is all we need.

The core of this is that `mbtn_left_down/up` and `wheel_down/up` events are now only fired on a `cursor` object, which now looks like this:

```lua
cursor = {
	x = 0,
	y = 0,
	hidden = true,
	hover_raw = false,
	-- Event handlers that are only fired on cursor, bound during render loop. Guidelines:
	-- - element activations (clicks) go to `mbtn_left_down` handler
	-- - `mbtn_button_up` is only for clearing dragging/swiping
	on_primary_down = nil,
	on_primary_up = nil,
	on_wheel_down = nil,
	on_wheel_up = nil,
	-- Called at the beginning of each render
	reset_handlers = function()
		cursor.on_primary_down, cursor.on_primary_up = nil, nil
		cursor.on_wheel_down, cursor.on_wheel_up = nil, nil
	end,
	-- Enables pointer key group captures needed by handlers (called at the end of each render)
	decide_keybinds = function()
		local mbtn_left_decision = (cursor.on_primary_down or cursor.on_primary_up) and 'enable' or 'disable'
		local wheel_decision = (cursor.on_wheel_down or cursor.on_wheel_up) and 'enable' or 'disable'
		mp[mbtn_left_decision .. '_key_bindings']('mbtn_left')
		mp[wheel_decision .. '_key_bindings']('wheel')
	end
}
```

`cursor.reset_handlers()` (clear all listeners) is called at the beginning of every frame before anything is rendered, and `cursor.decide_keybinds()` _after_ everything is rendered.

This means elements need to set their listeners on `cursor.on_*` properties in their render functions.

This allows elements to take over cursor event handling from their parents, and removes the disconnect between cursor events and rendered visibility, which in the past was causing invisible elements to still receive mbtn/wheel events, so we needed to care about element visibility inside `on_*` handlers. Now invisible elements just don't bind any listeners, so the separation of concerns is a bit cleaner imo.

It also allows us to split any element into multiple event handling zones very easily. For example, implementing #480 would be just an if statement:

```lua
if get_point_to_rectangle_proximity(cursor, playlist_position_rect) == 0 then
	cursor.on_primary_down = function() --[[ open playlist menu ]] end
end
```

Another nice use case I have for this is adding secondary buttons to menus, for example a delete button to playlist menu, or extending menu API to allow people to add any buttons they want.

ref #481

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -6,8 +6,8 @@ opt = require('mp.options')
 utils = require('mp.utils')
 msg = require('mp.msg')
 osd = mp.create_osd_overlay('ass-events')
-infinity = 1e309
-quarter_pi_sin = math.sin(math.pi / 4)
+INFINITY = 1e309
+QUARTER_PI_SIN = math.sin(math.pi / 4)
 
 -- Enables relative requires from `scripts` directory
 package.path = package.path .. ';' .. mp.find_config_file('scripts') .. '/?.lua'
@@ -452,7 +452,7 @@ function update_cursor_position(x, y)
 	-- we receive a first real mouse move event with coordinates other than 0,0.
 	if not state.first_real_mouse_move_received then
 		if x > 0 and y > 0 then state.first_real_mouse_move_received = true
-		else x, y = infinity, infinity end
+		else x, y = INFINITY, INFINITY end
 	end
 
 	-- add 0.5 to be in the middle of the pixel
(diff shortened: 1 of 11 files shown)


tomasklaen:
Damn. Forgot to push to main before making the PR and now it contains 2 old commits :/

PR #508, opened 2023-04-15
tweak: elements with min_visibility > 0.5 cause a margin

christoph-heinrich:
ref. https://github.com/tomasklaen/uosc/commit/6d58ef4c9986b6b38ad60551a891f7d690b6c1a5#commitcomment-109122443

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -472,17 +472,19 @@ end
 function update_margins()
 	if display.height == 0 then return end
 
-	local function is_persistent(element) return element and element.enabled and element:is_persistent() end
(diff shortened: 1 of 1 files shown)


tomasklaen:
The function now has a bigger scope, so renamed it to `should_margin`, `causes_margin` or something better if you can think of anything. Sorry for nitpick 😶.

christoph-heinrich:
Yeah I was somewhat bothered by that as well, but then I thought as far as the margin is concerened, those elements are still considered persistent. I'll rename it to avoid any confusion with the other `is_persistent()`.

Issue #359, opened 2022-10-31
Support for old mpv versions


po5:
You mentioned [here](https://github.com/tomasklaen/uosc/commit/d4697134a1347eb82d68def837a28c1031f58307#commitcomment-88041927) that you want to support versions before 0.33.0.

A fundamental part of the UI, mouse handling, broke support for pre-0.33.0 in 9dd78e85eb1723cf7c2f9faac49296cc59991ff2.

That's because `mouse-pos` was not a property back then, and we relied on `mp.get_mouse_pos()`.  
https://github.com/mpv-player/mpv/commit/a7686679564de584eaa720a6f682c8bd8ad062e6

`mp.get_mouse_pos()` is still usable today, and the commit message says they intend on keeping the wrapper.  
https://github.com/mpv-player/mpv/commit/0d5055fe93e35770235902e11ff39b3231e6e1b8

The obvious issues are that we have to change the way we watch for mouse movements (no property to observe), and that there is no "hover" state communicated by the old API (is it something we could compute ourselves?).

Is there a specific mpv version that you want to support (more specific than "pre-0.33.0")?  
I went with 0.29.0 for thumbfast because that's what I feel comfortable supporting, having applied the same compatibility fixes to mpv_sponsorblock years ago (it may actually support lower versions but I haven't tried it).  
But I think that going lower than 0.30.0 isn't worth the time, [shinchiro's builds](https://sourceforge.net/projects/mpv-player-windows/files/release/) only go back that far, and are the only Windows builds still linked on official pages.

tomasklaen:
The reason I wanted to support pre-0.33 was that someone in issues here said that some semi popular distro still ships with 0.32 by default. But if the numbers are not that significant, I wouldn't mind setting support to 0.33 or higher. My support target is usually 95% of whats used in the world, but I don't think there's any data source to get these numbers, so I guess we need to guesstimate it.

I'd like to support the oldest version that won't make the development and maintenance too unreasonable, and this `mouse-pos` and `enter` & `leave` events missing issue sounds like it'd do just that. It's not like anyone here is getting payed to support ancient versions.

FichteFoll:
[Debian bullseye](https://packages.debian.org/bullseye/mpv) and [Ubuntu 20.04LTS](https://packages.ubuntu.com/focal/mpv) ship 0.32. I don't know if people actually us debian for a desktop, though.

tomasklaen:
[...]

PR #518, opened 2023-04-21
feat: support internationalization

natural-harmonia-gropius:
Tested with https://github.com/natural-harmonia-gropius/mpv-config/pull/20
image

**Todo:**
- [x] Option to set language (doesn't support hot-reload)
- [x] Support custom menus
- [x] Support third-party scripts (menu title only)
- [x] Document (script-opts)


tomasklaen:
Is there a reason we can't read from `../intl/{language}.json` and need to use `expand-path`? I don't want to have absolute paths in the code, because they will change in the future.

natural-harmonia-gropius:
> Is there a reason we can't read from ../intl/{language}.json

`io.open` follow mpv.exe's cwd on my pc. 

> and need to use expand-path

If drop uosc_shared, `mp.get_script_directory() .. '/intl/' .. lang .. '.json'` will work.

https://github.com/natural-harmonia-gropius/mpv-config/pull/20/commits/7a96f68be5b860e542e8eec81c9fa993a4bdc14c
https://github.com/natural-harmonia-gropius/uosc/commit/67bbd13bf8dd70f4ea4964ad9cc8cfa8635fb0c9

tomasklaen:
Can we switch localization files to lua:

```lua
-- intl/language.lua
return {
  foo: 'bar'
}
```

and use error handled `require()` to load them?

natural-harmonia-gropius:
> Can we switch localization files to lua

https://github.com/tomasklaen/uosc/pull/518/commits/1e64420d77bde6a439836262ad13517e613d376d

hooke007:
How to prevent the clash of synonyms in this implement?

> Is there a reason we can't read from ../intl/{language}.json and need to use expand-path? I don't want to have absolute paths in the code, because they will change in the future.

I think bundling other language-files in the repo is not a good idea. Reading a conf/json files in config-dir seems more reasonable. Once other contributers help provide more language, and uosc changed or update some texts, but contributers could not make PRs in time. Whether to waiting for all languages to be updated or removed the outdated ones?

natural-harmonia-gropius:
> How to prevent the clash of synonyms in this implement?

Use placeholders like `"#title_menu_audio"` in raw strings, then add en.json.

> Reading a conf/json files

I will re-add json supports.

> uosc changed or update some texts

It would fallback to raw string for now.
I will add `options.languages=custom.json,custom2.json,zh-hans,en,slang`, the final lexicon is a composite.

> contributers could not make PRs in time

Translator doesn't require to understand the code, and I think most people use release version of script, we can wait.

tomasklaen:
So where do we want the localization files to live? I though they'd be part of uosc and just accept PRs with new/updated files. That way people don't have to hunt around to localize this.

Do you guys think they should be separate in each translator's individual repos and just link to them in readme?

natural-harmonia-gropius:
> So where do we want the localization files to live?

It is most convenient in this repo.

> they should be separate in each translator's individual repos

It would be nice add suggested translation sources will be good, and I think it's ok to merge them in every release.

hooke007:
You don't need to link each of them. Just link to a new discussion where everyone could post their trans-files.

dyphire:
> > Is there a reason we can't read from ../intl/{language}.json
> 
> `io.open` follow mpv.exe's cwd on my pc.
> 
> > and need to use expand-path
> 
> If drop uosc_shared, `mp.get_script_directory() .. '/intl/' .. lang .. '.json'` will work.
> 
> [natural-harmonia-gropius/mpv-config@7a96f68](https://github.com/natural-harmonia-gropius/mpv-config/commit/7a96f68be5b860e542e8eec81c9fa993a4bdc14c) [natural-harmonia-gropius@67bbd13](https://github.com/natural-harmonia-gropius/uosc/commit/67bbd13bf8dd70f4ea4964ad9cc8cfa8635fb0c9)

In view of this, I think it's time to revert commit https://github.com/tomasklaen/uosc/commit/4399904bba27b4082bc1be1b8ae810acfa092df4 and drop support for 0.32 and lower.
In fact, uosc have never truly supported versions before 0.33.0: https://github.com/tomasklaen/uosc/issues/359.

scripts/uosc_shared/lib/intl.lua
```diff
@@ -0,0 +1,72 @@
...
+		return {}
+	end
+
+	local json = json_file:read('*all')
+	json_file:close()
+
+	return utils.parse_json(json)
+end
+
+function get_translations()
```
tomasklaen, line 50: Rename to `make_locale()`

tomasklaen (changes requested):

tomasklaen:
> As an aside, I think it's time to revert commit https://github.com/tomasklaen/uosc/commit/4399904bba27b4082bc1be1b8ae810acfa092df4 and drop support for 0.32 and lower.

This would be a major bump, so I'm waiting for thumbfast changes (also a major bump), so that we can bundle them together.

natural-harmonia-gropius:
Done and squashed.

script-opts/uosc.conf
```diff
@@ -219,3 +219,10 @@ chapter_ranges=openings:30abf964,endings:30abf964,ads:c54e4e80
 # Add alternative lua patterns to identify beginnings of simple chapter ranges (except for `ads`)
 # Syntax: `{type}:{pattern}[,{patternN}][;{type}:{pattern}[,{patternN}]]`
 chapter_range_patterns=openings:オープニング;endings:エンディング
+
+# The uosc will display in this language.
+# slang is a keyword to get the values from --slang.
+# json file (ending in `.json`) will be merged into the translation table.
+# Example: `languages=~~/custom.json,slang,en`
+# custom.json: `{ "More": "更多" }`
```
tomasklaen, line 227: Change option description to:

```
# Localization language priority from highest to lowest.
# Built in languages can be found in `uosc_shared/intl`.
# `slang` is a keyword to inherit values from `--slang` mpv config.
# Supports paths to custom json files: `languages=~~/custom.json,slang,en`
```

tomasklaen (changes requested):

tomasklaen:
Last nitpick, I swear!

script-opts/uosc.conf
natural-harmonia-gropius: Done.

po5:
I'd be interested in a way to translate custom script menus.

One idea is that scripts e.g. [mpv-quality-menu](https://github.com/christoph-heinrich/mpv-quality-menu) could ship their own folder of translations that the user includes with `languages=~~/scripts/quality-menu/es.json`.
An obvious issue is that this doesn't allow for automatic language selection, but it also has the lowest burden on the script author and uosc.
It also merges the default translations with the script's, possibly causing conflicts. Especially bad with multiple translated scripts.

A slightly more annoying option which would be to pass all available translations upon menu creation.
This requires each script to figure out how to store their translations, it'll be very inconsistent between scripts.
The upside is that it allows auto selection, and per-script translations. Annoys both script authors and uosc.

The best of both worlds could be for scripts to pass a translation directory path upon menu creation, which uosc will check for the user's preferred languages.
The latency introduced by loading translations would be <1ms, shouldn't be an issue.
Run every item through the translation function (with a new locale table for the menu), only if the script passed a translation directory. This prevents accidental "translation" of unrelated menu items.
It puts pretty much all burden is on uosc, and encourages script authors to adopt a common internationalization workflow.

There is the remaining issue of translating parts of an item title, e.g. if there is a "Seek +n seconds" item that is dynamically set, it's impossible to translate without hardcoding every conceivable value.
I guess that's just a pattern to avoid, and it won't matter for the end user.
Reword such that dynamic info goes in the hint. "Seek" and "+5" in this example.

natural-harmonia-gropius:
> ship their own folder of translations that the user includes with `languages=~~/scripts/quality-menu/es.json`.

Ship `~~/scripts/quality-menu/intl` is enough, selection should be done in `intl.lua`. I will think about how to implement this.

> a way to translate custom script menus.

How about using `user-data` to share translations between scripts? This would be closer to modern intl libraries, but I don't know how the performance would be.

> translating parts of an item title, e.g. if there is a "Seek +n seconds" item

I think scripts should use their own `t()`.

tomasklaen:
Scripts can now send a message to add additional localization directories:

```lua
mp.commandv('script-message-to', 'uosc', 'add-intl-directory', '~~/intl-test')
```

Combined with dynamic localization strings:

```lua
t('Foo %d %s', 42, 'bar')
```

This should now cover all localization needs?

po5:
> This should now cover all localization needs?

No, since script menu items are never sent to the translation function.

`t()` can be added here, but it will likely cause issues if done unconditionally (on top of trying to double-translate items with special handling like the audio device selection menu). This is why I proposed having menu-scoped translations.
https://github.com/tomasklaen/uosc/blob/e23eea6e431553c00ce732ab0a15bee0bc8e5937/scripts/uosc_shared/elements/Menu.lua#L776
https://github.com/tomasklaen/uosc/blob/e23eea6e431553c00ce732ab0a15bee0bc8e5937/scripts/uosc_shared/elements/Menu.lua#L787

natural-harmonia-gropius:
Here's what I envision: 

```
mp.commandv('script-message-to', 'uosc', 'get-intl', '~~/other-script/intl')
```

```
return [
    ...locale,
    ...JSON.parse('~~/other-script/intl/lang.json')
]
```

Other script implement and use `t()` by it self.

---

The biggest problem with the current code is pollution of translations, and lua pattern is pertty horrible for me, I think we shouldn't use it everywhere.

tomasklaen:
I was thinking adding new options to menu data structure:

```
translate_items?: boolean;
translate_hints?: boolean;
```

Menu titles will be translated always, but items only on demand otherwise we are risking translating stuff that shouldn't be, like filenames.

We'll then translate inside `Menu:update()`, which already walks the whole menu data structure.

If people want to translate on their side, they can just copy our `t()` and use their jsons, I don't think they'd need our locale. `get-intl` as proposed would be also hitting race conditions as other scripts might extend it after current one asked for a locale, so it'd have to be a subscription, which is getting a bit convoluted imo.

PR #520, opened 2023-04-24
fix: translate audio channels and devices list

po5:
The channel count in track list couldn't be translated.
Also translate mpv's hardcoded entries in the audio devices list.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -1174,7 +1174,7 @@ bind_command('audio-device', create_self_updating_menu_opener({
 				local hint = string.match(device.name, ao .. '/(.+)')
 				if not hint then hint = device.name end
 				items[#items + 1] = {
-					title = device.description,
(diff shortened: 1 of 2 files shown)


natural-harmonia-gropius:
I will update `zh-hans` after this PR is merged.
Also, should the "No file" in title bar be translated?

tomasklaen:
I've just pushed support for dynamic localization strings. It's just a cached lua `string.format()`, so if you rebase, you can swap this to:

```lua
device.description:sub(1, 7) == 'Default' and t('Default %s', device.description:sub(8)) or t(device.description)
```

and

```lua
t('%d channels', track['audio-channels'])
```

po5:
Used `device.description:sub(9)`, 9 instead of 8, to prevent a double space, since `device.description` already has a space in it `Default (device name)`.
8 means having the translation string be `Default%s` to prevent the double space, looked uglier than `Default %s`.

tomasklaen:
Whoops, missed that you're using `%d` in there. I guess just switch `%`s that are not `%s` in your other PR.

PR #528, opened 2023-04-25
fix: take parameter values into account when caching translations

po5:
The translation format in https://github.com/tomasklaen/uosc/pull/520#issuecomment-1521431723 wouldn't work, because translations get cached independently of their parameters.
This means that opening a file with two audio tracks with differing channel counts would display both tracks as having the same channels.
Compute a unique string for caching with the given parameters. `t('Hello %s %d', 'world', 123)` uses `Hello %s %d*world*123` as a cache key.
The actual translation key remains the same, e.g. `Hello %s %d`.


po5:
An issue that is present in both this and the original code, is that uosc will crash if a script menu title contains formatting e.g. `%s`.
Lua correctly complains that string.format got the wrong number of arguments.
```
[uosc] 
[uosc] stack traceback:
[uosc]  [C]: in function 'format'
[uosc]  /home/eva/.config/mpv/scripts/uosc_shared/lib/intl.lua:81: in function 't'
[uosc]  /home/eva/.config/mpv/scripts/uosc_shared/elements/Menu.lua:802: in function 'draw_menu'
[uosc]  /home/eva/.config/mpv/scripts/uosc_shared/elements/Menu.lua:829: in function </home/eva/.config/mpv/scripts/uosc_shared/elements/Menu.lua:675>
[uosc]  (...tail calls...)
[uosc]  /home/eva/.config/mpv/scripts/uosc_shared/lib/utils.lua:575: in function 'cb'
[uosc]  mp.defaults:361: in function 'process_timers'
[uosc]  mp.defaults:530: in function 'dispatch_events'
[uosc]  mp.defaults:503: in function <mp.defaults:502>
[uosc]  [C]: in ?
[uosc]  [C]: in ?
[uosc] Lua error: /home/eva/.config/mpv/scripts/uosc_shared/lib/intl.lua:81: bad argument #2 to 'format' (no value)
```
Similar issue when adding more formatting params to a translation file than expected.
```
[uosc] 
[uosc] stack traceback:
[uosc]  [C]: in function 'format'
[uosc]  /home/eva/.config/mpv/scripts/uosc_shared/lib/intl.lua:81: in function 't'
[uosc]  /home/eva/.config/mpv/scripts/uosc.lua:1177: in function 'serializer'
[uosc]  /home/eva/.config/mpv/scripts/uosc_shared/lib/menus.lua:51: in function 'callback'
[uosc]  /home/eva/.config/mpv/scripts/uosc.lua:880: in function 'fn'
[uosc]  mp.defaults:230: in function 'fn'
[uosc]  mp.defaults:65: in function 'handler'
[uosc]  mp.defaults:380: in function 'handler'
[uosc]  mp.defaults:510: in function 'call_event_handlers'
[uosc]  mp.defaults:552: in function 'dispatch_events'
[uosc]  mp.defaults:503: in function <mp.defaults:502>
[uosc]  [C]: in ?
[uosc]  [C]: in ?
[uosc] Lua error: /home/eva/.config/mpv/scripts/uosc_shared/lib/intl.lua:81: bad argument #3 to 'format' (no value)
```
This is an issue because future changes to the uosc text can cause crashes if translation files aren't immediately updated.

po5:
Can be worked around by supplying at least as many dummy parameters as there are %'s.
```diff
diff --git a/scripts/uosc_shared/lib/intl.lua b/scripts/uosc_shared/lib/intl.lua
index e41338a..4f43b0a 100644
--- a/scripts/uosc_shared/lib/intl.lua
+++ b/scripts/uosc_shared/lib/intl.lua
@@ -79,7 +79,12 @@ function t(text, ...)
 	if not text then return '' end
 	local params = string.format('%s' .. string.rep('*%s', #table.pack(...)), text, ...)
 	if cache[params] then return cache[params] end
-	cache[params] = string.format(locale[text] or text, ...)
+	local tl = locale[text] or text
+	local fill = {}
+	for i = 1, select(2, string.gsub(tl, '%%', '')) do
+		fill[#fill+1] = 0
+	end
+	cache[params] = string.format(tl, ... or 0, unpack(fill))
 	return cache[params]
 end
 ```
Another issue that can arise is if we ever change a parameter from a string to a number value. It will crash.
```
[uosc] 
[uosc] stack traceback:
[uosc]  [C]: in function 'format'
[uosc]  /home/eva/.config/mpv/scripts/uosc_shared/lib/intl.lua:87: in function 't'
[uosc]  /home/eva/.config/mpv/scripts/uosc.lua:1177: in function 'serializer'
[uosc]  /home/eva/.config/mpv/scripts/uosc_shared/lib/menus.lua:51: in function 'callback'
[uosc]  /home/eva/.config/mpv/scripts/uosc.lua:880: in function 'fn'
[uosc]  mp.defaults:230: in function 'fn'
[uosc]  mp.defaults:65: in function 'handler'
[uosc]  mp.defaults:380: in function 'handler'
[uosc]  mp.defaults:510: in function 'call_event_handlers'
[uosc]  mp.defaults:552: in function 'dispatch_events'
[uosc]  mp.defaults:503: in function <mp.defaults:502>
[uosc]  [C]: in ?
[uosc]  [C]: in ?
[uosc] Lua error: /home/eva/.config/mpv/scripts/uosc_shared/lib/intl.lua:87: bad argument #2 to 'format' (number expected, got string)
```
For that reason I suggest we only use `%s` in uosc's built in translations, even for number values. It doesn't cause any issues when a number is fed.
This crash potential is why I made the fallback value `0` instead of an empty string in the above diff, but if we adopt string formatting for everything in uosc it wouldn't be an issue to use an empty string.
It's still possible for translations to intentionally cause problems e.g. by using `%c`, but this should cover all accidents.

tomasklaen:
Also, since I've merged #520 prematurely, rebase to get it and switch all templates to `%s` here. I think it's only that one from #520.

po5:
Added a delimiter to `key`, to avoid creating collisions. [This doesn't impact performance.](https://stackoverflow.com/a/19139394)
Made `Track N` use formatting. Updated translations I currently manage.

PR #532, opened 2023-04-25
feat: german translation

christoph-heinrich:


tomasklaen:
The `"%s channels"` string now also needs a non pluralized variant for `"%s channel"`.

christoph-heinrich:
Added translation for #533

PR #533, opened 2023-04-26
fix: translate topbar when no file

natural-harmonia-gropius:
Need to add translation, after https://github.com/tomasklaen/uosc/pull/532 and https://github.com/tomasklaen/uosc/pull/534.


tomasklaen:
Lets merge this first so that other localization PRs can rebase and add this in, and we won't need another 2 PRs afterwards.

PR #539, opened 2023-04-27
fix: translate built-in menu titles

po5:
Not-so-redundant (:


tomasklaen:
Oh, I thought these were already inside `t()`s 😳. Should've checked. Thx!

PR #540, opened 2023-04-27
fix: translate fallback chapter title

po5:
As well as handling of `(unnamed)` for mkv demuxer, empty string for disc demuxer.
https://github.com/mpv-player/mpv/blob/c5211dbf4ae38583b4a55ab63c5c07f8547f73b8/demux/demux_mkv.c#L964
https://github.com/mpv-player/mpv/blob/c5211dbf4ae38583b4a55ab63c5c07f8547f73b8/demux/demux_disc.c#L285
Other demuxers either don't supply a title, or use NULL as a fallback value.

Took care of German and Simplified Chinese too, since those have verified translations on Google.

### DIFF
--- a/scripts/uosc_shared/intl/de.json
+++ b/scripts/uosc_shared/intl/de.json
@@ -5,6 +5,7 @@
     "Audio devices": "Audiogeräte",
     "Audio tracks": "Audiospuren",
     "Autoselect device": "Automatische Geräteauswahl",
+    "Chapter %s": "Kapitel %s",
     "Chapters": "Kapitel",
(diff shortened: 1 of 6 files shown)


po5:
Should we mangle existing chapter titles if they follow the `^Chapter %d$` format?
It's common for files to have these, and it's weird to keep it untranslated.

christoph-heinrich:
Can confirm that the German translation is correct.

Chapter translations can even be taken a step further by also translating common chapter names like "Intro".
I'm not sure if that's desirable tho.

PR #546, opened 2023-04-30
fix: show message when menu items is empty

natural-harmonia-gropius:
and minor fixes for `zh-hans`


tomasklaen:
Use just "Empty" as a placeholder string.

natural-harmonia-gropius:
> Use just "Empty" as a placeholder string.

Done

po5:
Doesn't seem to work with script menus?

PR #547, opened 2023-04-30
feat: add `options.top_bar_position`

natural-harmonia-gropius:


hooke007:
Why we need a strange option?

tomasklaen:
No, sorry. This is a config bloat to display irrelevant data.

PR #548, opened 2023-05-01
feat: reintroduce lines chapter style

po5:
It had been removed in 3c85721f1ad24d030ff3f0a437aa6d5a254ff779 in favor of diamonds, for readability reasons with the introduction of uncached ranges.  
I never liked them, even if the chapter snapping feature is nice.

Overview of this new implementation of line chapters:
- Customizable chapter width
- Close-by chapters cannot produce ugly double-opacity overlaps
- Styles for full line, top, bottom, and a new option for middle (it came free with the implementation)
- Plays nice with cached ranges

Disregard the thumbnail spam, if it wasn't obvious that it's unrelated.  

I'm Embarrassed For Her _watch?v=COALkXIzN3E_000558 658

We no longer invert cached ranges.  
I've removed some of the cached ranges mangling (merging nearby ranges, skipping short ranges), it will have to be redone.  
Cached ranges no longer rely on tricky positioning based on font size. The texture is guaranteed to be continuous.

### DIFF
--- a/script-opts/uosc.conf
+++ b/script-opts/uosc.conf
@@ -22,8 +22,12 @@ timeline_opacity=0.9
 timeline_border=1
 # When scrolling above timeline, wheel will seek by this amount of seconds
 timeline_step=5
+# Timeline chapters style. available: never, diamonds, lines, lines-top, lines-bottom, lines-middle
+timeline_chapters=diamonds
 # Opacity of chapter indicators in timeline, 0 to disable
 timeline_chapters_opacity=0.8
+# Timeline chapter lines width
+timeline_chapters_width=4
 # Render cache indicators for streaming content
 timeline_cache=yes
 
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -30,7 +30,9 @@ defaults = {
 	timeline_opacity = 0.9,
 	timeline_border = 1,
 	timeline_step = 5,
+	timeline_chapters = 'diamonds',
 	timeline_chapters_opacity = 0.8,
+	timeline_chapters_width = 4,
 	timeline_cache = true,
 
 	controls = 'menu,gap,subtitles,<has_many_audio>audio,<has_many_video>video,<has_many_edition>editions,<stream>stream-quality,gap,space,speed,space,shuffle,loop-playlist,loop-file,gap,prev,items,next,gap,fullscreen',
@@ -400,7 +402,7 @@ state = {
 	has_playlist = false,
 	shuffle = options.shuffle,
 	mouse_bindings_enabled = false,
-	uncached_ranges = nil,
+	cached_ranges = nil,
 	cache = nil,
 	cache_buffering = 100,
 	cache_underrun = false,
@@ -796,7 +798,7 @@ mp.observe_property('demuxer-via-network', 'native', create_state_setter('is_str
 	Elements:trigger('dispositions')
 end))
 mp.observe_property('demuxer-cache-state', 'native', function(prop, cache_state)
-	local cached_ranges, bof, eof, uncached_ranges = nil, nil, nil, nil
+	local cached_ranges, bof, eof = nil, nil, nil
 	if cache_state then
 		cached_ranges, bof, eof = cache_state['seekable-ranges'], cache_state['bof-cached'], cache_state['eof-cached']
 		set_state('cache_underrun', cache_state['underrun'])
@@ -804,7 +806,7 @@ mp.observe_property('demuxer-cache-state', 'native', function(prop, cache_state)
 
 	if not (state.duration and (#cached_ranges > 0 or state.cache == 'yes' or
 		(state.cache == 'auto' and state.is_stream))) then
-		if state.uncached_ranges then set_state('uncached_ranges', nil) end
+		if state.cached_ranges then set_state('cached_ranges', nil) end
 		return
 	end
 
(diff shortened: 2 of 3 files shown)


po5:
Haven't shuffled code around to keep the diff readable, I know indentation is wrong inside newly added conditions.

The idea used in both of these commits is to render the entire bar in a single ASS event, and \clip out areas to keep.  
This avoids any unintentional overlaps.

christoph-heinrich:
I must be missing something here, as I've never worked with the scaled clipping before.
In the generation of `chapters_clip` you use regular coordinates, but then when inserting that text into clip a scale of 4 (factor 8) is specified. How come the coordinates don't need to be scaled 8x for that?

po5:
mpv uses scale 4 by default for drawings:
https://github.com/mpv-player/mpv/blob/94c414bd1cf8744ab979373e29d3c40aed15b86b/player/lua/assdraw.lua#L5-L7
https://github.com/mpv-player/mpv/blob/94c414bd1cf8744ab979373e29d3c40aed15b86b/player/lua/assdraw.lua#L16-L18

christoph-heinrich:
I had no idea those coordinates get scaled and made in to integers internally by the `coord()` function instead of having a decimal in the text :laughing: 
Just goes to show how little I actually know about that. Thanks for the pointer.

po5:
Doesn't play nice with `window_border_size`, wait before merging.

christoph-heinrich:
Would be great if that could also support the chapter snapping, there wouldn't be any downside to it with any of the chapter styles except for `lines`.

po5:
Can someone take a look? It now works with `window_border_size`, and I reintroduced cached ranges filtering.
I want to make sure the filtering is still correct now that it's no longer inverted, and also check that `buffered_playtime` logic is correct.
It took me a while to figure out that `ass_mt:texture` is *supposed* to overextend, lot of time wasted trying to figure out positioning when it's expected to get clipped out.

I will try to implement chapter snapping.

natural-harmonia-gropius:
Don't show indicator that time <= 1 and make lines thinner would looks better.

image
image

po5:
> Don't show indicator that time <= 1

Unrelated to this PR, diamond style chapters also show up.

> make lines thinner

Old default was 6, if anything it should be changed to match. https://github.com/tomasklaen/uosc/pull/119#issuecomment-1221527976
https://github.com/tomasklaen/uosc/blob/852b724b6acfa95795af124c87054f66f2b60cc8/scripts/uosc.lua#L160

po5:
> This is more code to maintain

Changes are small and confined enough that you shouldn't need to even think about the chapter lines code when touching other code.

> you can't click to snap to chapter anymore, and there's no way to add it back, because it'd mean removing ability to seek around the chapter

There are line modes that don't take up the entire timeline height, e.g. `lines-top`, which can act just like diamonds where a tiny portion at the top of the timeline is obscured by snapping.

Extra modes all share the same code, but if you would rather only have `lines-top` I don't care much about getting this merged, I think `lines` looks better and it's what I would use myself.

Maybe it's the diamond shape I find off-putting, so I tried removing `ass:line_to(chapter_x, chapter_y - radius)` to turn it into a triangle.
It wasn't enough so I tried setting `chapter_y` to `fay`, set `\bord0`.
The current chapter style doesn't fit the style of the rest of uosc's flat elements, it feels like it's tacked on top.
Even with these changes it's not enough, because it's still trying to tack another layer of color onto the timeline.
In my mind it should feel like a divot, not a floating icon.

Twin Peaks S01E01 Pilot 720p BluRay DD5 1 x264-DON mkv_003103 695

A-B loop indicators have the same issue, they look like they don't belong. I'm not sure how to fix it completely.
We could \clip out chapter triangles from the two sides of timeline_style=bar, which would allow us to maintain two consistent colors within the timeline.
I don't see what could be done for A-B loop.

I will submit cached ranges updates as a separate PR, feel free to close this if you're against line chapters but I'd like to hear your view on the issue of floating elements that imo don't belong on the flat timeline.

christoph-heinrich:
The notches are an interesting idea and would fit in with the volume bar, but how would that work with a minimized timeline?

dyphire:
> Maybe it's the diamond shape I find off-putting, so I tried removing `ass:line_to(chapter_x, chapter_y - radius)` to turn it into a triangle.

I prefer diamond style chapters because they are more beautiful. I have never liked chapters with line styles, as they appear chaotic and lack aesthetic appeal. But this is a personal preference difference, and I don't object to this PR.

tomasklaen:
> Changes are small and confined enough

The amount of code is not that big of an issue, it's that every time we need to touch timeline it's one more thing that needs to be tweaked (or potentially 4 in this case). Also, if we ever need to add some new indicator to timeline, it's surely going to clash with some of the chapter indicators, so I'll either have to design around it somehow (unlikely), or remove the indicator and cause a major bump.

> The current chapter style doesn't fit the style of the rest of uosc's flat elements.

We'd have to agree to disagree here, I think they look good, and I wouldn't expect uosc's design to be limited to rectangles only.

> A-B loop indicators have the same issue, they look like they don't belong.

I don't share this view either, but if you only expect rectangles than no other shape will belong I guess. Overall, having different indicators use different shapes helps people differentiate between them.

But thanks for remind me of a-b loop - here's an indicator that already clashes with lines.

> _screenshot_

I wouldn't mind this triangle alternative. It could be implemented by simply skipping drawing the top point of the diamond, and enabled with `timeline_chapters_compact=yes/no` option (or `_discrete` maybe?).

The indicator should revert to normal full diamond rendering when it's hovered.

But I do consider this an unnecessary change just to appease your particular hatred of diamonds :)

po5:
Doubt this is getting merged or implemented in any capacity, see https://github.com/tomasklaen/uosc/issues/550#issuecomment-1752004918.

Issue #538, opened 2023-04-26
Localizing menus comming from other scripts


tomasklaen:
Currently, people can use `add-intl-direcory` message to provide their own localization json files which will be merged into uosc's ones, but that doesn't on it's own translate their menus yet.

I was thinking of maybe adding new options to menu data structure:

```
translate_title?: boolean;
translate_items?: boolean;
translate_hints?: boolean;
```

since these values need to only be translated on demand otherwise we are risking translating stuff that shouldn't be, like filenames.

We'll then translate inside Menu:update(), which already walks the whole menu data structure.

But this solution doesn't support translating dynamic strings like `"%s of something"`.

Should we instead just implement `get-option <option_name>` message, which will reply with the value of requested uosc's option so that scripts can inherit `langauges`, handle translations on their end, and just send us already translated menu data?

Because I don't see any way uosc could handle translation for other scripts without incredibly messy solutions full of compromises. Nor it makes much sense to be honest since most scripts add uosc menus as an optional feature, so they can't depend on it for localization.

christoph-heinrich:
Scripts should take care of the translation themselves, it's the simplest way to get good results.

I'm not sure if providing them with the languages list all that useful. If the script does support multiple languages, it'll need it's own way of specifying the language anyway.

po5:
Scripts can already do `options.read_options(uosc_options, "uosc")` if they want to retrieve the raw uosc config.

I guess scripts that want the feature should just copy uosc's language priority and translation code, so they can use translations in their custom-drawn menus.

tomasklaen:
All right, makes sense, I'll leave it as it is.

PR #555, opened 2023-05-07
feat: `get_locale` api for scripts localization

natural-harmonia-gropius:
and deprecates `add-intl-directory`.

Since https://github.com/tomasklaen/uosc/issues/538 `add-intl-directory` no longer works as expected. Also `get_locale` avoids pollution between multiple translation sources.

Use example: https://github.com/natural-harmonia-gropius/mpv-config/pull/23/files

---

I'll update readme later if you feel ok with this.

### DIFF
--- a/scripts/uosc.lua
+++ b/scripts/uosc.lua
@@ -1257,7 +1257,11 @@ mp.register_script_message('set-min-visibility', function(visibility, elements)
 end)
 mp.register_script_message('flash-elements', function(elements) Elements:flash(split(elements, ' *, *')) end)
 mp.register_script_message('overwrite-binding', function(name, command) key_binding_overwrites[name] = command end)
-mp.register_script_message('add-intl-directory', function(path) intl.add_directory(path) end)
+mp.register_script_message('get-locale', function(script, path)
+	local locale = intl.get_locale(path)
+	local json = utils.format_json(locale)
+	mp.commandv('script-message-to', script, 'uosc-locale', json)
+end)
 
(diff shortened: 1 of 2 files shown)


tomasklaen:
I think we've decided already that scripts should handle their localization themselves. Your example is also weird. quality-menu is a standalone script with optional uosc integration. Should it only be localizable when used alongside of uosc?

It just doesn't make sense for uosc to load other script's localization files.

We should also probably remove `add-intl-directory`. People can already add custom localization files into the `languages` option, so I don't see much use for it.

natural-harmonia-gropius:
Every script uses their own language options and translation of "Disabled"/"Video" is costly.

I want all additional translate in one json. As https://github.com/cvzi/mpv-youtube-download/pull/25, it uses an option stores json, this PR is a better way to handle that in my opinion.

> People can already add custom localization files into the languages

Menu title item and hint didn't.

tomasklaen:
uosc is a UI, not a translation library. 3rd party scripts can't depend on it for localization.

> Menu title item and hint didn't.

Every script is responsible for translating their menus before they send them to us.

---

The solution to unified localization for all scripts might be extracting uosc's intl into it's own script, which will look for languages in `scripts/*/intl/*.json`, and produce a locale other scripts can ask for and use. I'd support it and integrate it into uosc if someone decided to maintain it.

But this is not and should not be uosc's job. It doesn't make sense for scripts to depend on an optional UI library for localization.

PR #568, opened 2023-05-24
doc: add windows commands

unavailable500:
Hi,
just added equivalent commands for windows to install the project.
Let me know if it's ok.
Regards

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,6 @@ Most notable features:
     wget -P /tmp/ https://github.com/tomasklaen/uosc/releases/latest/download/uosc.zip
     unzip -od ~/.config/mpv/ /tmp/uosc.zip
     rm -fv /tmp/uosc.zip
-    wget -NP ~/.config/mpv/scripts/ https://raw.githubusercontent.com/po5/thumbfast/master/thumbfast.lua
     ```
     
     On Windows these equivalent PowerShell commands can be used:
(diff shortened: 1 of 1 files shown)


tomasklaen:
I don't think thumbfast installation belongs here. It's awesome and recommended, but still a separate project with optional integration in uosc, it shouldn't be a mandatory part of uosc installation.

unavailable500:
That's the same logic that you found in the Linux installation in the line 51th. 
`wget -NP ~/.config/mpv/scripts/ https://raw.githubusercontent.com/po5/thumbfast/master/thumbfast.lua`

If you want I can remove that. Do you want that I remove that also for the Linux installation script?

tomasklaen:
I guess I merged it without noticing :)

Well, since it has been there all this time, any other inputs on this? Should we include thumbfast in these installation commands/scripts?

tomasklaen:
Here's where I'm coming from:

First, I use uosc on multiple machines, but don't need thumbfast on all of them. So I feel weird about having it be forced, because I'd have to remove it manually afterwards.

Second, I think thumbfast is going through a redesign atm that might split it into multiple files. And ther emight be further changes in the future. This means we now have to maintain an installation code for a disconnected 3rd party script, which in turn means there now are external factors that can break this, and it will remain broken until someone notices and decides to create an issue.

christoph-heinrich:
I don't think it should be part of the installation instructions.
Mentioning as an optional script that pairs well with uosc (as is already done) is enough.

unavailable500:
Done! Yes I also think it's better since thumbfast installation process could change in the future

unavailable500:
> Why is it `APPDATA` only?

What do you mean?
That's where the config file goes on windows: https://mpv.io/manual/stable/
> You can find the exact path by running echo %APPDATA%\mpv\mpv.conf in cmd.exe.

hooke007:
I cannot say all here but the win users I know most use `portable_config` only.

tomasklaen:
I use `portable_config` as well, but according to mpv docs (https://mpv.io/manual/stable/#files-on-windows) `%APPDATA%/mpv` is the primary/main intended place for mpv to read the config from, so I think this is appropriate. Everyone else can extract to their portable config folders.

PR #579, opened 2023-06-28
fix: cap char byte count of UTF-8 to string length

### LINKED ISSUE #515: `text.lua` would be broken by embedd chapters? files
sample here
[test3.zip](https://github.com/tomasklaen/uosc/files/11275575/test3.zip)

christoph-heinrich:
Capping the character byte count to the remaining string length prevents out of bounds access.

Closes #515


po5:
Can confirm this behaves the same way as vanilla osc on the full set of chapters from my original test file.  
Have you checked that wrapping works as expected?

christoph-heinrich:
Wrapping with valid utf-8 strings works as expected, I haven't tested long invalid strings. But since this only changes the length of the last char, it shouldn't make much of a difference.

PR #609, opened 2023-08-21
feat: fast seek in timeline based on drag speed

### LINKED ISSUE #607: Don't seek to keyframe when dragging timeline very slowly
Looks like it always seek to keyframe then exact.

https://github.com/tomasklaen/uosc/assets/50797982/3380abea-0521-4d61-aa55-05780283ee58

christoph-heinrich:
So far any dragging in a sufficiently long video would cause fast seeks. That was annoying when dragging slowly, because the seek would always go back to the last keyframe.
With the adaption of the drag speed measurement from the menu, we can decide if we want to do fast seeks based on cursor speed.

Closes #607

### DIFF
--- a/scripts/uosc/main.lua
+++ b/scripts/uosc/main.lua
@@ -316,10 +316,7 @@ cursor = {
 	on_wheel_down = nil,
 	on_wheel_up = nil,
 	allow_dragging = false,
-	---@type {x: number, y: number, time: number}[]
-	history = {},
-	history_size = 10,
-	velocity = {x = 0, y = 0},
+	history = CircularBuffer:new(10),
 	-- Called at the beginning of each render
 	reset_handlers = function()
 		cursor.on_primary_down, cursor.on_primary_up = nil, nil
@@ -342,6 +339,26 @@ cursor = {
 			cursor.wheel_enabled = enable_wheel
 		end
 	end,
+	find_history_sample = function()
+		local time = mp.get_time()
+		for _, e in cursor.history:iter_rev() do
+			if time - e.time > 0.1 then
+				return e
+			end
+		end
+		return cursor.history:tail()
+	end,
+	get_velocity = function()
+		local snap = cursor.find_history_sample()
+		if snap then
+			local x, y, time = cursor.x - snap.x, cursor.y - snap.y, mp.get_time()
+			local time_diff = time - snap.time
+			if time_diff > 0.001 then
+				return { x = x / time_diff, y = y / time_diff }
+			end
+		end
+		return { x = 0, y = 0 }
+	end,
 	move = function(x, y)
 		local old_x, old_y = cursor.x, cursor.y
 
@@ -360,8 +377,8 @@ cursor = {
 			Elements:update_proximities()
 
 			if cursor.x == INFINITY or cursor.y == INFINITY then
-				cursor.hidden, cursor.history = true, {}
-				cursor.velocity.x, cursor.velocity.y = 0, 0
+				cursor.hidden = true
+				cursor.history:clear()
 
 				-- Slowly fadeout elements that are currently visible
 				for _, element_name in ipairs({'timeline', 'volume', 'top_bar'}) do
@@ -375,28 +392,12 @@ cursor = {
 
 				Elements:trigger('global_mouse_leave')
 			elseif cursor.hidden then
-				cursor.hidden, cursor.history = false, {}
-				cursor.velocity.x, cursor.velocity.y = 0, 0
+				cursor.hidden = false
+				cursor.history:clear()
 				Elements:trigger('global_mouse_enter')
 			else
 				-- Update history
-				local new_index = #cursor.history + 1
-				local last = {x = x, y = y, time = mp.get_time()}
-				if #cursor.history >= cursor.history_size then
-					new_index = #cursor.history
-					for i = 1, #cursor.history, 1 do cursor.history[i] = cursor.history[i + 1] end
-				end
-				cursor.history[new_index] = last
-
-				-- Update velocity
-				if #cursor.history > 5 then
-					local first = cursor.history[1]
-					local time_delta = (last.time - first.time)
-					cursor.velocity.x = (last.x - first.x) / time_delta
-					cursor.velocity.y = (last.y - first.y) / time_delta
-				else
-					cursor.velocity.x, cursor.velocity.y = 0, 0
-				end
+				cursor.history:insert({x = cursor.x, y = cursor.y, time = mp.get_time()})
 			end
 
 			Elements:proximity_trigger('mouse_move')
@@ -424,11 +425,8 @@ cursor = {
 	-- Calculates distance in which cursor reaches rectangle if it continues moving on the same path.
 	-- Returns `nil` if cursor is not moving towards the rectangle.
 	direction_to_rectangle_distance = function(rect)
-		if cursor.hidden or #cursor.history < 10 then
-			return false
-		end
-
-		local prev = cursor.history[#cursor.history - 9]
+		local prev = cursor.find_history_sample()
+		if not prev then return false end
 		local end_x, end_y = cursor.x + (cursor.x - prev.x) * 1e10, cursor.y + (cursor.y - prev.y) * 1e10
 		return get_ray_to_rectangle_distance(cursor.x, cursor.y, end_x, end_y, rect)
 	end
@@ -455,7 +453,6 @@ state = {
 	duration = nil, -- current media duration
 	time_human = nil, -- current playback time in human format
 	destination_time_human = nil, -- depends on options.destination_time
-	framerate = 1,
 	pause = mp.get_property_native('pause'),
 	chapters = {},
 	current_chapter = nil,
@@ -829,7 +826,6 @@ end)
 mp.observe_property('display-hidpi-scale', 'native', create_state_setter('hidpi_scale', update_display_dimensions))
 mp.observe_property('cache', 'string', create_state_setter('cache'))
 mp.observe_property('cache-buffering-state', 'number', create_state_setter('cache_buffering'))
-mp.observe_property('demuxer-rawvideo-fps', 'number', create_state_setter('framerate'))
 mp.observe_property('demuxer-via-network', 'native', create_state_setter('is_stream', function()
 	Elements:trigger('dispositions')
 end))
--- a/scripts/uosc/main.lua
+++ b/scripts/uosc/main.lua
@@ -316,8 +316,10 @@ cursor = {
 	on_wheel_down = nil,
 	on_wheel_up = nil,
 	allow_dragging = false,
-	history = {}, -- {x, y}[] history
+	---@type {x: number, y: number, time: number}[]
+	history = {},
 	history_size = 10,
+	velocity = {x = 0, y = 0},
 	-- Called at the beginning of each render
 	reset_handlers = function()
 		cursor.on_primary_down, cursor.on_primary_up = nil, nil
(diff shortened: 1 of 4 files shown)


hooke007:
Anyway hr-seek is terrible for timeline drag-seeking. Especially when you load some large files or using ancient devices.
I checked all players I know, none of them give the satisfying solution.

IMO, if some one really need it, add another option for them.(like https://mpv.io/manual/master/#on-screen-controller-seekbarkeyframes)

natural-harmonia-gropius:
LGTM

tomasklaen:
We are already keeping track of cursor history on the `cursor` object in the `main.lua`, it'd be simpler and nicer to consolidate it there by extending it to also calculate velocities, and then len timeline just access it from there. I wanted to refactor it a bit anyway, I'll put something together.

christoph-heinrich:
The threshold is now based on the video duration, so that the user has to move the cursor more then 30 video seconds per second to get fast seeking. Maybe that's more to your liking.

tomasklaen:
All right, and we're back here :) So here are my thoughts on the current PR:

#### 1.
The buffer size should be 10, as 10 pointer updates are enough to judge both velocity and direction, so unless there's something obvious I'm missing, I don't think we need more. The selected reference snap should than be either the first one that is at least 100ms old, or the tail of the buffer.

---

#### 2.
Since the velocity is going to be used in a menu, where you can stop, not move for a while, and then release (which doesn't cause a move event), it should be a getter called on demand, and use current time and the last snap position to calculate the velocity. So only add to history buffer inside `cursor.move()`.

---

#### 3.
Since we have `history.size`, we don't need `history_size` anymore.

---

#### 4.
Add my fast seek condition:

```lua
if not state.is_video or math.abs(cursor.velocity.x) > 80 * math.max(1, self.pixels_per_frame) then
```

Filtering only video is obvious, and I'll copy & paste the reasoning behind `pixels_per_frame` here:

Lets imagine an extreme case where you have a very short video, and a big window, where 100px on a timeline represents a single frame. Surely you don't want a keyframe based seeking here. Multiplying threshold by `math.max(1, self.pixels_per_frame)` would increase it 100x here, ensuring you always get frame accurate seeks.

---

#### 5.
As mentioned in #611, do you think we still need the seek timer for slow seeks? We now have a UX to seek accurately, so all the timer does is make the current frame on screen jump after half a second of fast seeking, which just feels bad and glitchy. I think we should remove it.

christoph-heinrich:
Soo... I've just spent the last hour making the circular buffer grow on demand. To keep a 100ms history requires at most 15 entries for me.

> Lets imagine an extreme case where you have a very short video, and a big window, where 100px on a timeline represents a single frame. Surely you don't want a keyframe based seeking here. Multiplying threshold by math.max(1, self.pixels_per_frame) would increase it 100x here, ensuring you always get frame accurate seeks.

In such a case my current condition would also be fine.
Such a video in a 1900px wide window would be 19s long, so in order to reach the 30s threshold the user would have to move so fast that they move across the whole timeline in 0,633s. If they are really moving that fast, keyframe seeking is still fine.

christoph-heinrich:
There you go.

I've added the video check to the condition, but otherwise it's still the same because I'm convinced that works pretty well for everything. I'm on the fence about increasing the threshold or not, but the concept behind the condition is solid imo.
I dare you to find a pathological use-case :laughing: 

The history still grows automatically to be able to hold 100ms worth of history independent of update rate.
If you really want me to limit it to 10 entries, then I'll change it.

natural-harmonia-gropius:
Don't remove timer, it's necessary to avoid the gap here. Just kill and resume it in every move rather than only in fast seek.
image

natural-harmonia-gropius:
And I think we call seek too frequently, need to reduce it.

image

image

My always exact seek approach, with state based condation. (Much worse performance than keyframe with 8k sample)
I hope this could bring some inspiration to you.

```diff
diff --git a/portable_config/scripts/uosc/elements/Timeline.lua b/portable_config/scripts/uosc/elements/Timeline.lua
index 4acea26..9fcfb8e 100644
--- a/portable_config/scripts/uosc/elements/Timeline.lua
+++ b/portable_config/scripts/uosc/elements/Timeline.lua
@@ -18,7 +18,7 @@ function Timeline:init()
        self.has_thumbnail = false
 
        -- Delayed seeking timer
-       self.seek_timer = mp.add_timeout(0.05, function() self:set_from_cursor() end)
+       self.seek_timer = mp.add_timeout(0.05, function() self:set_from_cursor(true) end)
        self.seek_timer:kill()
 
        -- Release any dragging when file gets unloaded
@@ -89,10 +89,19 @@ function Timeline:get_time_at_x(x)
        return state.duration * progress
 end

+local requested = -1
+
 ---@param fast? boolean
-function Timeline:set_from_cursor(fast)
+function Timeline:set_from_cursor(force)
        if state.time and state.duration then
-               mp.commandv('seek', self:get_time_at_x(cursor.x), fast and 'absolute+keyframes' or 'absolute+exact')
+               local expect = self:get_time_at_x(cursor.x)
+               local actual = state.time
+               local ready = math.abs(requested - actual) < 0.125 or requested == -1
+               if ready or force then
+                       requested = expect
+                       mp.commandv('seek', requested, 'absolute+exact')
+                       print('seek to: ' .. requested)
+               end
        end
 end

@@ -106,6 +115,7 @@ function Timeline:handle_cursor_down()
        mp.set_property_native('pause', true)
        self:set_from_cursor()
        cursor.on_primary_up = function() self:handle_cursor_up() end
+       requested = -1
 end
 function Timeline:on_prop_duration() self:decide_enabled() end
 function Timeline:on_prop_time() self:decide_enabled() end
@@ -127,11 +137,9 @@ function Timeline:on_global_mouse_move()
        if self.pressed then
                self.pressed.distance = self.pressed.distance + get_point_to_point_proximity(self.pressed.last, cursor)
                self.pressed.last.x, self.pressed.last.y = cursor.x, cursor.y
-               if self.width / state.duration < 10 then
-                       self:set_from_cursor(true)
-                       self.seek_timer:kill()
-                       self.seek_timer:resume()
-               else self:set_from_cursor() end
+               self:set_from_cursor()
+               self.seek_timer:kill()
+               self.seek_timer:resume()
        end
 end
 function Timeline:handle_wheel_up() mp.commandv('seek', options.timeline_step) end
```

christoph-heinrich:
@natural-harmonia-gropius I don't think that throttling seek commands is helpful.
Here is a little patch that waits after each seek command until the time changes and only then sends the next command.
```diff
diff --git a/scripts/uosc/elements/Timeline.lua b/scripts/uosc/elements/Timeline.lua
index 615642b..bda4d54 100644
--- a/scripts/uosc/elements/Timeline.lua
+++ b/scripts/uosc/elements/Timeline.lua
@@ -16,6 +16,9 @@ function Timeline:init()
 	self.top_border = options.timeline_border
 	self.is_hovered = false
 	self.has_thumbnail = false
+	---@type nil|boolean
+	self.seek_queue = nil
+	self.seek_waiting = false
 
 	-- Release any dragging when file gets unloaded
 	mp.register_event('end-file', function() self.pressed = false end)
@@ -88,7 +91,14 @@ end
 ---@param fast? boolean
 function Timeline:set_from_cursor(fast)
 	if state.time and state.duration then
-		mp.commandv('seek', self:get_time_at_x(cursor.x), fast and 'absolute+keyframes' or 'absolute+exact')
+		local time = self:get_time_at_x(cursor.x)
+		if self.seek_waiting then
+			self.seek_queue = fast == true
+		else
+			mp.commandv('seek', time, fast and 'absolute+keyframes' or 'absolute+exact')
+			self.seek_waiting = true
+			self.seek_queue = nil
+		end
 	end
 end
 
diff --git a/scripts/uosc/main.lua b/scripts/uosc/main.lua
index 482406b..47e0d3c 100644
--- a/scripts/uosc/main.lua
+++ b/scripts/uosc/main.lua
@@ -763,7 +763,11 @@ mp.observe_property('playback-time', 'number', create_state_setter('time', funct
 			else handle_file_end() end
 		end
 	end
-
+	local tl = Elements.timeline
+	tl.seek_waiting = false
+	if tl.seek_queue ~= nil then
+		tl:set_from_cursor(tl.seek_queue)
+	end
 	update_human_times()
 	select_current_chapter()
 end))
```
Seeking doesn't seem any faster to me then it does without that.

tomasklaen:
> Soo... I've just spent the last hour making the circular buffer grow on demand.

Noo, that was so unnecessary :( Simplify it back and set the buffer size to 10.

As I've mentioned, 10 samples is enough to get an accurate metric of cursor's velocity or direction, but the main reason I want it to be 10 is that it'll provide a more responsive measurements for high Hz devices. On these devices, we will almost always be using the tail of the buffer, which will be significantly younger than 100ms, therefore there will be less lag behind real and measured velocity/direction. With growing buffer, you indirectly eliminate the advantage of high Hz devices, making them just as inaccurate and laggy as all the other ones.

image

christoph-heinrich:
> Noo, that was so unnecessary :( Simplify it back and set the buffer size to 10.

I implemented it before I saw your comment, but that's fine as I've been wondering how a growing circular buffer could even be implemented and now I know :smiley:  All the circular buffers I've made up to that point were in C and fixed size.

> With growing buffer, you indirectly eliminate the advantage of high Hz devices, making them just as inaccurate and laggy as all the other ones.

I actually doubt that one would notice any additional delay in velocity measurement, after all 100ms is still pretty quick, but it's fixed 10 entries now and the buffer can't grow anymore.

scripts/uosc/main.lua
```diff
@@ -340,9 +339,77 @@ cursor = {
...
+	get_velocity = function()
+		local snap = cursor.find_history_sample()
+		if snap then
+			local x, y, time = cursor.x - snap.x, cursor.y - snap.y, mp.get_time()
+			local time_diff = time - snap.time
+			if time_diff > 0.001 then
+				return { x = x / time_diff, y = y / time_diff}
+			end
+		end
+		return { 0, 0 }
```
tomasklaen, line 360: Fallback should be a map instead of tuple.

christoph-heinrich:
I've made a new branch with cleaned up commits (including the fix for the fallback map) https://github.com/christoph-heinrich/uosc/tree/timeline_seek_metric2.
Would those be ok like that in terms of commit structure? Then I'll change this branch to that.

While dealing with that I noticed that your suggested condition has `not state.is_video or`, while I read it as `not state.is_video and` and thought the `not` was a mistake. What's the reasoning behind always using fast seeks for audio only? Does it even make a difference?

And regarding @natural-harmonia-gropius complaint about removing the timer. Should we add it back in? I thought it would be fine without because you can move the cursor slowly if you want to be more accurate, but without the timer it can also lead to unexpected situations for users like the one they pointed out.

tomasklaen:
> Would those be ok like that in terms of commit structure? Then I'll change this branch to that.

Sure, it's fine.

> What's the reasoning behind always using fast seeks for audio only? 

I assume audio is fast so slow seeking would just be a nuisance with no benefit. And if it works similar to fast video, it'd produce an audio that sounds like its stuttering the same couple milliseconds.

> the timer. Should we add it back in? 

No. The negatives of timer way outweigh the one positive, which is no longer necessary as you can control accurate seeks now.

hooke007:
There is no 'keyframes' in audio-only file. player always does hr-seek for audio.

christoph-heinrich:
I've tested fast/exact seeks with multiple audio codecs and it doesn't make a difference with any of them.
Doing the threshold check only makes sense for things where fast seeking actually makes a difference, so I'll leave it restricted to `state.is_video` for now.

PR #611, opened 2023-08-23
feat: fast seek in timeline based on cursor velocity

### LINKED ISSUE #607: Don't seek to keyframe when dragging timeline very slowly

tomasklaen:
Does this work fine for you guys? @christoph-heinrich @natural-harmonia-gropius 

closes #607, closes #609

### DIFF
--- a/scripts/uosc/elements/Timeline.lua
+++ b/scripts/uosc/elements/Timeline.lua
@@ -16,6 +16,7 @@ function Timeline:init()
 	self.top_border = options.timeline_border
 	self.is_hovered = false
 	self.has_thumbnail = false
+	self.pixels_per_frame = 1
 
 	-- Delayed seeking timer
 	self.seek_timer = mp.add_timeout(0.05, function() self:set_from_cursor() end)
@@ -52,6 +53,13 @@ end
 
 function Timeline:get_is_hovered() return self.enabled and self.is_hovered end
 
+function Timeline:update_pixels_per_frame()
+	if state.is_video and state.framerate and state.duration then
+		self.pixels_per_frame = self.width / (state.duration * state.framerate)
+	else
+		self.pixels_per_frame = 1
+	end
+end
 function Timeline:update_dimensions()
 	if state.fullormaxed then
 		self.size_min = options.timeline_size_min_fullscreen
@@ -74,6 +82,7 @@ function Timeline:update_dimensions()
 	if Elements.top_bar.enabled then available_space = available_space - Elements.top_bar.size end
 	self.obstructed = available_space < self.size_max + 10
 	self:decide_enabled()
+	self:update_pixels_per_frame()
 end
 
 function Timeline:get_time_at_x(x)
@@ -107,7 +116,8 @@ function Timeline:handle_cursor_down()
 	self:set_from_cursor()
 	cursor.on_primary_up = function() self:handle_cursor_up() end
 end
-function Timeline:on_prop_duration() self:decide_enabled() end
+function Timeline:on_prop_duration() self:decide_enabled() self:update_pixels_per_frame() end
+function Timeline:on_prop_framerate() self:update_pixels_per_frame() end
 function Timeline:on_prop_time() self:decide_enabled() end
 function Timeline:on_prop_border() self:update_dimensions() end
 function Timeline:on_prop_fullormaxed() self:update_dimensions() end
@@ -127,7 +137,7 @@ function Timeline:on_global_mouse_move()
 	if self.pressed then
 		self.pressed.distance = self.pressed.distance + get_point_to_point_proximity(self.pressed.last, cursor)
 		self.pressed.last.x, self.pressed.last.y = cursor.x, cursor.y
-		if math.abs(cursor.velocity.x) > 80 then
+		if not state.is_video or math.abs(cursor.velocity.x) > 80 * math.max(1, self.pixels_per_frame) then
 			self:set_from_cursor(true)
 			self.seek_timer:kill()
 			self.seek_timer:resume()
--- a/scripts/uosc/elements/Timeline.lua
+++ b/scripts/uosc/elements/Timeline.lua
@@ -127,7 +127,7 @@ function Timeline:on_global_mouse_move()
 	if self.pressed then
 		self.pressed.distance = self.pressed.distance + get_point_to_point_proximity(self.pressed.last, cursor)
 		self.pressed.last.x, self.pressed.last.y = cursor.x, cursor.y
-		if self.width / state.duration < 10 then
+		if math.abs(cursor.velocity.x) > 80 then
 			self:set_from_cursor(true)
 			self.seek_timer:kill()
 			self.seek_timer:resume()
--- a/scripts/uosc/main.lua
+++ b/scripts/uosc/main.lua
@@ -316,8 +316,10 @@ cursor = {
 	on_wheel_down = nil,
 	on_wheel_up = nil,
 	allow_dragging = false,
-	history = {}, -- {x, y}[] history
+	---@type {x: number, y: number, time: number}[]
+	history = {},
 	history_size = 10,
+	velocity = {x = 0, y = 0},
 	-- Called at the beginning of each render
 	reset_handlers = function()
 		cursor.on_primary_down, cursor.on_primary_up = nil, nil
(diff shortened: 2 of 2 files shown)


natural-harmonia-gropius:
I feel good.

christoph-heinrich:
image

```diff
diff --git a/scripts/uosc/main.lua b/scripts/uosc/main.lua
index 8b1e51b..2153437 100644
--- a/scripts/uosc/main.lua
+++ b/scripts/uosc/main.lua
@@ -428,7 +428,7 @@ cursor = {
 			return false
 		end
 
-		local prev = cursor.history[#cursor.history - 10]
+		local prev = cursor.history[#cursor.history - 9]
 		local end_x, end_y = cursor.x + (cursor.x - prev.x) * 1e10, cursor.y + (cursor.y - prev.y) * 1e10
 		return get_ray_to_rectangle_distance(cursor.x, cursor.y, end_x, end_y, rect)
 	end
```

I prefer the filter window to be time based instead of sample based, because then it's able to quickly react to large changes. E.g. fast cursor movement, then stop for a bit and then continue slowly. The sample based approach will ramp down the speed while moving slowly, while the time based approach will immediately have low speeds.

Also it's better to adapt the threshold of cursor speed to window width. Imagine seeking in a narrow window, one will inevitably move the cursor slower then when the timeline takes up the whole screen width for the same seeking intent.

hooke007:
Should we consider about duration here?
A video with 5s and another one with 5 hours (Obvious it has enough keyframes for slow dragging) could share the same length of timeline, but still use the same logic, it looks not reasonable.

natural-harmonia-gropius:
> Should we consider about duration here?

This should work

```diff
diff --git a/portable_config/scripts/uosc/elements/Timeline.lua b/portable_config/scripts/uosc/elements/Timeline.lua
index c971626..d56879b 100644
--- a/portable_config/scripts/uosc/elements/Timeline.lua
+++ b/portable_config/scripts/uosc/elements/Timeline.lua
@@ -127,7 +127,7 @@ function Timeline:on_global_mouse_move()
        if self.pressed then
                self.pressed.distance = self.pressed.distance + get_point_to_point_proximity(self.pressed.last, cursor)
                self.pressed.last.x, self.pressed.last.y = cursor.x, cursor.y
-               if math.abs(cursor.velocity.x) > 80 then
+               if self:get_time_at_x(math.abs(cursor.velocity.x)) > 5 then
                        self:set_from_cursor(true)
                        self.seek_timer:kill()
```

tomasklaen:
> I prefer the filter window to be time based

I had it time based originally, but I swapped it for this because:
- It needed big/unknown history size. I'd often exceed hundreds of snaps in 100ms when moving cursor as I normally do.
- Implementation is a lot simpler.
- This actually optimizes for more accurate detection of the exact scenario you describe. For example, assuming window is 100ms, lets say we have infinite snaps, and you move mouse really fast, then stop for 99ms, and move it by 1px. If you have time based window, it'll select snap that is really far away, and resulting velocity for that last pixel will be way higher than if you selected the 10th last one. The slow to fast speed change detection is a bit less accurate, but this is virtually imperceptible since it's really fast to populate 10 snaps.
- I did test it, and I haven't perceived any difference. I think you have to be superhuman to do.

Also, I don't think the position updating frequency is chained to monitor refresh rate. It's way higher, at least for me. I visualized it with:

```lua
for _, snap in ipairs(cursor.history) do
	ass:new_event()
	ass:rect(snap.x-1, snap.y-1, snap.x+1, snap.y+1)
end
```

And even on 144hz monitor, when I move cursor really fast, I see around 100px between cursor "ghosts", but there's a lot of snapshot dots between each of them, almost one per each pixel moved. At normal cursor movement speed, there's a snapshot for every moved pixel, so the accuracy is really high. The frequency goes down only when I try visualizing a big history (500), and then the ass rendered probably overwhelms available resources, and the snaps become really far apart (100-200px).

So weighing these two solutions, I went with the one that had less cons and more pros.

> Why does the menu not also use that cursor velocity?

Cause it came before, already works, doesn't need touching, which means it's an effort and time to refactor something for no benefit atm :) But yes, I'd be nice if it used it too.

> Also it's better to adapt the threshold of cursor speed to window width. Imagine seeking in a narrow window, one will inevitably move the cursor slower then when the timeline takes up the whole screen width for the same seeking intent.

I thought about it, but the threshold is already pretty sensitive and easy to flip into fast seeking, so I don't see much value here. Lowering it further for small windows would make it almost impossible to do accurate seeks. And when I'm dragging to seek frame by frame, I do it at similar speed, regardless of the window size.

tomasklaen:
> Should we consider about duration here?

Yeah, I think we should multiply the threshold based on pixels per frame ratio, which should be maxed to 1. I'll look into it.

natural-harmonia-gropius:
And we should also kill the timer in exact seeking, to fix glitch when velocity near 80.

```diff
diff --git a/portable_config/scripts/uosc/elements/Timeline.lua b/portable_config/scripts/uosc/elements/Timeline.lua
index d9333d7..19a1abb 100644
--- a/portable_config/scripts/uosc/elements/Timeline.lua
+++ b/portable_config/scripts/uosc/elements/Timeline.lua
@@ -134,7 +134,10 @@ function Timeline:on_global_mouse_move()
                        self:set_from_cursor(true)
                        self.seek_timer:kill()
                        self.seek_timer:resume()
-               else self:set_from_cursor() end
+               else
+                       self:set_from_cursor()
+                       self.seek_timer:kill()
```

tomasklaen:
The threshold now requires way faster speed if there's more than 1 pixel per frame on the timeline.

> And we should also kill the timer in exact seeking, to fix glitch when velocity near 80.

What glitch when velocity near 80?

> Going further, I hope we deprecate timouts. Currently we call seek every refresh, what about merge x in script and calculate merged x to timestamp in fixed every 1/10 second then always call seek exact?

I don't know what you mean by this either.

hooke007:
Improved a little but performance regression is still noticeable for me. Tested by any modern movie.

natural-harmonia-gropius:
> What glitch when velocity near 80?

When you get velocities [79, 81, 79] They come commands [exact, keyframe, keyframe] by timer, the first keyframe make timestamp go back to some where before exact.

> I don't know what you mean by this either.

`mp.commandv('seek'` be called every mouse move now.

image

image

I want it from:

```
1
seek 1
1
seek 2
1
seek 3
1
seek 4
100
debounce(seek 104, 0.05)
1
seek 105
```

to:

```
interval(seek, 0.1)
1
time_diff or x_diff+=1
1
diff+=1
1
diff+=1
1
diff+=1
seek(diff)
```

christoph-heinrich:
> * This actually optimizes for more accurate detection of the exact scenario you describe. For example, assuming window is 100ms, lets say we have infinite snaps, and you move mouse really fast, then stop for 100ms, and move it by 1px. If you have time based window, it'll select snap that is really far away, and resulting velocity for that last pixel will be way higher than if you selected the 10th last one. The slow to fast speed change detection is a bit less accurate, but this is virtually imperceptible since it's really fast to populate 10 snaps.

When you stop for 100ms with a 100ms window and then move by 1px it'll use the latest and the second to latest position.
That would mean 1px/0.1s=10px/s.
If you took the 10th latest one it would be much more then 1px away and only a little over 100ms (because it used to move fast).
That isn't just me theorizing, I've put print statements in to confirm this.
Would one notice the difference? On the timeline probably not, but when reusing this for the Menu it would lead to situations where the user drags fast, then pauses, then lets go. The calculated speed would be too big and result in fling.

> And even on 144hz monitor, when I move cursor really fast, I see around 100px between cursor "ghosts", but there's a lot of snapshot dots between each of them, almost one per each pixel moved. At normal cursor movement speed, there's a snapshot for every moved pixel, so the accuracy is really high. The frequency goes down only when I try visualizing a big history (500), and then the ass rendered probably overwhelms available resources, and the snaps become really far apart (100-200px).

I get an update at most every ~8ms with a 60hz monitory, so about twice per frame, and the distance between updates is >30px when moving fast. Maybe it depends on the mouse. I currently use a cheap old office dell mouse because my other one died :/

> Cause it came before, already works, doesn't need touching, which means it's an effort and time to refactor something for no benefit atm :) But yes, I'd be nice if it used it too.

Already implemented it on my branch before reading this lol.
I also added a circular buffer with 100 entries, but if you get updates that often it might not be sufficient...

> I thought about it, but the threshold is already pretty sensitive and easy to flip into fast seeking, so I don't see much value here. Lowering it further for small windows would make it almost impossible to do accurate seeks. And when I'm dragging to seek frame by frame, I do it at similar speed, regardless of the window size. For it to behave differently based on window size feels like it'd drive users crazy (at least me).

When I initially tested it I noticed that I changed the speed based on window width, which is why I made it like that.
I guess that depends on the person...

> Yeah, I think we should multiply the threshold based on pixels per frame ratio

I don't see how adjusting the threshold based on the video frame rate makes any sense.

PR #621, opened 2023-09-10
feat: add a background to timeline tooltips

christoph-heinrich:
Depending on the current frame of the video the hovered time and chapter title might be hard to read.

Do we even want something like this?

And if so, how big should the background margins be? Should they be adjusted to always line up with the background of the other ones, like is the case in the screenshot?
Currently it's based on font size, but of course that doesn't necessarily line up neatly with the other backgrounds, and it uses the timeline opacity (0.7 in the screenshot).

I didn't have a good sample on hand, but here is what it currently looks like.
image

P.S. I had to change the line splitting in the wrap_text() because otherwise it always added an additional empty line at the end, which led to the wrong line count.

### DIFF
--- a/scripts/uosc/elements/Timeline.lua
+++ b/scripts/uosc/elements/Timeline.lua
@@ -161,6 +161,10 @@ function Timeline:render()
 	local hide_text_ramp = hide_text_below / 2
 	local text_opacity = clamp(0, size - hide_text_below, hide_text_ramp) / hide_text_ramp
 
+	local tooltip_gap = 2
+	local tooltip_margin = 10
+	local timestamp_gap = tooltip_gap
+
 	local spacing = math.max(math.floor((self.size_max - self.font_size) / 2.5), 4)
 	local progress = state.time / state.duration
 	local is_line = options.timeline_style == 'line'
@@ -290,7 +294,12 @@ function Timeline:render()
 				end
 
 				-- Render hovered chapter above others
-				if hovered_chapter then draw_chapter(hovered_chapter.time, diamond_radius_hovered) end
+				if hovered_chapter then
+					draw_chapter(hovered_chapter.time, diamond_radius_hovered)
+					timestamp_gap = tooltip_gap + round(diamond_radius_hovered)
+				else
+					timestamp_gap = tooltip_gap + round(diamond_radius)
+				end
 			end
 
 			-- A-B loop indicators
@@ -372,12 +381,10 @@ function Timeline:render()
 		local tooltip_anchor = {ax = ax, ay = ay, bx = bx, by = by}
 
 		-- Timestamp
-		local offset = #state.chapters > 0 and 10 or 4
-		local opts = {size = self.font_size, offset = offset}
+		local opts = {size = self.font_size, offset = timestamp_gap, margin = tooltip_margin}
 		local hovered_time_human = format_time(hovered_seconds, state.duration)
 		opts.width_overwrite = timestamp_width(hovered_time_human, opts)
-		ass:tooltip(tooltip_anchor, hovered_time_human, opts)
-		tooltip_anchor.ay = tooltip_anchor.ay - self.font_size - offset
+		tooltip_anchor = ass:tooltip(tooltip_anchor, hovered_time_human, opts)
 
 		-- Thumbnail
 		if not thumbnail.disabled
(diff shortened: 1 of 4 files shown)


tomasklaen:
Yes, this is nice.

> how big should the background margins be?

What you have on a screenshot looks fine to me.

> Should they be adjusted to always line up with the background of the other ones, like is the case in the screenshot?

I don't know what you mean by this.

---

My only nitpicks are mostly a couple misnamed variables that made reading this code a bit confusing:
- `margin` should be the distance between tooltip and the element (so the static `10` in this case). What I called `margin` (dunno why) are actually tooltip position limits. `margin` should also be configurable I think.
- Rename `bg_margin` to `padding`.
- Rename `w2` to `text_width_half`.
- Rename `bg_opacity` option to just `opacity`. Same as `timeline_opacity` also controls only the background.
- I think `bg_opacity = options.timeline_opacity` should be the default, so that we have readable tooltips everywhere by default.

Otherwise it looks fine to me.

christoph-heinrich:
> I don't know what you mean by this.

I should have explained this better. Currently the padding size depends on the font size, which just happens to work out to line up well in windowed mode (like in the screenshot), but when fullscreened the text size increases, leading to larger padding but still same distance between the text -> tooltip backgrounds overlap.

I could make the padding constant with the right size to have the backgrounds line up, but then the padding would look too small for large font sizes. Alternatively the distance between time and chapter title could also change with font size.

Another question that came up while thinking about this:
With the background being part of the tooltip, maybe the background should start at `element.ay - offset` instead of the text? That should make it easier to avoid having the background overlap with the element when using dynamic padding size.

> Rename `bg_margin` to `padding`

I took that name from the topbar :)

> Rename `bg_opacity` option to just `opacity`. Same as `timeline_opacity` also controls only the background.

That's what I had at first, but then I renamed it in case someone wants transparent text. There is probably not actually a valid use case for this though.

> I think `bg_opacity = options.timeline_opacity` should be the default

That makes sense as long as the timeline and controls are the only users of tooltip.
I'll do it for now as it can always be adjusted in the future if the need arises.

tomasklaen:
> Currently the padding size depends on the font size, which just happens to work out to line up well in windowed mode (like in the screenshot), but when fullscreened the text size increases, leading to larger padding but still same distance between the text -> tooltip backgrounds overlap.

Yeah I've noticed that and I'm fine with it being a fraction of font size. Didn't think about the overlap in fullscreen though. But here's a nice solution:

If `:tooltip()` returns the rect coordinates of the rendered tooltip, we can just pass it to the next `:tooltip()` with an appropriate `offset`.

> With the background being part of the tooltip, maybe the background should start at element.ay - offset instead of the text?

Yes, definitely.

christoph-heinrich:
With chapters and thumbnails:
image

With chapters and without thumbnails:
image

Without chapters or thumbnails:
image

Hovering over chapters ain't great though:
image

christoph-heinrich:
Having no gap between time and chapter title looks kind of weird for short chapter titles.

Here is what it looks like with 2px gap
image

and for comparison also a longer title with and without 2px gap
image
image

tomasklaen:
Definitely with the gap. And I've meant for `:tooltip()` to return the same type of table it accepts, so we can do:

```lua
-- Somewhere at the top, we define gaps:
local tooltips_gap = 2
local timestamp_gap = tooltips_gap

-- Inside chapters (if they are rendered), we update `timestamp_gap` so that it starts 2px above diamonds, if a diamond is hovered, 2px above that.

-- We use cursor line as initial tooltip anchor, just as it is now.
local tooltip_anchor = {...}

-- When rendering tooltip, we update the anchor:
tooltip_anchor = ass:tooltip(tooltip_anchor, hovered_time_human, {offset = tooltip_gap, ...})

-- If any, render thumbnail `tooltips_gap` above current `tooltip_anchor`, and update it with thumbnail's coordinates.

-- Then use that to render chapter title
ass:tooltip(tooltip_anchor, chapter_title, {offset = tooltip_gap, ...})
```

christoph-heinrich:
Here, but we don't want to overwrite the x coordinates, otherwise subsequent tooltips are limited in how close they can get to the edge by the clipping of the tooltips before them.

christoph-heinrich:
> Here, but we don't want to change the x coordinates, otherwise subsequent tooltips are limited in how close they can get to the edge by the clamping of the tooltips before them.

christoph-heinrich:
While using it the padding with `math.floor(font_size/4)` seemed a bit too big.
I've changed it to `round(font_size/x)` and for comparison here are the biggest (fullscreen chapter) and smallest (controls) tooltips with x being 4, 6 or 8:

image
image
image

image
image
image

Also as a reference the chapters in the top bar use a padding of `font_size/6.66666`. It's not rounded apparently and I never noticed any blurriness because it happens to get pretty close to whole numbers, but of course that's not necessarily the case.
I'll go with 6 for now for the consistency with chapters in the top bar, but 8 doesn't seem too bad to me either.

christoph-heinrich:
Done

PR #625, opened 2023-09-16
feat: make the menu searchable

christoph-heinrich:
A search can be started with `ctrl+f` or `/` and stopped with `esc`.

It currently works for anything using `create_self_updating_menu_opener()`, `open_file_navigation_menu()` and menus opened by scripts.
Internal menus do a case insensitive search of the title and hint, while scripts have to do their own filtering.
The reason why scripts have to do their own filtering is because that allows scripts like [memo](https://github.com/po5/memo) to do proper filtering despite splitting up the results in pages.

I've use a little bit of code from console.lua ([1](https://github.com/mpv-player/mpv/blob/3972fd1be4ef0f8fe083b43e41c3e06104003a3c/player/lua/console.lua#L258-L260) [2](https://github.com/mpv-player/mpv/blob/3972fd1be4ef0f8fe083b43e41c3e06104003a3c/player/lua/console.lua#L769-L770)), but I got permission from the original author.

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -261,7 +261,7 @@ function Menu:update_dimensions()
 	for _, menu in ipairs(self.all) do
 		local width = math.max(menu.search and menu.search.width or 0, menu.max_width)
 		menu.width = round(clamp(min_width, width, display.width * 0.9))
-		local title_height = (menu.is_root and menu.title) and self.scroll_step or 0
+		local title_height = (menu.is_root and menu.title or menu.search) and self.scroll_step or 0
 		local max_height = round(height_available * 0.9 - title_height)
 		local content_height = self.scroll_step * #menu.items
 		menu.height = math.min(content_height - self.item_spacing, max_height)
@@ -886,7 +886,7 @@ function Menu:render()
 		local is_current, is_parent, is_submenu = pos == 0, pos < 0, pos > 0
 		local menu_opacity = pos == 0 and opacity or opacity * (options.menu_parent_opacity ^ math.abs(pos))
 		local ax, ay, bx, by = x, menu.top, x + menu.width, menu.top + menu.height
-		local draw_title = menu.is_root and menu.title
+		local draw_title = menu.is_root and menu.title or menu.search
 		local scroll_clip = '\\clip(0,' .. ay .. ',' .. display.width .. ',' .. by .. ')'
 		local start_index = math.floor(menu.scroll_y / self.scroll_step) + 1
 		local end_index = math.ceil((menu.scroll_y + menu.height) / self.scroll_step)
@@ -1027,7 +1027,6 @@ function Menu:render()
 		if draw_title then
 			local title_ay = ay - self.item_height
 			local title_height = self.item_height - 3
-			menu.ass_safe_title = menu.ass_safe_title or ass_escape(menu.title)
 
 			-- Background
 			ass:rect(ax + 2, title_ay, bx - 2, title_ay + title_height, {
@@ -1044,6 +1043,7 @@ function Menu:render()
 					clip = '\\clip(' .. ax + 2 .. ',' .. title_ay .. ',' .. bx - 2 .. ',' .. ay .. ')',
 				})
 			else
+				menu.ass_safe_title = menu.ass_safe_title or ass_escape(menu.title)
 				ass:txt(ax + menu.width / 2, title_ay + (title_height / 2), 5, menu.ass_safe_title, {
 					size = self.font_size, bold = true, color = bg, wrap = 2, opacity = menu_opacity,
 					clip = '\\clip(' .. ax + 2 .. ',' .. title_ay .. ',' .. bx - 2 .. ',' .. ay .. ')',
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -1065,19 +1065,36 @@ function Menu:render()
 			local title_height = self.item_height - 3
 
 			-- Background
-			ass:rect(ax + 2, title_ay, bx - 2, title_ay + title_height, {
-				color = fg, opacity = menu_opacity * 0.8, radius = 2,
-			})
-			ass:texture(ax + 2, title_ay, bx - 2, title_ay + title_height, 'n', {
-				size = 80, color = bg, opacity = menu_opacity * 0.1,
-			})
+			if menu.search then
+				ass:rect(ax + 3, title_ay + 1, bx - 3, title_ay + title_height - 1, {
+					color = fg .. '\\1a&HFF', opacity = menu_opacity * 0.1, radius = 2,
+					border = 1, border_color = fg, border_opacity = menu_opacity * 0.8
+				})
+				ass:texture(ax + 3, title_ay + 1, bx - 3, title_ay + title_height - 1, 'n', {
+					size = 80, color = bg, opacity = menu_opacity * 0.1, anchor_x = ax + 2, anchor_y = title_ay,
+				})
+			else
+				ass:rect(ax + 2, title_ay, bx - 2, title_ay + title_height, {
+					color = fg, opacity = menu_opacity * 0.8, radius = 2,
+				})
+				ass:texture(ax + 2, title_ay, bx - 2, title_ay + title_height, 'n', {
+					size = 80, color = bg, opacity = menu_opacity * 0.1,
+				})
+			end
 
 			-- Title
 			if menu.search then
-				ass:txt(bx - 4, title_ay + (title_height / 2), 6, ass_escape(menu.search.query), {
-					size = self.font_size, color = bg, wrap = 2, opacity = menu_opacity,
-					clip = '\\clip(' .. ax + 2 .. ',' .. title_ay .. ',' .. bx - 2 .. ',' .. ay .. ')',
-				})
+				if menu.search.query ~= '' then
+					ass:txt(bx - spacing, title_ay + (title_height / 2), 6, ass_escape(menu.search.query), {
+						size = self.font_size, color = bgt, wrap = 2, opacity = menu_opacity,
+						clip = '\\clip(' .. ax + spacing .. ',' .. title_ay .. ',' .. bx - spacing .. ',' .. ay .. ')',
+					})
+				else
+					ass:txt(bx - spacing, title_ay + (title_height / 2), 6, 'search for...', {
+						size = self.font_size, italic = true, color = bgt, wrap = 2, opacity = menu_opacity * 0.5,
+						clip = '\\clip(' .. ax + spacing .. ',' .. title_ay .. ',' .. bx - spacing .. ',' .. ay .. ')',
+					})
+				end
 			else
 				menu.ass_safe_title = menu.ass_safe_title or ass_escape(menu.title)
 				ass:txt(ax + menu.width / 2, title_ay + (title_height / 2), 5, menu.ass_safe_title, {
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -11,7 +11,7 @@ local Element = require('elements/Element')
 ---@alias MenuStackItem MenuStackValue|MenuStack
 ---@alias MenuStackValue {title?: string; hint?: string; icon?: string; value: any; active?: boolean; bold?: boolean; italic?: boolean; muted?: boolean; keep_open?: boolean; separator?: boolean; selectable?: boolean; align?: 'left'|'center'|'right'; title_width: number; hint_width: number}
 ---@alias Fling {y: number, distance: number, time: number, easing: fun(x: number), duration: number, update_cursor?: boolean}
----@alias Search {query: string; timeout: unknown; source: {scroll_y: number; selected_index?: integer; items?: MenuDataItem[]}}
+---@alias Search {query: string; timeout: unknown; width: number; top: number; source: {scroll_y: number; selected_index?: integer; items?: MenuDataItem[]}}
 
 ---@alias Modifiers {shift?: boolean, ctrl?: boolean, alt?: boolean}
 ---@alias MenuCallbackMeta {modifiers: Modifiers}
@@ -259,12 +259,14 @@ function Menu:update_dimensions()
 	local height_available = display.height - Elements.timeline.size_min
 
 	for _, menu in ipairs(self.all) do
-		menu.width = round(clamp(min_width, menu.max_width, display.width * 0.9))
+		local width = math.max(menu.search and menu.search.width or 0, menu.max_width)
+		menu.width = round(clamp(min_width, width, display.width * 0.9))
 		local title_height = (menu.is_root and menu.title) and self.scroll_step or 0
 		local max_height = round(height_available * 0.9 - title_height)
 		local content_height = self.scroll_step * #menu.items
 		menu.height = math.min(content_height - self.item_spacing, max_height)
-		menu.top = round((height_available - menu.height + title_height) / 2)
+		local search_top = menu.search and menu.search.top or height_available
+		menu.top = math.min(search_top, round((height_available - menu.height + title_height) / 2))
 		menu.scroll_height = math.max(content_height - menu.height - self.item_spacing, 0)
 		menu.scroll_y = menu.scroll_y or 0
 		self:scroll_to(menu.scroll_y, menu) -- clamps scroll_y to scroll limits
@@ -739,7 +741,7 @@ function Menu:search_start()
 		end))
 	end
 	menu.search = {
-		query = '', timeout = timeout,
+		query = '', timeout = timeout, width = menu.width, top = menu.top,
 		source = {
 			scroll_y = menu.scroll_y, selected_index = menu.selected_index,
 			items = not menu.on_search and menu.items or nil
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -1,16 +1,17 @@
 local Element = require('elements/Element')
 
 -- Menu data structure accepted by `Menu:open(menu)`.
----@alias MenuData {type?: string; title?: string; hint?: string; keep_open?: boolean; separator?: boolean; items?: MenuDataItem[]; selected_index?: integer;}
+---@alias MenuData {id?: string; type?: string; title?: string; hint?: string; keep_open?: boolean; separator?: boolean; items?: MenuDataItem[]; selected_index?: integer; on_search?: string|string[]|fun(search_text: string), search_debounce?: number|string}
 ---@alias MenuDataItem MenuDataValue|MenuData
 ---@alias MenuDataValue {title?: string; hint?: string; icon?: string; value: any; bold?: boolean; italic?: boolean; muted?: boolean; active?: boolean; keep_open?: boolean; separator?: boolean; selectable?: boolean; align?: 'left'|'center'|'right'}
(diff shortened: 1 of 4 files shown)


tomasklaen:
I have a bit different design in mind for search. It'll require a bit more work, which is why I didn't get to it yet. I understand if you don't have time and just want to wait for me to eventually do it :) But here it is:

---

### 1.

Search has to start on any unicode input while menu is open. No `ctrl+f` or `/` gates. Escape then cancels search before it closes the menu.

---

### 2.

I want there to be 2 search modes, internal, and external.

Internal is active when there's no `on_search` callback on the menu data, or menu options. It just seamlessly filters current items based in search query. Since every menu can use searching, this is better than making every menu implement its own filtering and updating mechanism, when all they need is to just filter current items.

For more advanced stuff, we have external mode, which is enabled by specifying `on_search` callback function on menu options, or callback command on menu data. In this mode, we just pass the query to the function/command and let outside forces update current items.

The way search should be implemented internally should be something like:

```lua
local menu = self.current

-- initial
menu.search = nil

-- on input
menu.search = {
    query = 'search text',
    -- in internal mode, we keep source data here to restore when search is canceled
    source = {
        selected_index = menu.selected_index,
        items = menu.items,
    }
}
```

`:update()` will need to inherit each menu's `search` data on updates.

Since each menu (submenu) can have its own search, the callbacks receive menu ID as 1st, and query as 2nd argument.

Menu IDs are defined with a new `id` property on each menu data, and they default to `'main'` for the main menu, and I guess `title` or an incrementing index for submenus.

Also, always preselect 1st item of the search result.

---

### 3. 

`search_style` (can't think of a better name) menu data option that controls how search is presented to the user. It can be one of:

- `disabled` - Search is not allowed, typing does nothing.
- `transparent` **(default)** - There's no search input until user starts typing. Empty search query (after escape or clearing query with backspace) hides the input and displays title again if any.
- `placeholder` - Always displays a search input while menu is open. When search query is empty, displays menu title as a placeholder text at 50% opacity. Example:
     image

Also, isn't there some vertical line character that doesn't have a visible width that could be used as a cursor? I recall looking for something like that but without luck. Alternatively we can use ASS underline (`\u1` enable, and `\u0` to disable).

---

### 4.

`search_debounce` menu data option that controls how often search filtering happens while user is typing in milliseconds.

This is needed menus that need to do some work, or fetch async stuff.

In internal search mode, it should default to `0`. In external mode to `300`.

It can also be a special string `'submit'`, which will only trigger search upon pressing `ctrl+enter`. This is needed to implement stuff like subtitle downloader.

---

### 5.

When `search_debounce` is `'submit'`, before we execute the search, we should replace current items with a placeholder loading indicator:

```lua
menu.items = {
  {
    title = 'Loading...',
    icon = 'spinner',
    italic = true,
    align = 'center',
    selectable = false,
    muted = true
  }
}
```

christoph-heinrich:
I shouldn't have submitted that right before going to bed, I completely forgot to mention that this wasn't supposed to be a complete implementation. The intention was to get initial pieces in place, simply because there hasn't been any progress on this front.

> 3. `search_style`

We can deal with the cursor the same way console does. It splits the text at the cursor position, then draws the part before the cursor, then cursor, then the part after the cursor. Then the same thing again but with the text visible to ensure that the cursor is in front of the text. It's a bit hacky but works well. The cursor is simply a line drawn with ass tags.

I didn't want to deal with the complexity the cursor brings (at least for now) and it's really not necessary for initial search support. I also thought of adding a search button in the top right corner of the menu, but just like the rest of styling and the cursor, it's not necessary for now.
That stuff can be a separate PR, be it by me or someone else.

I'll implement the rest though, those make sense.

christoph-heinrich:
There is still an issue that I have yet to solve, everything else should be fine.

1. open a menu with at least one submenu
2. search in that menu
3. open a submenu
4. search in that submenu
5. go back to it's parent
6. search there again

As a result the items in the submenu won't be filtered anymore.
That must have something to do with the original items getting stored in the search object and then the search based on those restores the items of the child.
It's pretty minor issue, unlikely to cause any problems during regular usage.
I'll have another look later or tomorrow.

scripts/uosc/elements/Menu.lua
```diff
@@ -150,6 +158,8 @@ function Menu:update(data)
 		local parent_id = menu.parent_menu and not menu.parent_menu.is_root and menu.parent_menu.id
 		if not menu.is_root then
 			menu.id = (parent_id and parent_id .. ' > ' or '') .. (menu_data.title or i)
+		else
+			menu.id = 'main'
```
tomasklaen, line 162: Add `id` to `MenuData` type so that menu implementers can define their own. Use it as 1st `if` statement here when defined:

```lua
if menu_data.id then
  menu.id = menu_data.id
elseif ...
```

scripts/uosc/elements/Menu.lua
```diff
@@ -608,6 +620,104 @@ function Menu:on_end()
...
+	end
+end
+
+function Menu:search_stop()
+	self:search_query_update('')
+	self.current.search = nil
+	self:search_ensure_keybinds()
+end
+
+function Menu:search_disable_key_bindings()
```
tomasklaen, line 710: I don't like the keybind juggling. It's fickle and leads to unstable states. Like searching menu, than going to submenu, searching that, than back to menu, canceling search there, then back to submenu, now when I try to cancel search here we instead exit the menu.

Rather, we bind all the bindings we need at init as we do now, and just fork the search specific handling in their handles if search exists.

And `ctrl+enter` is currently unused even when it's defined. It just falls back to normal open item handling, ignoring the ctrl modifier.

tomasklaen:
If user deletes the last letter with backspace, it should have the exact same effect as pressing escape. Meaning resetting `menu.search` to `nil`.

---

I don't like how the menu and the input is moving vertically when menu is shrinking by being reduced. I think we should save the original menu height on `menu.search.original_height` or something, and when updating dimensions, we calculate `ay` based on this height if it exists.

---

When searching menu that doesn't already have a title, we don't see any search input. Or are you just implementing searching logic here, and leaving visuals for me? Np if that's the case.

scripts/uosc/elements/Menu.lua
christoph-heinrich: The scenario you describe with canceling search is caused by that one issue I described, otherwise going back to the submenu and canceling search there would work as expected. That's why `search_ensure_keybinds()` exists.

And what do you mean that `ctrl+enter` ignores ctrl? I've tested the 'submit' search and it works as it should.

christoph-heinrich:
> I don't like how the menu and the input is moving vertically when menu is shrinking by being reduced. I think we should save the original menu height on `menu.search.original_height` or something, and when updating dimensions, we calculate `ay` based on this height if it exists.

But then a subtitle downloader would start a small menu, use the search and then list all results in that small menu.
We could use the max of both heights.

> When searching menu that doesn't already have a title, we don't see any search input. Or are you just implementing searching logic here, and leaving visuals for me? Np if that's the case.

This was intended to be logic only. Once that's in place the visuals can be dealt with in a separate change.

christoph-heinrich:
- Scripts can specify menu.id now.
- Keys relevant to the search are now sent to an intermediate function that decides what to do based on if there is a search or not. That additional function is required because some menu actions can be caused by multiple keybindings, but not all of them are relevant to the search.
- An empty search string now stops the search.
- Menu size remains the same while searching.
- Menus always show their title bar during an active search.

tomasklaen:
I've meant for the max menu height to only affect the `ay` coordinate of the menu, not the menu height. Leaving empty background with no items as it is now feels wrong/broken to me.

---

After thinking about it a bit, the `search_debounce` and `on_search` should be per (sub)menu options, not global for the whole menu stack. That way we can have top menu that has a special external search that populates items, but its submenus can still use internal search, and vice versa. This means completely removing `self.opts.on_search`, and only using `menu_data.on_search`, which internally should also accept functions.

We also don't need to do the menu ID stuff, nor pass it as a parameter anymore.

What to do on search should be completely contained in `Menu:search_submit()`:

- `on_search` is `nil` - use internal
- `on_search` is `function` - execute `on_search(query)`
- `on_search` is `string | string[]` - execute command with query as last

`on_search` handling that is currently inside `open_command_menu()` can also be removed.

---

Don't normalize `search_debounce == 'submit'` to `nil`. Only to either `'submit'`, or `number`. Then inside `Menu:search_query_update()`, do:

```lua
function Menu:search_query_update(query)
	local menu = self.current
	menu.search.query = query
	if menu.search_debounce ~= 'submit' then
		if menu.search.timeout then
			menu.search.timeout:kill()
			menu.search.timeout:resume()
		else
			self:search_submit(menu)
		end
	end
	request_render()
end
```

This way the 3rd logic path here is obvious. Before you had to study the code to realize it can be `nil`, and what it means.

---

I haven't tested it, but I think there's a race condition where by the time timer calls `:search_submit()`, the search might be deleted by `:search_stop()` if user pressed escape.

christoph-heinrich:
- menu.top (effectively ay) gets remembered instead of menu.height
- `search_debounce` and `on_search` are local to the menu (that also makes it easier for scripts :+1: )
- `on_search` is handled in `submit_search()`
- `search_debounce` gets normalized to `'submit'` instead of nil

Good catch with the race condition, I didn't confirm it in testing but it seems obvious in retrospect.
`search_submit()` now makes sure there is a search to prevent that.

christoph-heinrich:
Backspace is now repeatable, but doesn't cause a `Menu:back()` when the whole search gets deleted by holding it down.

tomasklaen:
Can't use numpad for search inputs. Keys `KP0` through `KP9`, and `KP_DEC` (dot) are for some reason not part of the `any_unicode` group :(. Rest of the numpad keys (+-*/) seem to work fine.

Otherwise it's working great. Good job 👍.

christoph-heinrich:
I just noticed that searching whenever something gets typed breaks toggling the menu.
For example I have the chapter menu on `C` an I used to be able to quickly look at the chapters and then close them by keeping my finger on the c and shift keys, now that searches for `'C'`.
Toggling quality-menu has the same problem now.



> Can't use numpad for search inputs. Keys `KP0` through `KP9`, and `KP_DEC` (dot) are for some reason not part of the `any_unicode` group 😞. Rest of the numpad keys (+-*/) seem to work fine.

We could add those keys here, but that sounds like it should be fixed in mpv itself.
Of course that would mean it only works in the latest mpv version.

tomasklaen:
> I just noticed that searching whenever something gets typed breaks toggling the menu.

The only fix for that is going back to a shortcut to start searching. So what is more annoying to you: having to press a shortcut to search, or having to always press esc to close the menu?

Maybe a `type_to_search=yes/no` option?

> We could add those keys here, but that sounds like it should be fixed in mpv itself. Of course that would mean it only works in the latest mpv version.

I guess we have to.

christoph-heinrich:
> I guess we have to.

I'll fix it in both.
console.lua has extra binds for those apparently.

christoph-heinrich:
As it turns out changing `any_unicode` for the keypad keys in mpv isn't as trivial as initially assumed, and I'm not even sure anymore that it should be changed. Maybe I'll do it at some point in the future.

- Keypad keys work
- new option `menu_type_to_search=yes/no`

I didn't like how the previous `search_disable_key_bindings()` worked, so instead there is a `self.key_bindings_search` now that is used for those temporary bindings.

During testing I noticed that `selected_index` was never restored when the search ended, and that lead me to realize that if we remember that, we should also remember `scroll_y`.

I don't know how restoring those should work for scripts, because all we do is inform them about an empty search and they have to then update the menu. They are able to set a `selected_index` but there is no way for them to specify where to scroll to, and there is also no way for them to know what was selected and how far down scroll was before the search.
We could leave it as is because it's really not that big of a deal.

Right before pushing I realized that runtime option updates don't play well with the way I implemented the keybindings and I don't have time to solve that properly right now. The option value simply gets stored in the menu itself upon opening for now, so to get updated behavior a new menu needs to be opened.

Also during testing I noticed this fun structure that comes form remembering the top position. Somehow I never noticed that before.
image

tomasklaen:
In `menu_type_to_search` comment, you have a `of if` typo. Also add there:

```
When enabled, you can no longer toggle a menu off with the same key that opened it, if the key is a unicode character.
```

Also `despide` -> `despite`.

---

> I don't know how restoring `scroll_y`, `selected_index` should work for scripts...

In non immediate/internal searches, I think we'll almost always be dealing with some loading time, so the list will be replaced with a loading indicator. I think it's expected to loose scroll/selected states there.

> ...runtime option updates don't play well with the way I implemented the keybindings...

It's good enough as is. I don't want runtime option updates to increase code complexity too much.

> I noticed this fun structure that comes form remembering the top position...

That kind of already happens when parent is tall, child is short, and it's the 1st parent item that has the child submenu. I thought about setting some `ay_override` on the child menu inside render function to position it next to the hovered/selected item, but I think that'd lead to more issues or super spaghetti code.

---

Do we want to even support cursor for search input? Like typing, then pressing left to move cursor left, and continue typing inside the word? Because right now left/right arrows go to parent/child menu, and it'll clash with it.

But the cursor might be useful. For example, when implementing subtitle search, the way how I'd envision it to work is you open the subtitle search menu, and the input will be pre-populated with filename, but the user might want to delete parts of that filename before submitting search.

Dunno how to reconcile these two.

christoph-heinrich:
Comments are fixed

> Also `despide` -> `despite`.

That typo made me chuckle :spider: 

> > ...runtime option updates don't play well with the way I implemented the keybindings...
> 
> It's good enough as is. I don't want runtime option updates to increase code complexity too much.

It shouldn't take much, a `search_ensure_key_bindings()` in `on_options()` and `search_ensure_key_bindings()` then has to also take care of `/` and `ctrl+f`. Maybe I'll give changing it a try later today or tomorrow to see if it'll actually be that simple.

> Do we want to even support cursor for search input?...
> But the cursor might be useful. For example, when implementing subtitle search...

It would certainly be useful for that, but when we have a cursor we also have to make sure that the cursor stays within the visible area, so we'll have to scroll the text. That could be done with the text width before and after the cursor, but those widths aren't exact so the cursor will jump around a little bit.

Always loosing the ability to navigate with arrow keys during a search might be annoying, so maybe there should be a flag to decide if the search is simple like right now or a full blown text box. If the text box looks like in that image of yours it would be visually distinct enough for users to realize that the arrow keys are now used for that.

In that case we could even switch between navigating the menu and typing in the text box. E.g. after submitting the search with `ctrl+enter` the arrow keys can be used for navigation again and text input is ignored, if the user wants to edit the search again they have to press `ctrl+f` or `/` to activate the text box and capture those keys.

For now I'd leave it as is.

tomasklaen:
Yeah makes sense.

So the things menu searching is missing now are:

- A flag that tells the menu that it's intended to be primarily searchable/filterable, where the search is always active, and query input never disappears. I think maybe `MenuData.palette: boolean`? Can't think of a better name.
- Making the input pretty. Honestly this just needs switching the background from rendering white texture to just a 1px border, making the text match the title in size but not bold, and have some padding from the border.
- Cursor rendering and control.

Do you want any of these to be part of this PR? I'm fine either way and merging as is and leaving these for later.

christoph-heinrich:
I tried replicating the design in that image.
image
image

> Do you want any of these to be part of this PR? I'm fine either way and merging as is and leaving these for later.

I'll leave the cursor and the flag for later.

junguler:
any chance for a fuzzy finding option? that would make using this feature way easier and less likely to skip items due to simple grammar issues

also hitting backspace one too many times will close the menu completely which is way too common for me so it would be nice if backspace was not able to close the menu when the search is active

oops i've made this comment on the wrong pr, just ignore that one and sorry about that

christoph-heinrich:
@junguler If you want to delete the whole input, hold backspace down and it won't close the menu.

Fuzzy search sounds nice, but no idea how to implement that. It would be an interesting problem to solve, no idea when I'd get to it though.

junguler:
thanks for your reply, good to know about holding backspace, makes sense

as for fuzzy searching, i'm glad to see you are interested in it and am willing to wait for as much as it would take

christoph-heinrich:
I've looked into the fuzzy search thing a little bit and it seems that instead of filtering we'd have to sort items based on some calculated score. We could try  introduce some sort of metric for cutting off less relevant results, but we'd have to try it and see how well that works.

@tomasklaen thoughts?

tomasklaen:
We are just filtering a single list. I really don't want to bloat it with complicated search algorithms. You usually get what you need after 2-3 letters in, and it's way too easy and fast to correct a typo. This doesn't seem worth it at all to me.

But I wouldn't be against constructing a keyword from the first letter of each word in item's `title+hint`, and comparing that query against it too, so that you can match stuff like "Stream quality" with "sq". That is simple enough and a nice feature to have.

junguler:
this would work but i think if there was a way to specify this delimiter it would be very helpful, these delimiters could be spaces, dashes, underlines and the such ...

christoph-heinrich:
fwiw our of curiosity I tried something simple based the sub-string Hamming distance, in case you want to try it out.
It cuts off distances longer then 1/3 query length.

```diff
diff --git a/scripts/uosc/elements/Menu.lua b/scripts/uosc/elements/Menu.lua
index 07d30ed..14f3108 100644
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -645,13 +645,30 @@ function Menu:search_update_items()
 	self:reset_navigation()
 end
 
+local function fuzzy_sort(items, query)
+	local q, q_n = utf8_chars(query)
+
+	local t, n, threshold = {}, 1, q_n / 3 + 1e-9
+	for _, item in ipairs(items) do
+		local td = item.title and hamming_distance(item.title:lower(), q, q_n) or q_n
+		local hd = item.hint and hamming_distance(item.hint:lower(), q, q_n) or q_n
+		local distance = td < hd and td or hd
+		if distance <= threshold then
+			t[n] = { distance, n, item }
+			n = n + 1
+		end
+	end
+	table.sort(t, function(a, b) return a[1] == b[1] and a[2] < b[2] or a[1] < b[1] end)
+	for i, e in ipairs(t) do
+		t[i] = e[3]
+	end
+	return t
+end
+
 ---@param menu MenuStack
 function Menu:search_internal(menu)
 	local query = menu.search.query:lower()
-	menu.items = query ~= '' and itable_filter(menu.search.source.items, function(item)
-		return item.title and item.title:lower():find(query, 1, true) or
-			item.hint and item.hint:lower():find(query, 1, true)
-	end) or menu.search.source.items
+	menu.items = query ~= '' and fuzzy_sort(menu.search.source.items, query) or menu.search.source.items
 	self:search_update_items()
 end
 
diff --git a/scripts/uosc/lib/text.lua b/scripts/uosc/lib/text.lua
index 241ed70..23dcf81 100644
--- a/scripts/uosc/lib/text.lua
+++ b/scripts/uosc/lib/text.lua
@@ -74,6 +74,33 @@ local function utf8_iter(str)
 	end
 end
 
+function utf8_chars(s)
+	local t, i = {}, 0
+	for _, c in utf8_iter(s) do
+		i = i + 1
+		t[i] = c
+	end
+	return t, i
+end
+
+function hamming_distance(text, p, p_n)
+	local t, t_n = utf8_chars(text)
+
+	local best_score = p_n
+	for i = 1, t_n - p_n + 1 do
+		local score = 0
+		for j = 1, p_n do
+			if t[i + j - 1] ~= p[j] then
+				score = score + 1
+			end
+		end
+		if score < best_score then
+			best_score = score
+		end
+	end
+	return best_score
+end
+
 ---Extract Unicode code point from utf-8 character at index i in str
 ---@param str string
 ---@param i integer
```

Something based on the Levenshtein distance would be nicer, but also more work.

PR #627, opened 2023-09-16
fix: don't select not selectable items

christoph-heinrich:
It was possible for not selectable items to be selected via page up/down as well as when updating the menu.

Page down now goes to the next best selectable item after adding to the index, like is the case for `next()`, with a fallback to an earlier index.
Page up is the same as page down but inverted direction.

Updating the menu now chooses the closes selectable item to the intended index.

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -185,18 +185,7 @@ function Menu:update(data)
 		local old_menu = self.by_id[menu.is_root and '__root__' or menu.id]
 		if old_menu then table_assign(menu, old_menu, {'selected_index', 'scroll_y', 'fling'}) end
 
-		if menu.selected_index then
-			local index = clamp(1, menu.selected_index, #menu.items)
-			local prev_index =
-				itable_find(menu.items, function(item) return item.selectable ~= false end, index, 1)
-			local next_index =
-				itable_find(menu.items, function(item) return item.selectable ~= false end, index)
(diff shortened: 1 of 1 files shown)


tomasklaen:
There's pretty much almost the same code in 4+ places now. I think we should consolidate it into 2 methods and then use that in all `:prev()`, `:next()`, `:on_pgup()`, `:on_pgdown()`,`:home()`, `:end()`, and reload. Something like:

```lua
function Menu:select_by_offset(offset)
    -- ...
end

function Menu:navigate_by_offset(offset)
    menu.selected_index = self:select_by_offset(offset)
    if menu.selected_index then self:scroll_to_index(menu.selected_index) end
end
```

Where offset dictates how close to the current `selected_index` should search for selectable start, and the sign (+/-) in which direction to move. Usage:

```lua
self:navigate_by_offset(1) -- next, navigates forward
self:navigate_by_offset(-1) -- prev, navigates backward
self:navigate_by_offset(items_per_page) -- pgdown
self:navigate_by_offset(-items_per_page) -- pgup
self:navigate_by_offset(Infinity) -- end
self:navigate_by_offset(-Infinity) -- home
```

And inside `:update()`:

```lua
menu.selected_index = self:select_by_offset(menu, 0) -- starts at current index
```

christoph-heinrich:
Page up/down are off by one compared to what they did previously when pressed while no item is selected, but that shouldn't be a problem.

PR #635, opened 2023-09-20
fix: shuffle history couldn't go back more than once

tomasklaen:
Shuffle now has its own history which inherits from `state.history` to not break the prev button.

It can populate the history with random items in both directions.

When shuffle is disabled, it clears its history (so all next states if any), but next time it's enabled it'll inherit `state.history` for its prev states again.

Did I miss anything?

ref #635

### DIFF
--- a/scripts/uosc/lib/utils.lua
+++ b/scripts/uosc/lib/utils.lua
@@ -402,11 +402,18 @@ function decide_navigation_in_list(paths, current_index, delta)
 	-- and removes all paths in it from the potential shuffle pool. This guarantees
 	-- no path repetition until at least 80% of the playlist has been exhausted.
 	if state.shuffle then
-		-- Going backward recalls from history.
-		if delta < 0 then
-			local path = state.history[#state.history + delta]
-			local index = itable_index_of(paths, path)
-			if path and index then return index, path end
+		state.shuffle_history = state.shuffle_history or {
+			pos = #state.history,
+			paths = itable_slice(state.history)
+		}
+		state.shuffle_history.pos = state.shuffle_history.pos + delta
+		if state.shuffle_history.pos < 1 or state.shuffle_history.pos > #state.shuffle_history.paths + 1 then
+			state.shuffle_history.pos = clamp(1, state.shuffle_history.pos, #state.shuffle_history.paths + 1)
+		else
+			local history_path = state.shuffle_history.paths[state.shuffle_history.pos]
+			if history_path then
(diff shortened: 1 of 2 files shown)


tomasklaen:
> I'm not stoked on introducing that new mechanism with on_* in the state. I'd rather keep it simple and handle that in the shuffle callback.

I've had it there originally, but CycleButtons in controls are only triggering `set_state()`, so there's no `shuffle` message/command.

> The return value from itable_index_of() needs to be checked, otherwise it might load a file from the history that isn't even in the playlist.

Yeah, I was thinking about `prev` as "I want to play the previous file", so I was OK with that. But you're right, it definitely should pick only one of the passed paths.

christoph-heinrich:
> I've had it there originally, but CycleButtons in controls are only triggering `set_state()`, so there's no `shuffle` message/command being triggered.

We could give a command button the ability to appear active based on some property (like the current @ thing for badges), then scripts could control if "their" button appears active or not.
Or create a main element that simply exists to listen to such events.
Or leave it as is :)

christoph-heinrich (approved):

christoph-heinrich:
We should leave a note that the shuffle history only works correctly with a `|delta| < 1`... or fix it.
Positive delta isn't a problem, but the negative one breaks. I think we'd have to look for the last entry that is still in our playlist.

tomasklaen:
I don't understand. Did you forget `=` in `<=`? In which case yeah, `clamp(-1, delta, 1)` might be fine, but there's nothing currently that passes delta bigger than 1 in the code. The parameter exists just to specify forward or backward pretty much. But sure I'll add it.

christoph-heinrich:
> I don't understand. Did you forget `=` in `<=`?

Yes I did.

Well my thinking was that at some point in the future there might some code path that ends up passing e.g. `-3` for `delta`.
Be it a "go back n position" or whatever, doesn't really matter. It's easy to miss that this doesn't work correctly with `delta < -1`, especially when there isn't even a comment about it.

With not working correctly I mean e.g.:
`delta` is `-3`, but there are only 2 files from the playlist before the current position, therefore `next_index` will be `nil`, which results in a new random file getting inserted at `pos` instead of `pos-2` as one would expect.

christoph-heinrich (approved):

junguler:
any chance for a fuzzy finding option? that would make using this feature way easier and less likely to skip items due to simple grammar issues

also hitting backspace one too many times will close the menu completely which is way too common for me so it would be nice if backspace was not able to close the option when the search is active

PR #641, opened 2023-09-22
fix: stream quality selection

christoph-heinrich:
Due to a recent change in mpv writing to `playlist-pos` doesn't restart playback anymore, which was necessary for the format change to work for the current file.
https://github.com/mpv-player/mpv/commit/4b2276b730c8542649b665aaec187e1c7e3b2612


tomasklaen:
Can you remove the now unnecessary comment above it?

christoph-heinrich:
Sure.

PR #643, opened 2023-09-22
fix: adapt to the new title bar state of mpv

dyphire:
mpv has added a new option to control window title bar state, which requires relevant adjustments to be added in uosc.
Ref https://github.com/mpv-player/mpv/commit/ee24dd04194aadfda882fd73553cba8d1e734567

### DIFF
--- a/scripts/uosc/elements/TopBar.lua
+++ b/scripts/uosc/elements/TopBar.lua
@@ -71,7 +71,7 @@ end
 
 function TopBar:decide_enabled()
 	if options.top_bar == 'no-border' then
-		self.enabled = not state.border or state.fullscreen
+		self.enabled = not state.border or state.title_bar == false or state.fullscreen
 	else
 		self.enabled = options.top_bar == 'always'
 	end
@@ -140,6 +140,11 @@ function TopBar:on_prop_title() self:decide_titles() end
 function TopBar:on_prop_alt_title() self:decide_titles() end
 
 function TopBar:on_prop_border()
+	self:decide_enabled()
+	self:update_dimensions()
+end
+
+function TopBar:on_prop_title_bar()
 	self:decide_enabled()
 	self:update_dimensions()
 end
--- a/scripts/uosc/elements/WindowBorder.lua
+++ b/scripts/uosc/elements/WindowBorder.lua
@@ -12,10 +12,12 @@ end
 
 function WindowBorder:decide_enabled()
 	self.enabled = options.window_border_size > 0 and not state.fullormaxed and not state.border
+	    and state.title_bar == false
 	self.size = self.enabled and options.window_border_size or 0
 end
 
 function WindowBorder:on_prop_border() self:decide_enabled() end
+function WindowBorder:on_prop_title_bar() self:decide_enabled() end
 function WindowBorder:on_prop_fullormaxed() self:decide_enabled() end
 function WindowBorder:on_options() self:decide_enabled() end
 
--- a/scripts/uosc/elements/Controls.lua
+++ b/scripts/uosc/elements/Controls.lua
@@ -328,6 +328,7 @@ end
 function Controls:on_dispositions() self:reflow() end
 function Controls:on_display() self:update_dimensions() end
 function Controls:on_prop_border() self:update_dimensions() end
+function Controls:on_prop_title_bar() self:update_dimensions() end
 function Controls:on_prop_fullormaxed() self:update_dimensions() end
 function Controls:on_timeline_enabled() self:update_dimensions() end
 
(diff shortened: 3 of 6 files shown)


natural-harmonia-gropius:
Is Topbar show/hide correctly when `cycle title-bar` / `cycle border`?

dyphire:
> Is Topbar show/hide correctly when `cycle title-bar` / `cycle border`?

Ah, this really doesn't work. It seems that the attributes in uosc are not updated correctly.

tomasklaen:
You need to add:

```lua
function Element:on_prop_title_bar() self:decide_enabled() end
```

So that the concerned element updates after `title_bar` changes.

dyphire:
> You need to add:
> 
> ```lua
> function ElementName:on_prop_title_bar() self:decide_enabled() end
> ```
> 
> So that the concerned element updates after `title_bar` changes.

Done

scripts/uosc/elements/TopBar.lua
```diff
@@ -71,7 +71,7 @@ end
 
 function TopBar:decide_enabled()
 	if options.top_bar == 'no-border' then
-		self.enabled = not state.border or state.fullscreen
+		self.enabled = not state.border or (state.title_bar ~= nil and not state.title_bar) or state.fullscreen
```
christoph-heinrich, line 74:
```suggestion
		self.enabled = not state.border or not state.title_bar or state.fullscreen
```
`nil` is treated as false, no need to check for it specifically.

scripts/uosc/elements/WindowBorder.lua
```diff
@@ -12,10 +12,12 @@ end
 
 function WindowBorder:decide_enabled()
 	self.enabled = options.window_border_size > 0 and not state.fullormaxed and not state.border
+	    and (state.title_bar ~= nil and not state.title_bar)
```
christoph-heinrich, line 15: same as above

christoph-heinrich (changes requested):

scripts/uosc/elements/TopBar.lua
dyphire: No, since `nil` is treated as false, this will always make the trigger condition true.
This does not meet expectations and is not what we want.

christoph-heinrich: Why would you want to hide the top bar when it's `nil`?

christoph-heinrich:
```suggestion
		self.enabled = not state.border or state.title_bar or state.fullscreen
```
That `not` in there confused me, I think this is what we want.

christoph-heinrich:
```suggestion
		self.enabled = not state.border or state.title_bar == false or state.fullscreen
```
Then this or not?

dyphire: I missed it.

christoph-heinrich:
Sorry for realizing this so late, but this causes me to not have window borders (because `title-bar=yes`).

Why does the `title-bar` property have to influence the border?

dyphire:
> Sorry for realizing this so late, but this causes me to not have window borders (because `title-bar=yes`).
> 
> Why does the `title-bar` property have to influence the border?

On Windows, both behaviors of `no-title-bar`and are`no-border` implemented without borders, so I did the same processing for `title-bar` property as `border`. 
Due to this being a property only on Windows, it seems to have caused some issues on Linux and needs to be fixed.

christoph-heinrich:
Isn't there a way on windows to have no title bar but still have borders?

christoph-heinrich:
I'm so confused here with your previous statement...
Let's try to make it clear

`title-bar=yes` and `border=yes` does have window borders
`title-bar=no` and `border=yes` does have window borders
`title-bar=yes` and `border=no` does **not** have window borders
`title-bar=no` and `border=no` does **not** have window borders

Is that correct?

dyphire:
> `title-bar=no` and `border=yes` does have window borders

Under this combination, mpv has an invisible window border. Although it has the same appearance as `title-bar=yes` and `border=no`, the difference is that it can be captured by Windows window captors.This is a special window behavior on Windows, which can indeed be confusing.

christoph-heinrich:
> Under this combination, mpv has an invisible window border. Although it has the same appearance as `title-bar=yes` and `border=no`, the difference is that it can be captured by Windows window captors.

When searching for "Windows window captors" I find pretty much nothing. I assume they are screenshot and screen recording things?
So the user doesn't see a window border but they do show up when the window gets captures by a different application?

WTF, why would they do that? That makes absolutely no sense. Title bar and border could be completely orthogonal to each other and instead they do that...

dyphire:
> > Under this combination, mpv has an invisible window border. Although it has the same appearance as `title-bar=yes` and `border=no`, the difference is that it can be captured by Windows window captors.
> 
> When searching for "Windows window captors" I find pretty much nothing. I assume they are screenshot and screen recording things? So the user doesn't see a window border but they do show up when the window gets captures by a different application?
> 
> WTF, why would they do that? That makes absolutely no sense. Title bar and border could be completely orthogonal to each other and instead they do that...

My fault is actually the window manager on Windows. This is actually very useful.

PR #644, opened 2023-09-23
feat: search first character of each word

christoph-heinrich:
I don't really understand why title and hint characters should be concatenated.
It might make some sense when the title is only one word, but otherwise it might lead to unexpected results.

I tried also a different implementation based on `find` instead of `gmatch` that I expected to be faster, because it only allocated half as many strings, but it turned out to be the same speed.

Ref. https://github.com/tomasklaen/uosc/pull/625#issuecomment-1732069651

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -649,8 +649,11 @@ end
 function Menu:search_internal(menu)
 	local query = menu.search.query:lower()
 	menu.items = query ~= '' and itable_filter(menu.search.source.items, function(item)
-		return item.title and item.title:lower():find(query, 1, true) or
-			item.hint and item.hint:lower():find(query, 1, true)
+		local title = item.title and item.title:lower()
+		local hint = item.hint and item.hint:lower()
+		return title and title:find(query, 1, true) or hint and hint:find(query, 1, true) or
+			title and table.concat(first_word_chars(title)):find(query, 1, true) or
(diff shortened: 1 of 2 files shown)


tomasklaen:
It matches non-word characters currently. For example for string "Delete file & next" it'll create a keyword `df&n` instead of `dfn`.

tomasklaen:
And yeah, title and hint probably shouldn't be concatenated, but matched separately.

christoph-heinrich:
We could go with a simple `%w`, but that doesn't work well with non latin alphabets.
We could do something similar to what we're doing for word wrapping and make a list of characters that separate words.

christoph-heinrich:
I've made a list with word separators, I hope I got everyone's favorite word separation characters in there.

tomasklaen:
Add `+&()[]{}`.

And I guess put`word_separators` outside of the function since it's static.

christoph-heinrich:
`wrap_text` also has some static lists, idk if you want to do the same thing there. If so I'll do that in the other PR that already touches `wrap_text`.

tomasklaen:
Oh, and `/\`, sorry 😐.

> wrap_text also has two static lists, idk if you want to do the same thing there

Yeah that would be nice.

christoph-heinrich:
I missed so many characters :disappointed: 
Actually `<>` should probably also be on that list.

junguler:
amazing work guys, i really appreciate it

PR #645, opened 2023-09-23
feat: fuzzy search based on Levenshtein distance

christoph-heinrich:
I don't know if that will ever go anywhere, but it's easier to test this way and it's a dedicated place to discuss this.

Ref. https://github.com/tomasklaen/uosc/pull/625#issuecomment-1730741965

Depends on #644, but doesn't add text and hint characters together.

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -755,6 +755,8 @@ function Menu:search_internal(menu)
 	local query = menu.search.query:lower()
 	menu.items = query ~= '' and fuzzy_search(menu.search.source, query) or menu.search.source.items
 	self:search_update_items()
+	self:select_index(1, menu)
+	self:scroll_to_index(1, menu)
 end
 
 ---@param menu? MenuStack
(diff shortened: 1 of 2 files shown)


tomasklaen:
I find it more confusing then helpful. For example, trying to match word `upscale` while leaving out letter `p` produces:

`u` - matches
`us` - nothing (I'd give up here and rewrite/fix my query)
`usc` - matches again

I'm really fine with what we have + #644. Implementing fulltext search features for a dumb single level list of short phrases is waay too much. I'd prefer a simple search implementation with slimmer `menu.lua` file over having this in. Please don't waste your time working on this :)

christoph-heinrich:
It's less confusing without the cutoff.

christoph-heinrich:
I like wasting my time :upside_down_face: 

As it turns out Levenshtein is **much** slower then Hamming, and the query length makes a huge difference. It should be fine for anything but extreme use-cases and I'd expect the query to typically be fairly short anyway.

That's probably as good as a fuzzy search gets.

PR #648, opened 2023-09-24
fix: render spaces at the end of the search query

christoph-heinrich:
Currently the text does not advance to the left when spaces are entered, which must mean that libass trims any trailing spaces. Appending a ZWNBSP at the end prevents any trimming.

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -1089,7 +1089,9 @@ function Menu:render()
 			-- Title
 			if menu.search then
 				if menu.search.query ~= '' then
-					ass:txt(bx - spacing, title_ay + (title_height / 2), 6, ass_escape(menu.search.query), {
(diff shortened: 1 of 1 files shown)


tomasklaen:
Add a comment what the line is doing or I'll remove it in future :)

christoph-heinrich:
Good point.

scripts/uosc/elements/Menu.lua
```diff
@@ -1089,7 +1089,9 @@ function Menu:render()
 			-- Title
 			if menu.search then
 				if menu.search.query ~= '' then
-					ass:txt(bx - spacing, title_ay + (title_height / 2), 6, ass_escape(menu.search.query), {
+					-- Add a ZWNBSP suffix to prevent libass from timming any trailing spaces
```
tomasklaen, line 1092: `timming` -> `trimming`

PR #650, opened 2023-09-27
feat: redesigned menu title & search query

tomasklaen:
- simpler & less cluttered design without textures
- search icon
- clickable icon for submittable searches
- search cursor
- localized placeholders
- clicking empty title now does nothing instead of closing the menu

menu_title_redesign

Anyone see any issues or something missing?

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -7,7 +7,7 @@ local Element = require('elements/Element')
 ---@alias MenuOptions {mouse_nav?: boolean; on_open?: fun(); on_close?: fun(); on_back?: fun(); on_move_item?: fun(from_index: integer, to_index: integer, submenu_path: integer[]); on_delete_item?: fun(index: integer, submenu_path: integer[])}
 
 -- Internal data structure created from `Menu`.
----@alias MenuStack {id?: string; type?: string; title?: string; hint?: string; selected_index?: number; keep_open?: boolean; separator?: boolean; items: MenuStackItem[]; on_search?: string|string[]|fun(search_text: string), search_debounce?: number|string; parent_menu?: MenuStack; submenu_path: integer[]; active?: boolean; width: number; height: number; top: number; scroll_y: number; scroll_height: number; title_width: number; hint_width: number; max_width: number; is_root?: boolean; fling?: Fling, search?: Search}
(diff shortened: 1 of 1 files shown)


scripts/uosc/elements/Menu.lua
```diff
@@ -1065,49 +1065,63 @@ function Menu:render()
...
-					size = self.font_size, bold = true, color = bg, wrap = 2, opacity = menu_opacity,
-					clip = '\\clip(' .. ax + 2 .. ',' .. title_ay .. ',' .. bx - 2 .. ',' .. ay .. ')',
+				ass:txt(rect.cx, rect.cy, 5, menu.ass_safe_title, {
+					size = self.font_size, bold = true, color = bgt, wrap = 2, opacity = menu_opacity,
+					clip = '\\clip(' .. rect.ax + 2 .. ',' .. rect.ay .. ',' .. rect.bx - 2 .. ',' .. rect.by .. ')',
 				})
 			end
+
+			-- Do nothing when user clicks title
+			if is_current and not cursor.on_primary_down and get_point_to_rectangle_proximity(cursor, rect) then
```
christoph-heinrich, line 1122: You're checking `cursor.on_primary_down` to not overwrite the icon event, but if some element below the menu already set an event we should still overwrite it.

Also maybe clicking the title should start a search?
On the other hand having a place to click while the menu is open might be nice for focusing the window.

christoph-heinrich:
I liked the look of the texture, but the increased contrast might be good for usability...
The icon and cursor are nice, but no border around the text field? (Does it look too busy together with icon and cursor?)

tomasklaen:
> if some element below the menu already set an event we should still overwrite it.

I've realized we are already overwriting cursor events at the top of `Menu:render()`.

> Also maybe clicking the title should start a search?

Noo, none would expect that to happen. When people click text, they want to select it, or reposition cursor. If we are not doing any of that, than it should do nothing.

> I liked the look of the texture, but the increased contrast might be good for usability...

I didn't :) I just wanted it to be distinct from active item. But I like this a lot more.

> no border around the text field? (Does it look too busy together with icon and cursor?)

Yes, I don't want borders bordering borders anymore. This is clean. Presence of icon+cursor clearly indicates this is a search input now. Doesn't look cluttered at all. Would definitely feel like it with more borders.

christoph-heinrich:
> > Also maybe clicking the title should start a search?
> 
> Noo, none would expect that to happen. When people click text, they want to select it, or reposition cursor. If we are not doing any of that, than it should do nothing.
> 
> > I liked the look of the texture, but the increased contrast might be good for usability...
> 
> I didn't :) I just wanted it to be distinct from active item. But I like this a lot more. There's no visual clashing of menu title and active item anymore.
> 
> > no border around the text field? (Does it look too busy together with icon and cursor?)
> 
> Yes, I don't want borders bordering borders anymore. This is clean. Presence of icon+cursor clearly indicates this is a search input now. Doesn't look cluttered at all. Would definitely feel like it with more borders.

Fair enough.

> The only thing I'm still not happy about is that the text is on the right, but I don't feel like implementing all that would be needed to align it to the left :)

It's not great, but I have no idea how we could make left alignment even work seamlessly for strings wider then the menu.

PR #651, opened 2023-09-27
fix: search not selecting 1st item in results

tomasklaen:
@christoph-heinrich did I break something? :)

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -630,32 +630,26 @@ function Menu:on_end()
 	self:navigate_by_offset(INFINITY)
 end
 
-function Menu:search_update_items()
-	local menus_to_serialize = {self.root}
-
-	for _, menu in ipairs(menus_to_serialize) do
-		for _, item in ipairs(menu.items or {}) do
-			if item.items then menus_to_serialize[#menus_to_serialize + 1] = item end
-		end
-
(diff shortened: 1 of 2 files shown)


christoph-heinrich:
That's much more of a problem with the fuzzy search then with the regular one, which is why I already made a [fix for that](https://github.com/tomasklaen/uosc/pull/645/commits/69e57a2243f6bd7460ff241afbda65b2037343a0) in there, but using `select_by_offset()` instead of `select_index()` is better.

But that makes me wonder if we should even include non selectable items in the search results? If we filter them out then `select_index()` is also fine.
Either way we should also scroll to the top when selecting the first item.

I'm not sure if the `reset_navigation()` is actually needed. I had it in the update function as a remnant of the original `update()`, but looking at it now the only useful thing it does for us is `self:scroll_to(menu.scroll_y)` and even that is for the current menu instead of the one passed into the search.
(Sidenote: `reset_navigation()` uses `scroll_to()` for clamping, wouldn't `set_scroll_to()` be a better fit?)

Other then that it seems fine. I was already wondering if there is any point in what was left of `search_update_items()`, so I'm all for it.

tomasklaen:
> if we should even include non selectable items in the search results?

True. I removed them.

christoph-heinrich (approved):
You might want to squash the second commit, maybe separate the non search changes into a separate commit.
But other then that LGTM.

PR #652, opened 2023-09-28
feat: palette menus

tomasklaen:
Each (sub)menu can now enable palette mode by setting its `palette` property to `true`.

In this mode:

- search input is always visible, doesn't have to be enabled, and can't be disabled
- `title` is used as input placeholder while search is empty

Any suggestions/bugs/comments?

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -219,7 +219,7 @@ function Menu:update(data)
 	-- Ensure palette menus have active searches, and clean empty searches from menus that lost the `palette` flag
 	local update_dimensions_again = false
 	for _, menu in ipairs(self.all) do
-		if menu.palette and not menu.search then
+		if not menu.search and (menu.palette or (menu.initial_query and itable_index_of(new_menus, menu))) then
 			update_dimensions_again = true
 			self:search_init(menu)
 		elseif not menu.palette and menu.search and menu.search.query == '' then
@@ -233,8 +233,8 @@ function Menu:update(data)
 		self:update_content_dimensions()
 		self:reset_navigation()
 	end
+	-- Execute initial search queries
 	for _, menu in ipairs(new_menus) do
-		print(menu.id..': '.. (menu.initial_query or ''))
 		if menu.initial_query then self:search_query_update(menu.initial_query, menu) end
 	end
 
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -273,26 +273,21 @@ function Menu:update_content_dimensions()
 end
 
 function Menu:update_dimensions()
-	-- Coordinates and sizes are of the scrollable area. Title is rendered
+	-- Coordinates and sizes are of the scrollable area to make
+	-- consuming values in rendering and collisions easier. Title is rendered
 	-- above it, so we need to account for that in max_height and ay position.
-	-- This is a debt from an era where we had different cursor event handling,
-	-- and dumb titles with no search inputs. It could use a refactor.
 	local min_width = state.fullormaxed and options.menu_min_width_fullscreen or options.menu_min_width
-	local vertical_padding = round(self.scroll_step / 2)
-	local height_available = display.height - vertical_padding * 2
+	local height_available = display.height - Elements.timeline.size_min
 
 	for _, menu in ipairs(self.all) do
 		local width = math.max(menu.search and menu.search.width or 0, menu.max_width)
 		menu.width = round(clamp(min_width, width, display.width * 0.9))
-		local title_height = (menu.title or menu.search) and self.scroll_step or 0
-		local max_height = height_available - title_height
+		local title_height = (menu.is_root and menu.title or menu.search) and self.scroll_step or 0
+		local max_height = round(height_available * 0.9 - title_height)
 		local content_height = self.scroll_step * #menu.items
 		menu.height = math.min(content_height - self.item_spacing, max_height)
 		local search_top = menu.search and menu.search.top or height_available
-		menu.top = math.max(
-			title_height + vertical_padding,
-			math.min(search_top, round((height_available - menu.height + title_height) / 2))
-		)
+		menu.top = math.min(search_top, round((height_available - menu.height + title_height) / 2))
 		menu.scroll_height = math.max(content_height - menu.height - self.item_spacing, 0)
 		menu.scroll_y = menu.scroll_y or 0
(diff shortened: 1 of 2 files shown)


scripts/uosc/elements/Menu.lua
```diff
@@ -255,21 +273,26 @@ function Menu:update_content_dimensions()
...
+		local title_height = (menu.title or menu.search) and self.scroll_step or 0
+		local max_height = height_available - title_height
 		local content_height = self.scroll_step * #menu.items
 		menu.height = math.min(content_height - self.item_spacing, max_height)
 		local search_top = menu.search and menu.search.top or height_available
-		menu.top = math.min(search_top, round((height_available - menu.height + title_height) / 2))
+		menu.top = math.max(
+			title_height + vertical_padding,
+			math.min(search_top, round((height_available - menu.height + title_height) / 2))
+		)
```
christoph-heinrich, line 295: In the root menu it now properly accounts for the search bar for small windows when it doesn't have a title.
image
image
But submenus are always moved down a bit because they do have titles and they require the `menu.is_root` check.
image

scripts/uosc/elements/Menu.lua
```diff
@@ -698,13 +722,20 @@ function Menu:search_query_update(query)
 	request_render()
 end
 
-function Menu:search_backspace()
-	local pos, search_text = #self.current.search.query - 1, self.current.search.query
-	while pos > 1 and search_text:byte(pos) >= 0x80 and search_text:byte(pos) <= 0xbf do
+---@param event? string
+function Menu:search_backspace(event)
+	local pos, old_query, is_palette = #self.current.search.query - 1, self.current.search.query, self.current.palette
+	while pos > 1 and old_query:byte(pos) >= 0x80 and old_query:byte(pos) <= 0xbf do
```
christoph-heinrich, line 728: I should have left a comment here that the while loop is for skipping utf8 continuation bytes.

christoph-heinrich (changes requested):

christoph-heinrich:
What's the point of that palette mode? Starting a search isn't that big of a deal for users.

scripts/uosc/elements/Menu.lua
christoph-heinrich: I forgot to mention that since this change is independent of the palette mode it should be a separate commit imo.

scripts/uosc/elements/Menu.lua
```diff
@@ -212,6 +213,23 @@ function Menu:update(data)
...
+	local update_dimensions_again = false
+	for _, menu in ipairs(self.all) do
+		if menu.palette and not menu.search then
+			update_dimensions_again = true
+			self:search_init(menu)
+		elseif not menu.palette and menu.search and menu.search.query == '' then
+			update_dimensions_again = true
+			menu.search = nil
+		end
+	end
```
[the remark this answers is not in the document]
tomasklaen: We update before _and_ after because search_inits need the initial un-searched menu's position and scroll state to save on the `search.source` table.

scripts/uosc/elements/Menu.lua
tomasklaen: True, I'll split it into a separate PR.

tomasklaen:
The point of palette mode is to make palette menus :) Like `ctrl+shift+p` in vscode.

Also for implementing stuff like subtitle downloader, where there's just an empty list initially with a search input at the top pre-populated with current file name.

I also have a `menu-palette` command in the pipe when this is merged. It'll display a flattened version of user defined menu to instantly search through, just like vscodes palette menu:

menu-palette

scripts/uosc/elements/Menu.lua
```diff
@@ -212,6 +215,29 @@ function Menu:update(data)
...
+	-- We update before _and_ after because search_inits need the initial un-searched
+	-- menu's position and scroll state to save on the `search.source` table.
+	if update_dimensions_again then
+		self:update_content_dimensions()
+		self:reset_navigation()
+	end
+	for _, menu in ipairs(new_menus) do
+		print(menu.id..': '.. (menu.initial_query or ''))
+		if menu.initial_query then self:search_query_update(menu.initial_query, menu) end
```
christoph-heinrich, line 238: I think that crashes if there isn't a currently active search (not a palette).

`search_query_update()` will cause a search to be executed, which is probably not what we want here?

And a `print()` slipped in.

christoph-heinrich (changes requested):

scripts/uosc/elements/Menu.lua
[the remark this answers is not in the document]
christoph-heinrich: I don't like it but I also can't think of something better :)

tomasklaen:
> search_query_update() will cause a search to be executed, which is probably not what we want here? I'm not even sure tbh.

I don't really know if I want it to be executed or not. I'm going with yes atm, but not sure about it. The use case is of course the subtitle downloader, where you make a menu with `on_search` callback, so setting `initial_query` will just prepopulate but not execute until `ctrl+enter` is pressed, effectively making it a search suggestion. It just feels useful to be able to execute initial queries for menu implementers.

I thought about making it just a `search_suggestion` (fills the query but doesn't execute search), and if we want scripts to execute searches from their end, implement something like `menu-search <type> <id> <query>` message. But what would be the behavior of search suggestion in instant internal search? It feels weird for it to not search and have an unfiltered menu that doesn't match query.

christoph-heinrich:
> It just feels useful to be able to populate initial queries for menu implementers.

Agree

> implement something like `menu-search <type> <id> <query>` message.

Isn't that the point of `on_search`? That's how the [search for memo](https://github.com/po5/memo/pull/19) is currently implemented. Scripts can put whatever identifying information they need in addition to the query into their specified callback, and then update the menu with the search results using `menu-update`.

> But what would be the behavior of search suggestion in instant internal search? It feels weird for it to not search and have an unfiltered menu that doesn't match query.

Scripts that want to provide a prefilled search field with the corresponding search result while providing `on_search` can simply set those items when opening the menu. They already know what the search query is, so no need for that callback.
But when the internal search is used we need to execute a search.

tomasklaen:
Renamed it to `search_suggestion`, but kept the behavior (executes internal search, but waits for `ctrl+enter` on external). It feels like it conveys what the purpose of it is better.

christoph-heinrich:
So you're fine with external searches calling `on_search` right after the menu was opened if they have a `search_suggestion` and `search_debounce ~= 'submit'` ?

tomasklaen:
No I'll fix it 😶.

Issue #587, opened 2023-07-18
5.0


tomasklaen:
A to-do list with things I want in 5.0 so I won't forget:

- [x] Sort out license
- [x] Update installation scripts (`uosc_shared` -> `uosc`)
- [x] Save installation scripts into `installers/uosc.{sh,ps1}` files and have only oneliners in readme that fetch and execute these.
- [x] ~~Menu item actions/buttons.~~
- [x] ~~Improved menu callback API with pressed modifiers, action, and other meta data.~~
- [x] Searchable menus to make command palettes.
- [x] Built in command palette with all keybinds from `input.conf`. Useful for checking what the shortcut for the thing I need right now is.

Stroked-through items will be left for later releases.

Zabooby:
Not sure if this is related but have you removed uosc.lua from the repo? I downloaded the code to get the latest changes but uosc.lua is missing.

tomasklaen:
Support for older versions of mpv was dropped, so uosc is now only inside the `scripts/uosc` folder. If you want to update to current master, delete `uosc_shared` and `uosc.lua`.

po5:
Some things that have been bugging me, around using the menu API's `on_close` for scripts that need to keep state of whether they're open or not to perform menu updates and close their own menu.

We can't react to an `on_close` callback immediately, only when our custom menu script frees the event loop again.  
It leads to two issues:
- We may call `update-menu` without yet knowing our menu was closed by the user, which causes it to reopen. Solution: ignore `update-menu` when there's no corresponding menu, but maybe that breaks someone else's use case? 
- In an attempt to close a menu, calling `open-menu` again without yet knowing our menu was closed by the user causes it to reopen. Solution: add a `close-menu` message to explicitly close a given menu, or do nothing if it doesn't exist.

These race conditions can be exacerbated by scripts that have to spend a lot of time in one event loop e.g. reading a large file, but affect any scripts that use menu updates or have to close their own menu.

PR #653, opened 2023-09-28
Menu script-message changes

christoph-heinrich:
Ref. https://github.com/tomasklaen/uosc/issues/587#issuecomment-1682294760

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -112,6 +112,7 @@ function Menu:init(data, callback, opts)
 	self.drag_last_y = nil
 	self.is_dragging = false
 
+	mp.set_property_native('user-data/uosc/menu/type', self.type)
 	self:update(data)
 
 	if self.mouse_nav then
@@ -131,13 +132,12 @@ function Menu:destroy()
 	self:disable_key_bindings()
 	self.is_closed = true
 	if not self.is_being_replaced then Elements.curtain:unregister('menu') end
+	mp.del_property('user-data/uosc/menu/type')
 	if self.opts.on_close then self.opts.on_close() end
 end
 
 ---@param data MenuData
 function Menu:update(data)
-	self.type = data.type
-
 	local new_root = {is_root = true, submenu_path = {}}
 	local new_all = {}
 	local new_by_id = {}
--- a/README.md
+++ b/README.md
@@ -483,7 +483,7 @@ mp.commandv('script-message-to', 'uosc', 'open-menu', json)
 
 ### `update-menu <menu_json>`
 
-Updates currently opened menu with the same `type`. If the menu isn't open, it will be opened.
+Updates currently opened menu with the same `type`.
 
 The difference between this and `open-menu` is that if the same type menu is already open, `open-menu` will close it (facilitating menu toggling with the same key/command), while `update-menu` will update it's data.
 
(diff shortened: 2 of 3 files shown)


tomasklaen:
That was the next thing on my to-do list :)

The only thing it's missing is keeping track of currently opened menu on the `user-data/uosc/menu/type` property, as using stuff like `update-menu` depends on it.

tomasklaen:
Also, all the menu toggling bindings (menu, subtitles, ...) need to be updated as they no longer toggle.

christoph-heinrich:
> The only thing it's missing is keeping track of currently opened menu on the `user-data/uosc/menu/type` property, as using stuff like `update-menu` depends on it.

Scripts that open their menu already know the type, no need to query a property to update their own menu.

> Also, all the menu toggling bindings (menu, subtitles, ...) need to be updated as they no longer toggle.

They do still toggle.
All builtin menus start with `if Menu:is_open(menu_type) then Menu:close() return end`.

tomasklaen:
Wait I don't understand now. You need to know if your menu is open to send either `open-menu` or `update-menu` message. So how can scripts know their menu is open if they can't get `user-data/uosc/menu/type`? The property check is not just to get the type, but to check if any menu is currently open, and if it's still the menu opened by the script.

I thought that the rework is supposed to implement @po5's requests that:

- `update-menu` should no longer open menus
- `open-menu` should no longer toggle

which I agree with. But then you need to be able to check what menu and of what type is currently open to re-implement these functionalities.

Also, `menu-close` should accept `type` as an optional parameter, and if it's empty, close any opened menu.

christoph-heinrich:
> So how can scripts know their menu is open if they can't get `user-data/uosc/menu/type`?

`on_close` can be used to know if the menu has been closed from the uosc side, so scripts can keep track if their menu is currently open or not.

tomasklaen:
> on_close can be used to know if the menu has been closed from the uosc side, so scripts can keep track of if their menu is currently open or not.

True, that is a viable alternative. But imo accessing a `user-data` prop is easier than setting up a `on_close` message listener and having a global flag that tracks it.

Regardless, I recall this being a requested feature (though I can't remember context anymore), it feels useful for scripts to be able to detect open menus, I think we should provide it, and it feels like it should be part of this PR. Since these are potentially breaking changes, I want to squash it into one `feat(api)!: menu script-message changes` commit with all of them listed in the description so that it's easier for people to inspect and adapt their code.

christoph-heinrich:
Added the new property.

scripts/uosc/elements/Menu.lua
```diff
@@ -131,12 +131,14 @@ function Menu:destroy()
...
 	self.is_closed = true
 	if not self.is_being_replaced then Elements.curtain:unregister('menu') end
+	mp.del_property('user-data/uosc/menu/type')
 	if self.opts.on_close then self.opts.on_close() end
 end
 
 ---@param data MenuData
 function Menu:update(data)
 	self.type = data.type
+	mp.set_property_native('user-data/uosc/menu/type', data.type)
```
tomasklaen, line 141: Is `set_property_native` going to trigger observers if the property is assigned the same value? If so, I don't think this should be in `:update()`. It's not like menu type can change. And if it can somehow, it's a bug.

christoph-heinrich: I think it triggers observers, didn't test. But even if it doesn't, `init()` is a better place for it.

PR #654, opened 2023-09-29
fix: menu positioning issues

tomasklaen:
Did I miss anything?

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -290,21 +290,26 @@ function Menu:update_content_dimensions()
 end
 
 function Menu:update_dimensions()
-	-- Coordinates and sizes are of the scrollable area to make
-	-- consuming values in rendering and collisions easier. Title is rendered
+	-- Coordinates and sizes are of the scrollable area. Title is rendered
 	-- above it, so we need to account for that in max_height and ay position.
+	-- This is a debt from an era where we had different cursor event handling,
+	-- and dumb titles with no search inputs. It could use a refactor.
 	local min_width = state.fullormaxed and options.menu_min_width_fullscreen or options.menu_min_width
-	local height_available = display.height - Elements.timeline.size_min
+	local vertical_padding = round(self.scroll_step / 2)
+	local height_available = display.height - vertical_padding * 2
 
 	for _, menu in ipairs(self.all) do
 		local width = math.max(menu.search and menu.search.width or 0, menu.max_width)
 		menu.width = round(clamp(min_width, width, display.width * 0.9))
 		local title_height = (menu.is_root and menu.title or menu.search) and self.scroll_step or 0
-		local max_height = round(height_available * 0.9 - title_height)
+		local max_height = height_available - title_height
 		local content_height = self.scroll_step * #menu.items
 		menu.height = math.min(content_height - self.item_spacing, max_height)
 		local search_top = menu.search and menu.search.top or height_available
-		menu.top = math.min(search_top, round((height_available - menu.height + title_height) / 2))
+		menu.top = math.max(
+			title_height + vertical_padding,
+			math.min(search_top, round((height_available - menu.height + title_height) / 2))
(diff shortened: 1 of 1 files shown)


christoph-heinrich:
`math.max(math.min())` looks like `clamp()` to me.

I have never noticed before that the title/search bar is less tall then menu items, and as a consequence the top of the menu shifts slightly down when starting a search in a menu without title and a window small enough that the menu is scrollable. But the menu items stay at the same height, so the calculation is correct.

With the vertical padding not being a fraction of the available space anymore, we could use the same padding horizontally too.

Also this is somewhat nitpicky, but it's really more of a margin then padding.

tomasklaen:
> I have never noticed before that the title/search bar is less tall then menu items

Hm... it shouldn't be, but I see it too. Probably forgetting to account for menu border, or item spacing somewhere.

> math.max(math.min()) looks like clamp() to me.

christoph-heinrich:
> But I was thinking if we shouldn't just align menus statically to the top with space for title/search always reserved. That way we don't have to care for any shifts, or keep track of the top property on `search` table.
> 
> But when item with a submenu is all the way on the bottom, and submenu is short, its placing will look even weirder than it does now 😢.

That would certainly make things easier, but I'd expect the context menu being at the top to look weird.

tomasklaen:
So I fixed the tiny shift, tweaked some stuff, and regarding the shifting of menu while searching, I've arrived at what I think is the best compromise here:

Since search starts, the menu's `top` position is only allowed to rise, never fall down, and goes back only after the search is canceled. This eliminates the top position jumping up and down as you search (issue in nested menus with `search_submenus` enabled), which was the main thing that bothered me.

christoph-heinrich:
Works well, but one thing I've noticed during testing is that the width changes while typing in a `search_submenus` search.
It becomes wider then the original menu, which makes sense with the submenu concatenation, but as the search narrows the items down it becomes less wide again.

We could either keep track of the max width it had during a search and not let it get below that until the search is over, or we get the max width of all the items at search start and then keep it at that.

tomasklaen:
I went with growing width as needed and not letting it shrink until search is canceled.

christoph-heinrich:
One more thing, sorry for not realizing this sooner, but `options.menu_min_width*` should define the minimum size of the content, but if the window gets too small it should still be allowed to get smaller then that.

As a side-effect of that someone could, if they wanted to, set the minimum width to be really big so that the menu always takes up the full width of the window.

christoph-heinrich:
This works.

```diff
diff --git a/scripts/uosc/elements/Menu.lua b/scripts/uosc/elements/Menu.lua
index 02e5a73..7bb8254 100644
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -295,12 +295,14 @@ function Menu:update_dimensions()
 	-- This is a debt from an era where we had different cursor event handling,
 	-- and dumb titles with no search inputs. It could use a refactor.
 	local margin = round(self.item_height / 2)
-	local min_width = state.fullormaxed and options.menu_min_width_fullscreen or options.menu_min_width
+	local width_available = display.width - margin * 2
 	local height_available = display.height - margin * 2
+	local min_width = state.fullormaxed and options.menu_min_width_fullscreen or options.menu_min_width
+	min_width = math.min(min_width, width_available)
 
 	for _, menu in ipairs(self.all) do
 		local width = math.max(menu.search and menu.search.max_width or 0, menu.max_width)
-		menu.width = round(clamp(min_width, width, display.width - margin * 2))
+		menu.width = round(clamp(min_width, width, width_available))
 		local title_height = (menu.is_root and menu.title or menu.search) and self.scroll_step or 0
 		local max_height = height_available - title_height
 		local content_height = self.scroll_step * #menu.items
```

christoph-heinrich:
This doesn't really matter because it's totally unusable at that size anyway, but since I noticed I might as well bring it up.
The clipping in the title could use some work.
image

PR #655, opened 2023-09-29
feat(api): added `search_recursive` prop to menus

tomasklaen:
`search_recursive` makes user search look into submenus as well, effectively flattening the menu for the duration of the search.

With this, we don't need a new `menu-palette` command anymore. It's just built into the default one:

search_recursive

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,7 +464,7 @@ When `Command.value` is a string, it'll be passed to `mp.command(value)`. If it'
 
 `search_debounce` controls how soon the search happens after the last character was entered in milliseconds. Entering new character resets the timer. Defaults to `300`. It can also have a special value `'submit'`, which triggers a search only after `ctrl+enter` was pressed.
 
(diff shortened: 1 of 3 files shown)


tomasklaen:
Maybe we should use `/` as a separator. `>` doesn't look good in non-monospace fonts, and takes too much space.

search_recursive_sep

christoph-heinrich:
`|` might be a good symbol for that, as it takes up even less space and people are less likely to have item titles with it.
But `/` is a commonly used path separator, so it's good from a semantic pov...

tomasklaen:
I went with `/`. It's pretty much a standard to delimit categories anywhere.

Also, should it be `search_recursive`, or `search_submenus`?

dyphire:
LGTM

christoph-heinrich:
`search_recursive` makes more sense because it doesn't only search submenus, although with the description that shouldn't really be a problem.

`search_all` maybe? But that's also not good because it doesn't search parent menus.

My personal favorite is `search_submenus`, even if `search_recursive` is technically more correct.

PR #661, opened 2023-10-02
feat: reworked timeline progress mode

### LINKED ISSUE #630: `toggle-progress` is not capable of forcefully showing the progress bar in fullscreen
It appears that `toggle-progress` is not capable of forcefully showing the progress bar in fullscreen when `timeline_size_min_fullscreen=0` (the default). Is this something that should be possible? How would the command know what size it should display the timeline at in fullscreen?

I'm starting to think that this cannot be achieved properly with just the `size_min` options and that a third option is necessary to communicate that the timeline should be visible in a window and hidden in fullscreen by default.

I can also make an issue out of this.

_Originally posted by @FichteFoll in https://github.com/tomasklaen/uosc/discussions/624#discussioncomment-7031394_

tomasklaen:
Config options:

```
timeline_line_width_fullscreen=3
timeline_line_width_minimized_scale=10
timeline_size_min=2
timeline_size_max=40
timeline_size_min_fullscreen=0
timeline_size_max_fullscreen=60
timeline_start_hidden=no
```

have been replaced with:

```
timeline_size=40
timeline_size_fullscreen=60
timeline_progress=windowed
timeline_progress_size=2
timeline_progress_line_width=20
```

This simplifies stuff a bunch, and enables timeline's progress mode to be togglable at all times. Previously you could only hide the progress, but not show it if it was configured to be hidden in current context.

closes #630

---

Can anyone test/review if I messed up or forgot something? :)

Known omission: `*_fullscreen` adjustments are missing on purpose, as next thing on my to-do list is to replace all `*_fullscreen` options with one `ui_scale_fullscreen`.


tomasklaen:
Hm..., I think it might be cleaner to drop `timeline_` prefix from `*progress*` options.

christoph-heinrich:
I didn't get a chance to look at this before it got merged, but I now updated to master and the minimized timeline line width is wrong (negative actually, it's like time goes back when the timeline minimizes).

tomasklaen:
Oh, I know what I did.

FichteFoll:
Thanks for these changes, works wonderfully.

PR #668, opened 2023-10-04
feat: clamp title and hint size to min 20% width

christoph-heinrich:
Doesn't yet have any of the non-linear stuff, but it's already a big improvement.

ref https://github.com/tomasklaen/uosc/pull/665#issuecomment-1745887138

For comparison with the screenshot in the comment:
image

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -1111,22 +1111,13 @@ function Menu:render()
 
 			local title_cut_x = content_bx
 			if item.hint_width > 0 then
-				local ratio = item.title_width / item.hint_width
-				if ratio <= 1 / 1.3 then
-					-- title is smaller
-					title_cut_x = math.ceil(content_ax + item.title_width)
-				elseif ratio >= 1.3 then
-					-- hint is smaller
-					title_cut_x = math.floor(content_bx - item.hint_width - spacing)
-				else
-					-- controls title & hint clipping proportional to the ratio of their widths
-					-- both title and hint get at least 20% of the width, unless they are smaller then that
-					local width = content_bx - content_ax
-					local title_min = math.min(item.title_width, width * 0.2)
-					local hint_min = math.min(item.hint_width, width * 0.2)
-					local title_ratio = item.title_width / (item.title_width + item.hint_width)
-					title_cut_x = round(content_ax + clamp(title_min, (width - spacing) * title_ratio, width - hint_min))
-				end
+				-- controls title & hint clipping proportional to the ratio of their widths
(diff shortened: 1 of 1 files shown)


christoph-heinrich:
I've tried doing a non linear interpolation together with that minimum size thing, and it adds a bunch of complexity to the calculation while also being susceptible to text with inaccuracies. It also has to account for the possibility that due to the non linearity one part actually gets much more space then it actually needs and so that then requires a clamp with the text width.

Overall it wasn't easy to get right, or rather I though it wasn't right due to the text width inaccuracy, and the benefit over a simple linear interpolation with clamping is questionable.

In case this ever becomes relevant in the future, I've took half of a 2 wide Hann function and transformed it from `y(x) =` to `x(y) =`, which results in `math.acos(1 - 2 * y) / math.pi`

christoph-heinrich:
Here it is, but I don't like it.
As the size of the menu changes, so does the decision for the clipping and as a result it jumps when resizing.
image
And after making it a few pixels smaller.
image

Notice how the gap jumped to the right, and also changed in size due to text width inaccuracies.

I also pushed an alternative that doesn't jump around and it generally works better imo.

However both have a major flaw, and that is that both title and hint can be very long, but differ in length enough to go into the "smaller one dictates cutoff" path, in which case it could be that there is nothing left of the larger one, and the smaller one even clips outside of the menu.

To fix that we'd have to clamp it to make sure there is at least something left of the longer one, in which case we effectively end up with the same with as I started off here, but the other way around (the smaller one gets more space then the larger one).

I still think the best thing is the one I started out with. We can increase the percentage if you think it's not enough, but the alternatives all suck in some way.

tomasklaen:
I really don't want long texts to cause clipping of really short ones. If text is long, it has a lot of noise and therefore is a prime candidate for clipping. If it's short, it's dense with information, and shouldn't be clipped.

So we need to find some ideal conditions to trigger this mode. Lets say the text can't be clipped, and all clipping happens on the other one if it's:

- smaller than half of the other one
- and smaller than `120` pixels

If neither satisfies this, we just use linear clipping we have now.

christoph-heinrich:
>    * smaller than half of the other one
>    * and smaller than `120` pixels

I'd rather not set a hard limit like 120 pixels, and instead make it relative to menu width or something.

I've now changed it to a 50% cap, so each one can only use more then that if the other one doesn't need it.
That will provide those 120 pixels or more for menus down to 240 pixels (actually a bit more for spacing and padding), which is still a relatively small menu.

tomasklaen:
> both are entitled to half

Yeah I really like this solution 👍.

PR #669, opened 2023-10-04
fix: progress line width calculation

christoph-heinrich:
Turns out it was merely a `-` instead of a `+`

Ref. https://github.com/tomasklaen/uosc/pull/661#issuecomment-1746096396


tomasklaen:
Oh, I pushed a fix before seeing this :)

PR #675, opened 2023-10-06
fix: menu icon fade out

christoph-heinrich:
It also used `text_opacity` instead of `menu_opacity` like the rest of the title/search bar.

But maybe everything in that bar should use `text_opacity` anyway?

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -1173,12 +1173,12 @@ function Menu:render()
 			-- Title
 			if menu.search then
 				-- Icon
-				local icon_size, icon_opacity = self.font_size * 1.3, requires_submit and text_opacity * 0.5 or 1
+				local icon_size, icon_opacity = self.font_size * 1.3, menu_opacity * (requires_submit and 0.5 or 1)
 				local icon_rect = {ax = rect.ax, ay = rect.ay, bx = ax + icon_size + spacing * 1.5, by = rect.by}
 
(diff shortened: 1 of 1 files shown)


tomasklaen:
> But maybe everything in that bar should use text_opacity anyway?

Yes, icons are essentially text.

christoph-heinrich:
> > But maybe everything in that bar should use text_opacity anyway?
> 
> Yes, icons are essentially text.

But the other text in the title/search bar doesn't use text_opacity, hence my confusion.

christoph-heinrich:
I've added a commit to fix the menu_opacity and text_opacity usage.
One of those is for the highlight, not sure if we want to change that one.

tomasklaen:
> One of those is for the highlight, not sure if we want to change that one.

Since the addition of `options.menu_opacity` it's all a bit misnamed so it's hard to follow. It needs a refactor.

`options.menu_opacity` should only affect the menu background. So, we should swap current `menu_opacity` and `text_opacity` for:

- `menu_opacity` - It should start at 1 for current menu, decay for (parent/sub)menus, and multiplied by `self.opacity`. This should be the base used by everything.
- `background_opacity = menu_opacity * options.menu_opacity` - and only used by the background, so I guess we don't need it as a value.

tomasklaen:
Don't bother with this, I'm working on the opacity options overhaul and this is in the path, so I'll include it there.

christoph-heinrich:
Alright I'll close it then.

PR #677, opened 2023-10-06
fix: updating the menu while using the internal search

### LINKED ISSUE #671: Internal menu search with updating items
Updating the menu while also using the internal search causes the items to not be filtered anymore.

At least that's the most obvious problem here, but the underlying problem goes much deeper.
Updating the menu results in new items being created and `item.search` being assigned over from the old items, which contains a list of the original item objects in `item.search.source.items`.

That already caused problems during the development of the search feature and the workaround was a new update function for the search itself, but as it turns out that wasn't enough.

We have to make the normal update function able to deal with this somehow.
My first guess is replacing the items in `item.search.source.items` with the new ones and performing a search, but I didn't see an easy way of doing that when I had a quick look. Maybe I'll give it another try later.

I noticed it while working on [my new script](https://github.com/christoph-heinrich/mpv-subtitle-lines), but it turned out to also be a problem with builtin menus (e.g. chapter menu).

### Reproduction steps

1. open a file with chapters
2. open the chapter menu during playback
3. search for something
4. wait until playback progresses to the next chapter

christoph-heinrich:
I hope I didn't miss anything.

Fixes #671

<hr/>

While testing this I noticed that the context menu uses recursive search, but only on the root menu, submenus perform a regular search. Is that intentional?
It's also missing from the menu data structure in the readme, so it's not clear if it's supposed to be a per menu or a global option.

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -237,13 +237,14 @@ function Menu:update(data)
 	end
 	-- Apply search suggestions
 	for _, menu in ipairs(new_menus) do
-		if menu.search_suggestion then
+		if menu.search_suggestion then menu.search.query = menu.search_suggestion end
+	end
+	for _, menu in ipairs(self.all) do
+		if menu.search then
+			-- the menu items are new objects and the search needs to contain those
+			menu.search.source.items = not menu.on_search and menu.items or nil
 			-- Only internal searches are immediately submitted
-			if menu.on_search then
-				menu.search.query = menu.search_suggestion
-			else
-				self:search_query_update(menu.search_suggestion, menu)
-			end
+			if not menu.on_search then self:search_submit(menu) end
 		end
 	end
 
@@ -687,7 +688,7 @@ function Menu:search_internal(menu)
 		menu.items = search_items(menu.search.source.items, query, menu.search_submenus)
 		-- Select 1st item in search results
 		menu.scroll_y = 0
-		self:select_index(1, menu)
+		if not self.mouse_nav then self:select_index(1, menu) end
 	end
 	self:update_content_dimensions()
 end


scripts/uosc/elements/Menu.lua
```diff
@@ -237,13 +237,14 @@ function Menu:update(data)
 	end
 	-- Apply search suggestions
 	for _, menu in ipairs(new_menus) do
-		if menu.search_suggestion then
+		if menu.search_suggestion then menu.search.query = menu.search_suggestion end
```
tomasklaen, line 240: This is a new menu that didn't have search initialized yet, so `menu.search` will be `nil` here, no?

tomasklaen:
> While testing this I noticed that the context menu uses recursive search, but only on the root menu, submenus perform a regular search. Is that intentional?
It's also missing from the menu data structure in the readme, so it's not clear if it's supposed to be a per menu or a global option.

Per menu, but parent menus should also enable it for submenus. I'll fix it.

scripts/uosc/elements/Menu.lua
christoph-heinrich: Doesn't the loop above that take care of it? Otherwise the same problem exists on the main branch.

tomasklaen: True about the above loop 😐.

scripts/uosc/elements/Menu.lua
```diff
@@ -237,13 +237,14 @@ function Menu:update(data)
...
+	end
+	for _, menu in ipairs(self.all) do
+		if menu.search then
+			-- the menu items are new objects and the search needs to contain those
+			menu.search.source.items = not menu.on_search and menu.items or nil
 			-- Only internal searches are immediately submitted
-			if menu.on_search then
-				menu.search.query = menu.search_suggestion
```
tomasklaen, line 243: By removing this, menus with `menu.on_search` will not get their input pre-filled with `menu.search_suggestion` anymore, no?

christoph-heinrich: line 240

PR #680, opened 2023-10-07
feat!: reworked opacity options

### LINKED ISSUE #584: Changing opacity of timeline progress
Adding this here as a to-do item.

### Discussed in https://github.com/tomasklaen/uosc/discussions/583

<div type='discussions-op-text'>

<sup>Originally posted by **croak9854** July  9, 2023</sup>
Can we change the opacity of just timeline_line just like timeline_chapters_opacity? 

It would be easier to see the progress if we can decrease the opacity of the timeline but not the timeline_line.</div>

Should be solved with an optional opacity param to `timeline_opacity` option. Example:

```
timeline_opacity=<full>
timeline_opacity=<background>:<progress>
```

### LINKED ISSUE #636: Add option to hide background for tooltips
If some users have the background off for the title, tooltips look kinda out of place with backgrounds. As people commented in #237 we really love unreadable text. 🙃

tomasklaen:
These options have been removed:

```
timeline_opacity
timeline_chapters_opacity
volume_opacity
speed_opacity
menu_opacity
menu_parent_opacity
top_bar_title_opacity
window_border_opacity
curtain_opacity
```

and replaced with one comma delimited `opacity` list to override all of the above and some new default opacity values.

Example:

```
opacity=timeline=0.5,title=0.5
```

List of opacity values available for override and their current defaults:

```
timeline=.9
position=1     # timeline progress or line
chapters=0.8
slider=0.9     # background of all sliders, such as volume
slider_gauge=1 # value bar of all sliders
speed=0.6
menu=1
submenu=0.4
border=1       # window border
title=1        # window title
tooltip=1      # tooltip background
thumbnail=1    # thumbnail border
curtain=0.5
```

closes #584, closes #636, closes #675

### DIFF
--- a/script-opts/uosc.conf
+++ b/script-opts/uosc.conf
@@ -7,14 +7,10 @@ timeline_size=40
 # Comma separated states when element should always be fully visible.
 # Available: paused, audio, image, video, idle, windowed, fullscreen
 timeline_persistency=paused
-# Timeline opacity
-timeline_opacity=0.9
 # Top border of background color to help visually separate timeline from video
 timeline_border=1
 # When scrolling above timeline, wheel will seek by this amount of seconds
 timeline_step=5
-# Opacity of chapter indicators in timeline, 0 to disable
-timeline_chapters_opacity=0.8
 # Render cache indicators for streaming content
 timeline_cache=yes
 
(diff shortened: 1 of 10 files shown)


tomasklaen:
I wonder if `speed` should just use `slider` value.

christoph-heinrich:
The speed control is aesthetically (and functionally) very different from the volume slider, so it makes sense to be separate.

PR #684, opened 2023-10-08
feat: update german translation

christoph-heinrich:
We really need some way of notifying people that a new translation is required, I almost missed that.

Maybe some sort of template translation file that gets updated in it's own commit to make it obvious that it changed.


tomasklaen:
Yes, I was thinking about having a `build/translation_template.lua` script that would extract all strings matching `%Wt('([%']+)'` and make a `script/intl/_template.json` out of them with empty values.

But since I don't care about i18n that much, it sits at the bottom of my priorities :)

christoph-heinrich:
I've used `rg --no-filename --only-matching -r '$1' "\st\('([^']*)'" | sort | uniq` to get a list.

PR #691, opened 2023-10-10
feat: scripts to install or update uosc with a single line

tomasklaen:
I haven't tested the script for Linux & macOS. If anyone could, it'd be nice :) It should install current latest stable.

I also want to figure out if we can provide a `uosc/update` command that people can just run from their menu and update uosc in the background.

### DIFF
--- a/installers/unix.sh
+++ b/installers/unix.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+zip_url=https://github.com/tomasklaen/uosc/releases/latest/download/uosc.zip
+conf_url=https://github.com/tomasklaen/uosc/releases/latest/download/uosc.conf
+data_dir="${XDG_CONFIG_HOME:-~/.config}/mpv"
+zip_file=/tmp/uosc.zip
+
+# Exit immediately if a command exits with a non-zero status
+set -e
+
+cleanup() {
+	echo "Deleting: $zip_file"
+	rm -f $zip_file
+}
+
+abort() {
+	cleanup
+	printf "%s\n" "$@" >&2
+	exit 1
+}
+
+# Check OS
+OS="$(uname)"
+if [ "${OS}" == "Linux" ]; then
+	data_dir="${XDG_CONFIG_HOME:-~/.config}/mpv"
+elif [ "${OS}" == "Darwin" ]; then
+	data_dir=~/Library/Preferences/mpv
+else
+	abort "This install script works only on linux and macOS."
+fi
+
+# Remove old and deprecated folders & files
+echo "Deleting old and deprecated uosc files and directories."
+rm -rf "$data_dir/scripts/uosc_shared"
+rm -rf "$data_dir/scripts/uosc"
+rm -f "$data_dir/scripts/uosc.lua"
+
+# Install new version
+echo "Downloading: $zip_url"
+curl -o $zip_file $zip_url
+echo "Extracting: $zip_file"
+unzip -od $data_dir $zip_file
+cleanup
+
+# Download default config if one doesn't exist yet
+scriptopts_dir="$data_dir/script-opts"
+conf_file="$scriptopts_dir/uosc.conf"
+if [ ! -f "$conf_file" ]; then
+	echo "Config not found."
+	mkdir -pv $scriptopts_dir
+	echo "Downloading: $conf_url"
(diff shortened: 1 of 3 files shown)


christoph-heinrich:
Github redirects the download url and curl can't deal with that without adding `-L`, and `~` needs to be expanded manually.
```diff
diff --git a/installers/unix.sh b/installers/unix.sh
index 8d30e88..97872a0 100644
--- a/installers/unix.sh
+++ b/installers/unix.sh
@@ -29,6 +29,8 @@ else
 	abort "This install script works only on linux and macOS."
 fi
 
+data_dir="${data_dir/#\~/$HOME}"
+
 # Remove old and deprecated folders & files
 echo "Deleting old and deprecated uosc files and directories."
 rm -rf "$data_dir/scripts/uosc_shared"
@@ -37,7 +39,7 @@ rm -f "$data_dir/scripts/uosc.lua"
 
 # Install new version
 echo "Downloading: $zip_url"
-curl -o $zip_file $zip_url
+curl -L -o $zip_file $zip_url
 echo "Extracting: $zip_file"
 unzip -od $data_dir $zip_file
 cleanup
@@ -49,7 +51,7 @@ if [ ! -f "$conf_file" ]; then
 	echo "Config not found."
 	mkdir -pv $scriptopts_dir
 	echo "Downloading: $conf_url"
-	curl -o $conf_file $conf_url
+	curl -L -o $conf_file $conf_url
 fi
 
 echo "uosc has been installed."
```

installers/unix.sh
```diff
@@ -0,0 +1,55 @@
...
+abort() {
+	cleanup
+	printf "%s\n" "$@" >&2
+	exit 1
+}
+
+# Check OS
+OS="$(uname)"
+if [ "${OS}" == "Linux" ]; then
+	data_dir="${XDG_CONFIG_HOME:-~/.config}/mpv"
```
christoph-heinrich, line 25:
```suggestion
	data_dir="${XDG_CONFIG_HOME:-$HOME/.config}/mpv"
```
`~` doesn't get automatically expanded

installers/unix.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Remove old and deprecated folders & files
+echo "Deleting old and deprecated uosc files and directories."
+rm -rf "$data_dir/scripts/uosc_shared"
+rm -rf "$data_dir/scripts/uosc"
+rm -f "$data_dir/scripts/uosc.lua"
+
+# Install new version
+echo "Downloading: $zip_url"
+curl -o $zip_file $zip_url
```
christoph-heinrich, line 40:
```suggestion
curl -L -o $zip_file $zip_url
```
curl doesn't deal with the redirect unless `-L` is passed

installers/unix.sh
```diff
@@ -0,0 +1,55 @@
...
+cleanup
+
+# Download default config if one doesn't exist yet
+scriptopts_dir="$data_dir/script-opts"
+conf_file="$scriptopts_dir/uosc.conf"
+if [ ! -f "$conf_file" ]; then
+	echo "Config not found."
+	mkdir -pv $scriptopts_dir
+	echo "Downloading: $conf_url"
+	curl -o $conf_file $conf_url
```
christoph-heinrich, line 52: same as above

christoph-heinrich (changes requested):
Can't test on mac, but this is required for linux.

tomasklaen:
Is there no way in mpv to spawn an async subprocess that won't freeze the UI and listen to its stdout and exit events with callbacks?

tomasklaen:
I completely missed that 😐.

christoph-heinrich:
I think it's possible to start a process directly from lua and then treat it's stdin and stdout like a file, but I never tried that myself.

tomasklaen:
> I think it's possible to start a process directly from lua and then treat it's stdin and stdout like a file, but I never tried that myself.

What do you mean by this? Make the process output into a file with `command > file.txt` and read the file periodically?

christoph-heinrich:
> What do you mean by this?

https://www.lua.org/manual/5.1/manual.html#pdf-io.popen

tomasklaen:
Dunno why I always assume we don't have access to stuff like this in here. But I can't find any docs, or snippets on how to use it asynchronously without blocking the thread.

Issue #686, opened 2023-10-09
Improve music experience


33KK:
- Show something instead of a black screen like the logo mpv osc shows in idle state, or some custom image
- Show metadata like https://github.com/vc-01/metadata-osd

tomasklaen:
Audio indicator was quick and easy, but displaying meta will be a bit more involved and require new options, so I'll leave it for some time later.

christoph-heinrich:
I don't see why this has to be part of uosc, one can simply use a different script in combination with uosc.

christoph-heinrich:
Great...
![image](https://github.com/tomasklaen/uosc/assets/8932183/55d3a91f-a0be-412a-affb-559694c8ece4)

dyphire:
I also don't understand what is preventing you using other scripts in combination with uosc.

NaiveInvestigator:
same i also personally use [vc-01/metadata-osd](https://github.com/vc-01/metadata-osd) alongside with uosc for audio.

tomasklaen:
I do think that it is valuable and a uosc's concern to indicate to the user there is no video. It's a difference between user thinking "this part of the video is black", and "there is no video".

But it of course _should_ be replaceable with an equalizer script or something similar. That'll be possible with `disable_elements=audio_indicator` option once I get around to adding it.

And I see that there are good scripts for displaying meta, so I won't be adding _that_ in.

mrfragger:
I only listen to audiobooks with transcribed subtitles and always enable metadata for all audio but not for video.  Most important for me was to always show the current chapter playing which it does.  Here's a screenshot of what it looks like playing a song.  Only downside is if you try to use dual subs the one on top it blocks.  When mpv 0.38 comes out there'll be a --secondary-sub-pos to deal with this however.

in script-opts/usoc.conf

```
top_bar=always
top_bar_size=40
top_bar_controls=no
top_bar_title=yes
top_bar_alt_title=${time-remaining} (${percent-pos}%) ${filename/no-ext}
top_bar_alt_title_place=below
top_bar_flash_on=video,audio
top_bar_persistency=audio
```
```
<img width="1437" alt="Screenshot 2024-01-13 at 4 52 39 PM" src="https://github.com/tomasklaen/uosc/assets/80211672/5e07f23d-d5ab-4172-8274-c4386db115c8">
```

zaeboba:
Could you please share the settings? Very interesting icons on the panel and what they do

[...]

PR #695, opened 2023-10-12
feat: added `disable_elements` option and `disable-elements` script message

### LINKED ISSUE #592: Using uosc with other UIs
Here's a case for #467.

It's perfectly possible to use uosc alongside other UI scripts, by disabling conflicting elements.  
Below is my config for disabling everything:
```
timeline_size_max=0
timeline_size_max_fullscreen=0
controls=never
top_bar=never
window_border_size=0
pause_indicator=none
volume=none
```
The ability to mix & match uosc with other UIs is great, it allows for example to add a title bar and volume controls to [mfpbar](https://codeberg.org/NRK/mpv-toolbox/src/branch/master/mfpbar).

Disabling all uosc on-screen actions also lets users benefit from uosc's menu system as requested in #94.  
This is useful for user-defined menus and scripts that are enhanced by uosc:  
https://github.com/christoph-heinrich/mpv-quality-menu  
https://github.com/po5/memo  
or scripts that even require it:  
https://github.com/natural-harmonia-gropius/recent-menu

I think the buffering indicator is the only uosc element that isn't explicitly brought up by the user, that can't be disabled.

tomasklaen:
Allows disabling elements and various indicators by adding their IDs to the list:

```conf
disable_elements=timeline,audio_indicator
```

Also includes a new script message listener `disable-elements`, that allows scripts to do the same thing:

```lua
local id = mp.get_script_name()
mp.commandv('script-message-to', 'uosc', 'disable-elements', id, 'timeline,audio_indicator')
```

It'll register what elements each script wants disabled. The element will be enabled only when it was not disabled by neither user nor any script.

To cancel or re-enable the elements, pass an empty list:

```lua
mp.commandv('script-message-to', 'uosc', 'disable-elements', id, '')
```

ref #686, closes #592

---

This was way more work then I expected :(

I hope I didn't introduce many bugs.

Also, is it really not possible to get the name of a script that sent the message? We wouldn't need the `script_id` parameter on the message then.

### DIFF
--- a/scripts/uosc/elements/Element.lua
+++ b/scripts/uosc/elements/Element.lua
@@ -1,4 +1,4 @@
----@alias ElementProps {enabled?: boolean; ax?: number; ay?: number; bx?: number; by?: number; ignores_menu?: boolean; anchor_id?: string;}
+---@alias ElementProps {enabled?: boolean; render_order?: number; ax?: number; ay?: number; bx?: number; by?: number; ignores_menu?: boolean; anchor_id?: string;}
 
 -- Base class all elements inherit from.
 ---@class Element : Class
@@ -8,6 +8,7 @@ local Element = class()
 ---@param props? ElementProps
 function Element:init(id, props)
 	self.id = id
+	self.render_order = 1
 	-- `false` means element won't be rendered, or receive events
 	self.enabled = true
 	-- Element coordinates
@@ -24,6 +25,8 @@ function Element:init(id, props)
 	self.ignores_menu = false
 	---@type nil|string ID of an element from which this one should inherit visibility.
 	self.anchor_id = nil
+	---@type fun()[] Disposer functions called when element is destroyed.
+	self._disposers = {}
 
 	if props then table_assign(self, props) end
 
@@ -40,6 +43,7 @@ function Element:init(id, props)
 end
 
 function Element:destroy()
+	for _, disposer in ipairs(self._disposers) do disposer() end
 	self.destroyed = true
 	Elements:remove(self)
 end
@@ -70,7 +74,10 @@ function Element:is_persistent()
 	local persist = config[self.id .. '_persistency']
 	return persist and (
 		(persist.audio and state.is_audio)
-		or (persist.paused and state.pause and (not Elements.timeline.pressed or Elements.timeline.pressed.pause))
+		or (
+			persist.paused and state.pause
+			and (not Elements.timeline or not Elements.timeline.pressed or Elements.timeline.pressed.pause)
+		)
 		or (persist.video and state.is_video)
 		or (persist.image and state.is_image)
 		or (persist.idle and state.is_idle)
@@ -152,6 +159,30 @@ function Element:flash()
 		self._flash_out_timer:kill()
 		self._flash_out_timer:resume()
 	end
+end
+
+-- Register disposer to be called when element is destroyed.
+---@param disposer fun()
+function Element:register_disposer(disposer)
+	if not itable_index_of(self._disposers, disposer) then
+		self._disposers[#self._disposers + 1] = disposer
+	end
+end
+
+-- Automatically registers disposer for the passed callback.
+---@param event string
+---@param callback fun()
+function Element:register_mp_event(event, callback)
+	mp.register_event(event, callback)
+	self:register_disposer(function() mp.unregister_event(callback) end)
+end
+
+-- Automatically registers disposer for the observer.
+---@param name string
+---@param callback fun(name: string, value: any)
+function Element:observe_mp_property(name, callback)
+	mp.observe_property(name, 'native', callback)
+	self:register_disposer(function() mp.unobserve_property(callback) end)
 end
 
 return Element
--- a/scripts/uosc/elements/PauseIndicator.lua
+++ b/scripts/uosc/elements/PauseIndicator.lua
@@ -5,7 +5,7 @@ local PauseIndicator = class(Element)
 
 function PauseIndicator:new() return Class.new(self) --[[@as PauseIndicator]] end
 function PauseIndicator:init()
-	Element.init(self, 'pause_indicator')
+	Element.init(self, 'pause_indicator', {render_order = 3})
 	self.ignores_menu = true
 	self.paused = state.pause
 	self.fadeout_requested = false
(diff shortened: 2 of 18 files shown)


christoph-heinrich:
Is it really necessary do destroy all disabled elements? I would have expected it to have some additional management around `element.enabled` and that's it.

Sets in lua can be made with `set[value] = true`, I don't think checking the final table if a value is already in there is faster or simpler (maybe for small lists? I didn't benchmark it).

tomasklaen:
> Is it really necessary do destroy all disabled elements? I would have expected it to have some additional management around element.enabled and that's it.

`element.enabled` is intended for temporary toggling during use (like timeline hiding when obstructed, or no duration/time available). `disable_elements` is for "I never want to see this element again". In practice `element.enabled` won't cleanup everything, like bound mp event listeners, or property observers, but destroying an element will now.

> Sets in lua can be made with set[value] = true, I don't think checking the final table if a value is already in there is faster or simpler (maybe for small lists? I didn't benchmark it).

That `Manager.disabled` table is to be read by outside stuff that is not a full element, like audio indicator (and idle indicator I'll add when this is merged), which is just a condition in `render()`.

But I'll simplify it a bit with a set.

christoph-heinrich:
There is no need for `itable_join_unique()`...

christoph-heinrich:
image

christoph-heinrich:
I'll have a closer look at the code tomorrow, but other then those two crashes it has worked well so far.

scripts/uosc/lib/std.lua
```diff
@@ -146,13 +156,13 @@ function itable_slice(itable, start_pos, end_pos)
...
----@param b T[]|nil
+---@param ...T[]|nil
 ---@return T[]
-function itable_join(a, b)
+function itable_join(...)
 	local result = {}
-	if a then for _, value in ipairs(a) do result[#result + 1] = value end end
-	if b then for _, value in ipairs(b) do result[#result + 1] = value end end
+	for _, table in ipairs({...}) do
+		if table then for _, value in ipairs(table) do result[#result + 1] = value end end
```
christoph-heinrich, line 164: Is it actually necessary to check table for nil?

scripts/uosc/elements/PauseIndicator.lua
```diff
@@ -5,7 +5,7 @@ local PauseIndicator = class(Element)
 
 function PauseIndicator:new() return Class.new(self) --[[@as PauseIndicator]] end
 function PauseIndicator:init()
```
christoph-heinrich, line 7: I didn't verify this in a test, but I think this needs to call `decide()`, because otherwise the static pause indicator won't show if a new pause indicator gets created because of a script-message.

Option changes cause `on_prop_pause()` to be called, so it's not a problem there.

scripts/uosc/elements/Element.lua
```diff
@@ -154,4 +161,28 @@ function Element:flash()
 	end
 end
 
+-- Register disposer to be called when element is destroyed.
+---@param disposer fun()
+function Element:register_disposer(disposer)
+	if not itable_index_of(self._disposers, disposer) then
```
christoph-heinrich, line 167: The disposer is always a new function, so `itable_index_of()` will never find anything.

christoph-heinrich (commented):
Instead of having a hard coded number for `render_order` in each element, we could have a static list with element ids and then use the index as the `render_order`. That would make keeping track of what gets rendered in which order easier.

scripts/uosc/elements/Volume.lua
```diff
@@ -222,33 +200,40 @@ local Volume = class(Element)
...
 function Volume:update_dimensions()
-	local width = round(options.volume_size * state.scale)
-	local controls, timeline, top_bar = Elements.controls, Elements.timeline, Elements.top_bar
-	local min_y = top_bar.enabled and top_bar.by or 0
-	local max_y = (controls and controls.enabled and controls.ay) or (timeline.enabled and timeline.ay)
-		or display.height - top_bar.size
+	self.size = round(options.volume_size * state.scale)
+	local min_y = Elements:ev('top_bar', 'by', 0)
+	local max_y = Elements:ev('controls', 'ay') or Elements:ev('timeline', 'ay')
+		or display.height - Elements:ev('top_bar', 'size', 0)
```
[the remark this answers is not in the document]
tomasklaen: I guess that was supposed to be `window_border`.

scripts/uosc/lib/std.lua
tomasklaen: As the function's signature suggests, I want it to accept stuff like `itable_join(array, nil, array, ...)` and just ignore nils, but after testing it, the iterator thinks `nil` is an end of the array and stops there. I'll fix it.

scripts/uosc/elements/PauseIndicator.lua
tomasklaen: I'll add it.

scripts/uosc/elements/Element.lua
tomasklaen: It is a new function in current implementations, but it doesn't have to be in future ones. And it's not a good idea to call disposers twice.

tomasklaen:
> Instead of having a hard coded number for render_order in each element, we could have a static list with element ids and then use the index as the render_order. That would make keeping track of what gets rendered in which order easier.

I don't like the idea of it being disjointed from the element's implementation on a new state config. It would make sense if elements were actually completely self contained entities oblivious about the outside world, but since they know about each other and use each other's states, I think it's fine for them to configure themselves like that in their `:init()`s, just as they are configuring their own IDs in the `Elements` collection. At least in the current code base. And it's simple :)

Damn I wish I had time to rewrite it all. I don't like how the current element system and rendering works at all.

christoph-heinrich:
It really wouldn't that complicated. e.g.
```lua
render_order_ids = {
	'window_border',
	'buffer_indicator',
	...
}

function Element:init(id, props)
	self.id = id
	self.render_order = itable_index_of(render_order_ids, id)
	...
```
But if you'd rather have `{render_order = x}` in each element, then that's fine too I guess.

po5:
> Also, is it really not possible to get the name of a script that sent the message? We wouldn't need the script_id parameter on the message then.

I looked into how events are propagated internally a while ago and I don't believe there is a way to get that info.

tomasklaen:
> But if you'd rather have {render_order = x} in each element, then that's fine too I guess.

We need to have the init passing of render_order anyway, since bunch of elements need to inherit their parent order (control buttons/speed, volume slider).

christoph-heinrich:
> We need to have the init passing of render_order anyway, since bunch of elements need to inherit their parent order (control buttons/speed, volume slider).

That's only necessary in those few instances and those two things are not mutually exclusive.

PR #699, opened 2023-10-14
feat: optimize Volume icons

dyphire:


tomasklaen:
I already tried this once as I like the idea, but in practice, the way the icons are positioned in the font, the speaker part of the icon keeps shifting horizontally as the icon changes.

dynamic_volume

And for me personally, the shifting annoyance outweighs the eye candy from dynamic icon.

But if others don't mind, I'd merge this.

natural-harmonia-gropius:
> the shifting annoyance outweighs the eye candy from dynamic icon.

`ass:icon(self.ax + (width / 2), self.by, width * 0.7, icon_name,`

Change (width / 2) to a fixed value?

dyphire:
Yes, the icons themselves are not aligned. I personally don't mind much about this, but let's wait for the opinions of others.

christoph-heinrich:
It should be possible to prevent the shifting by using alignment 4 instead of 5, but then the icon won't always be perfectly centered. I don't think the shifting would bother me all that much.

osc.lua also has a symbol for >100% volume
image
But that doesn't seem to be an official symbol. mpv ships it's own osd-font, and we happen to also ship font files, so someone could add a symbol for it... or we figure out how to put a text `!` behind the icon and make it look fine, but simply putting a ! behind the icon name does not look fine.
image

tomasklaen:
> It should be possible to prevent the shifting by using alignment 4 instead of 5

I've tried that already and it didn't work. It seems like the icons are monospaced and align to the middle of their monospace width.

The only way to make it align would probably be to manually shift each icon by the necessary width fraction.

tomasklaen:
So I tweaked it:

volume_unshift

christoph-heinrich:
Interesting that the offset required is exactly width/16 for each stage.
I'm honestly not sure if I prefer the stationary or always centered one.

natural-harmonia-gropius:
> I'm honestly not sure if I prefer the stationary or always centered one.

Add an opacity "full volume" as background? 

image

PR #700, opened 2023-10-14
feat: added `update` command to update uosc

tomasklaen:
https://github.com/tomasklaen/uosc/assets/47283320/93ce48e3-02d3-4df9-85cc-3af8df7c3b1b

If anyone can test it on linux/mac, it'd be great, but be aware that it's going to replace your current uosc in `~/.config/mpv` with the latest release, which is 4.7.

The only issue this has now (that I know of) is that to do a windows portable update, I need to know the path to mpv binary so the updating process can run in its working directory.

Is there a way to get it somehow? And is there some arg to `mp.command_native_async()` to set the working directory? Can't find info on any of these. But I haven't tried _super_ hard, so hopefully it's still doable :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,19 +292,19 @@ Open directory with `mpv.conf` in file explorer.
 
 #### `update`
 
-Updates uosc to the latest stable release right from the UI. Available in default menu in the "Utils" section.
+Updates uosc to the latest stable release right from the UI. Available in the "Utils" section of default menu .
 
 Supported environments:
 
 | Env | Works | Note |
 |:---|:---:|---|
-| Windows | ✔️ | _Not tested on older PowerShell versions. And you might need to `Set-ExecutionPolicy` from the install instructions and install with the terminal command first._ |
+| Windows | ✔️ | _Not tested on older PowerShell versions. You might need to `Set-ExecutionPolicy` from the install instructions and install with the terminal command first._ |
 | Linux (apt) | ✔️ | |
 | Linux (flatpak) | ✔️ | |
-| Linux (snap) | ❌ | We're not allowed to access commands like `curl` even if they're installed. (Or at least I think this is the issue.) |
+| Linux (snap) | ❌ | We're not allowed to access commands like `curl` even if they're installed. (Or at least this is what I think the issue is.) |
 | MacOS | ❌ | `(23) Failed writing body` error, whatever that means. |
 
-If you know about a solution to fix self-updater in one of the currently broken environments, please make an issue/PR and share it with us!
+If you know about a solution to fix self-updater for any of the currently broken environments, please make an issue/PR and share it with us!
 
 **Note:** The terminal commands from install instructions still work fine everywhere, so you can use those to update instead.
 
--- a/README.md
+++ b/README.md
@@ -290,6 +290,10 @@ Switch audio output device.
 
 Open directory with `mpv.conf` in file explorer.
 
+#### `update`
+
+Updates uosc to the latest stable release.
+
 ## Menu
 
 **uosc** provides a way to build, display, and use your own menu. By default it displays a pre-configured menu with common actions.
@@ -401,6 +405,7 @@ ctrl+s      async screenshot                       #! Utils > Screenshot
 alt+i       script-binding uosc/inputs             #! Utils > Inputs
 O           script-binding uosc/show-in-directory  #! Utils > Show in directory
 #           script-binding uosc/open-config-directory #! Utils > Open config directory
+#           script-binding uosc/update             #! Utils > Update uosc
 esc         quit #! Quit
 ```
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 	<a href="https://user-images.githubusercontent.com/47283320/195073006-bfa72bcc-89d2-4dc7-b8dc-f3c13273910c.webm"><img src="https://user-images.githubusercontent.com/47283320/195072935-44d591d9-00bb-4a55-8795-9cf81f65d397.png" alt="Preview screenshot"></a>
 </div>
 
-Notable features:
+Features:
 
 -   UI elements hide and show based on their proximity to cursor instead of every time mouse moves. This provides 100% control over when you see the UI and when you don't. Click on the preview above to see it in action.
 -   When timeline is unused, it can minimize itself into a small discrete progress bar.
@@ -31,7 +31,7 @@ Notable features:
 
 [Changelog](https://github.com/tomasklaen/uosc/releases).
 
-## Installation
+## Install
 
 1. These commands will install or update **uosc** and place a default `uosc.conf` file into `script-opts` if it doesn't exist already.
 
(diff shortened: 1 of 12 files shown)


dyphire:
I prefer to use this [mpv_manager](https://github.com/po5/mpv_manager) script to uniformly manage mpv scripts and shader updates, which is simpler and more convenient. Maybe you just need to link it in the Docs and provide a usage method.
Example:
```json
[
  {
    "git":"https://github.com/tomasklaen/uosc",
    "whitelist":"scripts$",
    "dest":"~~/scripts",
    "branch":"main"
  }
]
```

tomasklaen:
The quotes are usually omitted when string is passed as an array argument. Just as there are no quotes in powershell command above either, and it works fine. But duno, maybe it's different for bash? Did you try adding the quotes to the args array?

All right, I'll install some linux VM.

po5:
In my case it seems to be caused by the wrong shebang:
`[uosc] bash: line 1: #!/bin/bash: No such file or directory`
This is the proper portable shebang, same used in thumbfast:
`#!/usr/bin/env bash`
You'll want to update the bash call in Updater.lua as well.

tomasklaen:
I'll look into it more tomorrow, but in the meantime:

@po5 you seem to have a lot of experience with running subprocesses from mpv scripts :) Is it possible to:
- Get the stdout as it is coming in asynchronously? Currently the updater just uses `mp.command_native_async()` which just accumulates the whole stdout and dumps it to the final callback. I'd like to display the progress messages so it doesn't feel frozen.
- Get path to directory with current mpv binary?

christoph-heinrich:
idk about having scripts update themselves. It's very reminiscent of the windows ecosystem where each application has it's own update mechanism built in (although I heard something that package mangers are becoming more common?).
I guess it could be useful for people that only use/update uosc and no other scripts.

tomasklaen:
> It's very reminiscent of the windows ecosystem where each application has it's own update mechanism built in 

Yeah I absolutely hate it too, but what else are they supposed to do when there's no package manager? Windows store absolutely sucks.

And it's the same here. If I could make the update screen be a UI for some mpv script manager, I would, but there's nothing established or not abandoned, let alone something with an API. So at least I make it for uosc.

> I guess it could be useful for people that only use/update uosc and no other scripts.

That's exactly my case on my media laptop :) I just want to quickly update uosc from time to time and get to watching. That's the main reason I'm adding this updater - I want to make my life easier.

po5:
To run the update script, this works: `{'bash', '-c', 'source <(curl -fsSL https://raw.githubusercontent.com/tomasklaen/uosc/HEAD/installers/unix.sh)'}`

> Get the stdout as it is coming in asynchronously

Don't think you can do that, but you can redirect output to a file and read it through lua.

> Get path to directory with current mpv binary

What are you trying to do with this? I expect this to only touch the mpv config dir, whose path you already have access to.

tomasklaen:
> To run the update script, this works: {'bash', '-c', 'source <(curl -fsSL https://raw.githubusercontent.com/tomasklaen/uosc/HEAD/installers/unix.sh)'}

Thank you! I was just scratching my head on this :)

> What are you trying to do with this? I expect this to only touch the mpv config dir, whose path you already have access to.

I need to run the subprocess in the same directory as mpv.exe, so that the `windows.ps1` install script can detect if there's `portable_config` folder next to the current `mpv.exe`, and use that as the config root.

Or maybe I should instead figure out how to set some `MPV_CONFIG_DIR` environment variable for the subprocess, and use that in the installers when present.

Pointers for each would be appreciated :)

po5:
> I need to run the subprocess in the same directory as mpv.exe

You don't need to do that. If the user previously installed uosc in that dir, then it will be running from that dir, and that's what will be returned by `mp.get_script_directory()`.

tomasklaen:
So I've solved most issues. The updater now always updates exactly the uosc instance that started it. The config path is just added to existing environment variables as `MPV_CONFIG_DIR`, and installers pick it up.

The only issue is this command args table:

```
{'/bin/bash', '-c', 'source <(curl -fsSL https://raw.githubusercontent.com/tomasklaen/uosc/HEAD/installers/unix.sh)'
```

Works fine on PopOS, but on Ubuntu I get "curl: command not found" error, even though curl is installed (using the linux install command from readme in terminal works fine), and I'm out of ideas how to fix it :(

christoph-heinrich:
What if you download the script to /tmp/ and then run it with bash in a separate `mp. command_native_async()` call, that way an unsuccessful curl execution would also not result in `success=true`. That would also allow passing arguments to the script, instead of having to set environment variables.

I don't know what to do about the curl call failure. You can try `/bin/env curl` like in the shebang, but no idea if it'll help.

tomasklaen:
I did `which curl` which told me it's at `/usr/bin/curl`, so I hardcoded it into the command:

```lua
local url = 'https://raw.githubusercontent.com/tomasklaen/uosc/HEAD/installers/unix.sh'
local args = {'/bin/bash', '-c', 'source <(/usr/bin/curl -fsSL ' .. url .. ')'}
```

And the result:

image


If I try first downloading the script with:

```lua
local args = {'curl', '-Ls', '-o', '/tmp/uosc-unix-installer.sh', url}
```

I get:

```
[uosc] success: true 
[uosc] error: nil 
[uosc] result: {"status" = -3, "stdout" = "", "killed_by_us" = false, "stderr" = "", "error_string" = "init"} 
```
No stdout or err, just an empty fail.

If I try `{'which', 'curl'}` first to get path to curl, I get:

```
[uosc] success: true 
[uosc] error: nil 
[uosc] result: {"error_string" = "", "stdout" = "", "stderr" = "", "status" = 1, "killed_by_us" = false} 
```

Another empty error.

`{'/bin/env', 'curl', '--help'}`:

```
[uosc] success: true 
[uosc] error: nil 
[uosc] result: {"status" = -3, "killed_by_us" = false, "stdout" = "", "error_string" = "init", "stderr" = "/bin/env: ‘curl’: No such file or directory
[uosc] "} 
```

`{'/bin/env curl', '--help'}`:

```
[uosc] success: true 
[uosc] error: nil 
[uosc] result: {"stderr" = "", "stdout" = "", "status" = -3, "error_string" = "init", "killed_by_us" = false} 
```

All commands above executed with:

```lua
mp.command_native_async({
	name = 'subprocess',
	capture_stderr = true,
	capture_stdout = true,
	playback_only = false,
	args = args,
}, function(success, result, error)
	print('success:', utils.to_string(success))
	print('error:', utils.to_string(error))
	print('result:', utils.to_string(result))
end)
```

I don't know what else to try.

This is Ubuntu 23 btw, and only when running anything through `mp.command_native_async()`. Using these commands in terminal works fine. In PopOS this PR works fine too.

It's a fresh install of `ubuntu-23.10.1-desktop-amd64.iso` and then running `sudo apt install curl` as it wasn't preinstalled. I think this is the most popular Linux distro? So kinda important to make it work there.

Oh wait, the mpv is installed as a snap package. I think that might have something to do with it.

christoph-heinrich:
Splitting up `There has been an error. See above for clues.` might enable the first part to be reused for some other error in the future.

Also isn't `An error has occured` more common in error messages? It's fine either way.

PR #701, opened 2023-10-15
fix: update german translation

christoph-heinrich:


tomasklaen:
I'll put this on hold a bit since the updater has some more strings that'll need translating :)

PR #704, opened 2023-10-15
feat: add transparent full volume icon

christoph-heinrich:
image

ref. https://github.com/tomasklaen/uosc/pull/699#issuecomment-1763430615

### DIFF
--- a/scripts/uosc/elements/Timeline.lua
+++ b/scripts/uosc/elements/Timeline.lua
@@ -418,9 +418,8 @@ function Timeline:render()
 			ass:rect(ax, ay, bx, by, {
 				color = bg,
 				border = 1,
-				opacity = config.opacity.thumbnail,
+				opacity = {other = config.opacity.thumbnail, border = 0.08 * config.opacity.thumbnail},
 				border_color = fg,
-				border_opacity = 0.08 * config.opacity.thumbnail,
 				radius = state.radius,
 			})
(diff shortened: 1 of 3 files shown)


po5:
Borders blend together making them stronger than they should be. Removing the outline from the real icon and giving the faded full icon a regular opacity border (which it seems like uosc's ass wrappers don't support) would be better.

christoph-heinrich:
> which it seems like uosc's ass wrappers don't support

Now they do :smile: or at least text does.

tomasklaen:
Looks great 👍, but I don't like that this opacity handling is copy pasted in 2 places, especially since there's already an `ass_mt:opacity()` above. It'd be better to refactor _that_ to accept your table signature, and reuse it in `:icon()` and `:txt()`.

But rename `other` to `main`.

christoph-heinrich:
> I don't like that this opacity handling is copy pasted in 2 places, especially since there's already an `ass_mt:opacity()` above. It'd be better to refactor _that_ to accept your table signature

And I don't like that this isn't as simple as it sounds, I also don't like the solution...

tomasklaen:
I don't think the `self == nil` thing is necessary. Wouldn't it work to just finish the tags, apply them, then call `:opacity()`, and then apply value?

Maybe I'm misremembering, but I don't think all tags need to be in the same `{...}`. That's how `:opacity()` worked until now.

christoph-heinrich:
Yes but then that appends to `self.text` 3 times instead of one and I don't like that.

PR #708, opened 2023-10-16
Fix touch input handlers

### LINKED ISSUE #706: Touch input problem with the current way of handling cursor events
Setting event handlers during render comes with one big problem: they don't get updated until there is a render.

That leads to situations like tapping the volume slider to get an event handler, then actually changing the volume, then taping outside of the slider and changing the volume again because the handler is still from the volume slider.

christoph-heinrich:
This not only prevents the problem described in the issue, but it also allows all rendered elements to be tapped, even when a different element was hovered on last render.

It still requires hovering an element before being able to tap anything, but that's probably unavoidable without always having mbtn events registered. Registering clicks on mbtn up as well we mbtn down could alleviate that problem for buttons, but that wouldn't work universally.

Known problems
* `cursor.allow_dragging = true` in the top bar only works when the cursor was already on the top bar on the last render, not a big deal
* ~~timeline chapter markers require hover before being clickable because of their round hitbox. We can either make the hitbox squares or make `cursor:find_main_handler()` able to deal with circular hitboxes.~~

Fixes #706

### DIFF
--- a/scripts/uosc/elements/Button.lua
+++ b/scripts/uosc/elements/Button.lua
@@ -34,7 +34,7 @@ end
 function Button:render()
 	local visibility = self:get_visibility()
 	if visibility <= 0 then return end
-	cursor:on_main('primary_down', self, function() self:handle_cursor_down() end)
+	cursor:zone('primary_down', self, function() self:handle_cursor_down() end)
 
 	local ass = assdraw.ass_new()
 	local is_hover = self.proximity_raw == 0
--- a/scripts/uosc/elements/Button.lua
+++ b/scripts/uosc/elements/Button.lua
@@ -34,9 +34,7 @@ end
 function Button:render()
 	local visibility = self:get_visibility()
 	if visibility <= 0 then return end
-	if self.proximity_raw == 0 then
-		cursor.on_primary_down = function() self:handle_cursor_down() end
-	end
+	cursor:on_main('primary_down', self, function() self:handle_cursor_down() end)
 
 	local ass = assdraw.ass_new()
 	local is_hover = self.proximity_raw == 0
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -1065,12 +1065,11 @@ function Menu:render()
 		end
 	end
 
-	cursor.on_primary_down = self:create_action(function() self:handle_cursor_down() end)
-	cursor.on_primary_up = self:create_action(function() self:handle_cursor_up() end)
-	if self.proximity_raw == 0 then
-		cursor.on_wheel_down = self:create_action(function() self:handle_wheel_down() end)
-		cursor.on_wheel_up = self:create_action(function() self:handle_wheel_up() end)
-	end
+	local display_rect = {ax = 0, ay = 0, bx = display.width, by = display.height}
+	cursor:on_main('primary_down', display_rect, self:create_action(function() self:handle_cursor_down() end))
+	cursor:on_main('primary_up', display_rect, self:create_action(function() self:handle_cursor_up() end))
+	cursor:on_main('wheel_down', self, function() self:handle_wheel_down() end)
+	cursor:on_main('wheel_up', self, function() self:handle_wheel_up() end)
 
 	local ass = assdraw.ass_new()
 	local spacing = self.item_padding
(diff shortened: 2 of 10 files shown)


christoph-heinrich:
I was curious how ugly round hitbox support would be, so I've added it. It's not so bad.

tomasklaen:
There's a lot I don't like about this. Mainly that it doesn't solve "primary_down fired by tap before `render()` registered handler for that position" problem.

But also, when you look at the current API:

```lua
if self.proximity_raw == 0 then
  cursor.on_primary_down = function() end
end
```

It's obvious what is going on: when the event is being bound, and that it can be overwritten, as this is just a simple prop assignment.

But with this:

```lua
cursor:on_main('primary_down', self, function() end)
```

Almost everything is implicit and anyone looking at it would have no idea why are we binding at all times, and that it can be ignored, or overwritten by something else. You have to read the implementation to understand this. And this code base is already pretty hard to pick up without introducing head scratchers like this :)

---

But I came up with an alternative solution that solves hopefully all/most major issues, but I don't have a touch laptop around me atm to actually test it on.

So what do you think about #710?

christoph-heinrich:
> Almost everything is implicit

We could directly add to the list instead of having a method do it for us if you think that will aid readablility, or maybe rename the method? "register_event_area" maybe.

tomasklaen:
Yeah, renaming to `cursor:zone()`, `cursor:clear_zones()` and `cursor.zone_handlers` would be a lot better.

christoph-heinrich:
renamed

tomasklaen:
All right, I have a different angle to approach this issue, but first, can you please check if `move` is fired _before_ of _after_ `down` when you tap?

christoph-heinrich:
I've checked how osc.lua deals with that, and they listen to `mtbn_left` whenever the osc is visible. Interestingly they also use `mouse_move`, `mouse_enter` and `mouse_leave` instead of `mouse-pos`.

christoph-heinrich:
> Is the `mouse-pos` being retrieved here already updated with the correct position where the down happened?

yes

tomasklaen:
All right, so we don't need any event queuing then.

My idea is to always bind mbtn_left, parse `mp.get_property_native('input-bindings')` for a mbtn_left binding if there's one (we can just reuse already existing code from inputs menu), and call it manually on `mbtn_left_down` when no zone matches.

We would bind the mbtn_left with `allow-vo-dragging` by default, and only disable it if something calls `cursor:disable_dragging()`, like speed/volume elements on primary_down, and then `cursor:enable_dragging()` when dragging is over.

Though there's a question how it behaves. Can dragging be disabled _after_ the primary_down? If not, we'd have to be juggling this flag depending on what the cursor hovering at the moment.

christoph-heinrich:
Event forwarding should work now, but I haven't dealt with `allow_dragging` yet.

We need to somehow make sure that we get all `mbtn_left` events. Currently it works because enabling happens on first render and other scripts that want to always listen to those events register their handlers immediately on startup.

christoph-heinrich:
Do you have any idea why window dragging doesn't work when the cursor is dragged on a scrollable surface (volume, speed, menu), but does work otherwise?
This makes no sense...

christoph-heinrich:
Apparently as soon as there is a mouse binding that doesn't explicitly allow window dragging, it doesn't drag anymore.

A surface that enables right click keybind without `allow-vo-dragging`? No dragging for you.
A surface that enables scroll wheel keybinds without `allow-vo-dragging`? No dragging for you.

Based on that I'd say allow dragging everywhere except scrollable surfaces?

christoph-heinrich:
I've tested it on my laptop now.

Immediately disabling window dragging when touching e.g. volume doesn't work unfortunately, and there is nothing we can do about that. I even tried adding a `mouse_move` key bind to update the cursor position and call `decide_keybinds()` but to no avail.

It works pretty well otherwise.

tomasklaen:
> Idk if you want to do something more sophisticated

I think we should bind it and other handlers only if there's any zone for them on the screen. Doing it dynamically like that would solve some other script taking over potentially, and I also don't want to break the use case @po5 was asking for of having some other UI, but using uosc only for the menu. If we just indiscriminately bind mbt_left, it'll break other UI's. It might break them even now (like slightly visible volume takes over all mbtn_left), but lets at least try to prevent it as much as we can.

christoph-heinrich:
> Doing it dynamically like that would solve some other script taking over potentially

But when the timeline is always at least a little bit visible (as is the case for me) it won't ever reenable it.

>  I also don't want to break the use case @po5 was asking for of having some other UI, but using uosc only for the menu.

We forward unused events now, so that shouldn't be a problem.

tomasklaen:
> But when the timeline is always at least a little bit visible (as is the case for me) it won't ever reenable it, so we're back in the same situation again.

We shouldn't define timeline zones if it's in progress mode, only when `proximity > 0`, just like all other elements.

christoph-heinrich:
Done

PR #710, opened 2023-10-16
fix: touch events firing before handlers are bound

### LINKED ISSUE #706: Touch input problem with the current way of handling cursor events

tomasklaen:
Implements an event queue for main `*_down` events bound during render ticks.

All `*_down` events are delayed until the end of the next render tick.

All `*_up` events are ensured to be triggered after their `*_down` counterparts if there's one in the queue.

And all events are deduplicated so there's not more than one of the same type in the queue in sequence, although it's hard to imagine how that could happen.

closes #706

### DIFF
--- a/scripts/uosc/lib/cursor.lua
+++ b/scripts/uosc/lib/cursor.lua
@@ -28,6 +28,14 @@ local cursor = {
 	mbtn_left_enabled = nil,
 	mbtn_right_enabled = nil,
 	wheel_enabled = nil,
+	main_queue = {
+		required = make_set({'primary_down', 'secondary_down'}),
+		dependent = {primary_up = 'primary_down', secondary_up = 'secondary_down'},
+		---@type table<string, boolean>
+		has = {},
+		---@type string[]
+		events = {},
+	},
 }
 
 cursor.autohide_timer = (function()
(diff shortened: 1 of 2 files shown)


christoph-heinrich:
> it doesn't solve "primary_down fired by tap before render() registered handler for that position" problem

This doesn't behave any better in this regard and adds one frame delay do input events.

christoph-heinrich:
I've now tested it and other then the slight increase in delay it behaves identically as far as I can tell.

tomasklaen:
Wait I guess I don't understand the problem than.

Is the issue that taps can sometimes not trigger any move events? So we never get the updated position at all when someone taps down and up? Even when they tapped at different location than last move event?

I thought what was going on is that tap triggers primary_down, and then move (or the other way around, doesn't really matter), before the next render tick can bind correct handlers for the updated cursor position. In this case this PR should solve the problem.

christoph-heinrich:
I'll try to give a step by step example.
1. The cursor is close to volume slider, but not on top of it, so no event handlers are registered.
2. Tap on volume slider. Nothing happens because no event handler was registered yet, but on the next render a handler gets registered.
3. a. Tap somewhere in the middle of the window, outside of the volume slider.
    b. Tap on a visible close by element (e.g. close button)
4. Notice that the volume changed based on the y coordinate of the tap, because the registered event handler was still the one from the volume slider.

Both our solutions prevent step 3 from changing the volume, and even activate the other element in 3b, but making step 2 change the volume is only possible if we always listen to mbtn events, which we don't want ofc.

christoph-heinrich:
2. can't queue down yet, because we're not listening to `mbtn_left` yet.

PR #711, opened 2023-10-16
tweak: filter out keybindings from uosc in inputs menu

christoph-heinrich:
When looking searching keybinds it's not useful to get all the temporary bindings from uosc.


christoph-heinrich:
#708 already does this, so unless you insist on reverting that behavior I guess I can close this.

tomasklaen:
No no, I was just about to ask if there's a reason to not close this :)

PR #714, opened 2023-10-17
feat: broadcast uosc version as early as possible

christoph-heinrich:
Scripts that offer a UI variant based on the uosc menu might get called right after startup before uosc manages to respond to their `get-version` request. As a result the non uosc UI will be shown. If we proactively broadcast the version as soon as possible then we minimize the chance of that happening.

That came up because of https://github.com/po5/memo/issues/16, where in a little test script I had to add a timeout of >= 1 to get the uosc menu, anything less would not reliably open the uosc menu.
With the early broadcast a timeout of 0.8 is enough, which is still not great, but the chance is also greatly increased with smaller values. I'll have to look if something can be done about that on the mpv side at some point, but it's better then nothing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,19 @@ esc         quit #! Quit
 
 To see all the commands you can bind keys or menu items to, refer to [mpv's list of input commands documentation](https://mpv.io/manual/master/#list-of-input-commands).
 
+## Messages
+
+### `uosc-version <version>`
+
+Broadcasts the uosc version during script initialization. Useful if you want to detect that uosc is installed. Example:
+
+```lua
+-- Register response handler
+mp.register_script_message('uosc-version', function(version)
+  print('uosc version', version)
+end)
+```
+
 ## Message handlers
 
 **uosc** listens on some messages that can be sent with `script-message-to uosc` command. Example:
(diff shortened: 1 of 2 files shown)


christoph-heinrich:
I've looked into the init time a bit and somehow my mpv.conf makes it slow.
Timing how long it takes for main.lua to run from top to bottom takes somewhere between 400ms and 900ms, but if I reduce my mpv.conf to the bare minimum required for uosc (all other scripts and input.conf still there), it's consistently <50ms.

I still think an early broadcast is a good idea though.

tomasklaen:
Doesn't this risk potentially breaking something by triggering the setup in people's `uosc-version` message handler twice if they use the current `get-version` API?

christoph-heinrich:
idk, I was worried about that at first, but for memo and quality-menu it's not a problem.
We could rename the message to prevent any potential problems, just in case. `uosc-version-init` or something. 

But that then requires scripts to be modified to benefit from the early broadcast.

christoph-heinrich:
Or we drop `get-version`, because anyone interested is going to listen to `uosc-version` from the very beginning anyway, so it shouldn't matter if they send a `get-version` first or not.

tomasklaen:
Is it possible for us to send the message before other scripts bind their handlers? This unknown is why I made it be a reply message.

christoph-heinrich:
Only if they delay binding their handlers until after initialization.

tomasklaen:
Ok. So let's remove the `get-version` completely, add a new `## Messages` section above `## Message handlers` in readme, and describe `uosc-version` and its use there.

And add a new line between the value and `mp.commandv` call at the top :)

PR #719, opened 2023-10-20
feat: improved `escape` and `backspace` behavior in menus

tomasklaen:
`escape` now only closes the menu, and nothing else (before it would clear search if present first).

New shortcuts:

- `ctrl+backspace` - Delete search query by word.
- `shift+backspace` - Clear search query.

### DIFF
--- a/scripts/uosc/elements/Menu.lua
+++ b/scripts/uosc/elements/Menu.lua
@@ -88,8 +88,8 @@ function Menu:init(data, callback, opts)
 	self.opts = opts or {}
 	self.offset_x = 0 -- Used for submenu transition animation.
 	self.mouse_nav = self.opts.mouse_nav -- Stops pre-selecting items
-	---@type Modifiers|nil
-	self.modifiers = nil
+	---@type Modifiers
+	self.modifiers = {}
 	self.item_height = nil
 	self.min_width = nil
 	self.item_spacing = 1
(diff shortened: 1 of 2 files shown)


scripts/uosc/elements/Menu.lua
```diff
@@ -813,13 +813,26 @@ function Menu:search_query_update(query, menu)
...
+	if word_mode then
+		local word_pat, other_pat = '[^%c%s%p]+$', '[%c%s%p]+$'
+		local init_pat = old_query:sub(#old_query):match(word_pat) and word_pat or other_pat
+		-- First we match all same type consecutive chars at the end
+		local tail = old_query:match(init_pat)
+		-- If there's only one, we extend the tail with opposite type chars
+		if tail and #tail == 1 then
+			tail = tail .. old_query:sub(1, #old_query - #tail):match(init_pat == word_pat and other_pat or word_pat)
+		end
+		pos = pos - #tail
```
christoph-heinrich, line 828: I would have expected `ctrl+backspace` to always delete back to the beginning of the last word. From a quick test at least KDE things seem to consistently behave that way.
```suggestion
		pos = old_query:find('[^%c%s%p]*[%c%s%p]*$') - 1
```

christoph-heinrich: Apparently that's not the case for vscode, it behaves like your original code. It doesn't really matter i guess.

tomasklaen: Yeah I've tried making it like vscode, which is how I expected it to work from the beginning. 

For example, if I have a string like:

```
abc.--
```

I'd want `ctrl+backspace` to end before `c`, not delete the whole string.

vscode has a bit more conditional behaviors, but I didn't want to make it that complex.

FichteFoll: > I would have expected `ctrl+backspace` to always delete back to the beginning of the last word.

This behavior is iconsistent across applications, unfortunately. Sublime Text, for example, deletes until the next boundary between word characters, punctuation, or whitespace.

PR #726, opened 2023-10-21
fix: `delete-file-prev` not working as intended

christoph-heinrich:


tomasklaen:
Aren't we running a risk here that `state.path` will update between `navigate_item()` and `delete_file()`? It kind of relies on `navigate_item()` being implicitly asynchronous, and it also makes the code harder to follow.

Rather save it into a `current_path` at the top and use that to delete from menu and drive.

christoph-heinrich:
The script runs in a single thread, unless we update it in that code path it can't change until we're done and the property observer can run.

tomasklaen:
Yeah but look at the code flow from a perspective of someone not that familiar with all mpv/uosc internals (or me in a couple years going back trying to maintain it :)). They'll see `navigate()`, and then us deleting the newly open file, and go "huh?".

christoph-heinrich:
The last line of the original used state.path... 
```lua
if is_local_file then delete_file(state.path) end
```

tomasklaen:
> The last line of the original used state.path...

Yeah that was bad too then :) I'm just trying to make the code easy to follow, is all.

PR #728, opened 2023-10-21
add. ukrainian translation

Helltar:


tomasklaen:
Could you add a localization for `"Play/Pause"` string which just got added in?

PR #739, opened 2023-10-25
fix: runtime updates for progress option

christoph-heinrich:
Fixes #373

<hr/>

Now that elements can be created/destroyed at runtime, an alternative way of dealing with script-opts updates would be to destroy and recreate everything. If it works on startup, it'll also work during runtime.


tomasklaen:
Lets not do that unless necessary. I'm afraid of side effects of doing that while stuff like menu or updater or future color adjustments are open. Things the user is interacting with shouldn't just disappear from the screen because next file with different options profile started playing.

PR #744, opened 2023-10-26
build: added building, packaging, and localization scripts and tools

tomasklaen:
Main feature here is the `intl` tool that allows people to quickly create or update localization files with simple:

```
tools/intl de
```

This will parse the codebase for localization strings, and compare it against `de` locale, removing unused, and assigning `null` to new strings.

We can also make it even easier for people by running:

```
tools/intl all
```

after every localization string change, which will update all existing locales so people can just browse through theirs and fill in nulls.


po5:
That's nice, but instead of having binaries in the git repo it'd better to attach them to a release.

tomasklaen:
This is a development tool, it's not going to be in a release. It _should_ be a script, but since this project isn't built around any runtime, we can't have a script file without requiring people to install some lua, python, or node. I wanted something people can just use with no setup barriers.

I look at it as a script that needs to be compiled :)

I'm not a fan of binaries in git either, but this seems to be the least friction solution.

Hrxn:
Question regarding `tools/package.ps1` is this intended to be run by GitHub Actions or some other CI/CD tools?

Because otherwise I don't get the point.

https://github.com/tomasklaen/uosc/blob/55789e1d325d53e5792195514cc998cb0ac66055/tools/package.ps1#L33-L38

Is there another intermediate build step involved?
Maybe I'm just looking in the wrong place, but I don't see a `dist/fonts` directory in the repo?
And, as I understand it, the files in `/dist` are already automatically up-to-date builds made after changes have been made inside of `/src`, or do I get this wrong?

tomasklaen:
No actions are set up atm. Packaging tool is for me to make the release zip used by command line installers and people installing manually.

> Maybe I'm just looking in the wrong place, but I don't see a dist/scripts directory in the repo?

It's there.

> And, as I understand it, the files in /dist are already automatically up-to-date builds made after changes have been made inside of /src, or do I get this wrong?

They are not generated, they are the sources. The way it's divided is `src` is for things that need to be built. `dist` is for things that don't, and lua files don't need to be built, so they sit in `dist`.

Soon there will be an intermediate built step to build some binaries that'll need to be included in the release zip. Their build destination will be in `dist` but they'll be gitignored.

I agree it's a bit of a mess, but we're juggling a lot of different environments (go, lua, mpv) and assets (fonts, soon binaries), and this seems to be the easiest to develop with. I just simlink stuff from `dist` to my mpv and that's it. No build step necessary during development. And packaging is also easy, you just zip what's in `dist`.

Hrxn:
Ah, yeah, `dist/scripts/uosc` must've missed that somehow.

> They are not generated, they are the sources. The way I look at it is src is for things that need to be built. dist is for things that don't, and lua files don't need to be built, so they sit in dist.

Makes sense, good to know.
Also what I've been doing so far, simply taking the files fom `/dist` to "install" uosc.

tomasklaen:
@debugzxcv 😐 thx for noticing.

PR #747, opened 2023-10-28
fix: update german translation

christoph-heinrich:
The new intl tool already discovered a missing string: `%s are empty`

Removed were:
```
Autoselect device
Edition
Track
```


tomasklaen:
I was holding off on the localization PRs since there was the subtitle downloader PR which added more strings to translate. And now it needs to be updated _and_ rebased. Sorry about that 😐.

christoph-heinrich:
Updated

PR #748, opened 2023-10-28
fix: update simplified chinese translation

dyphire:


tomasklaen:
I was holding off on the localization PRs since there was the subtitle downloader PR which added more strings to translate. And now it needs to be updated _and_ rebased. Sorry about that 😐.

dyphire:
> I was holding off on the localization PRs since there was the subtitle downloader PR which added more strings to translate. And now it needs to be updated _and_ rebased. Sorry about that 😐.

Done.

Issue #722, opened 2023-10-21
Timeline font size in version 5.0.0


LampPrinter:
![iShot_2023-10-21_17 19 43](https://github.com/tomasklaen/uosc/assets/7559328/d15ba0a9-9913-42d5-88e0-a746b4a7090e)

Please see the attached photo. The font size in previous version of UOSC was larger.
I only see "font_scale=1" which makes all text bigger including title etc. which I do not want.

tomasklaen:
This should fix it. If you could test what's on main, refer back if it doesn't.

LampPrinter:
Yes it fixed it.
Another issue, all the text now has a dark background. Whereas earlier it was not like this.
![195072935-44d591d9-00bb-4a55-8795-9cf81f65d397](https://github.com/tomasklaen/uosc/assets/7559328/664bdd28-ca44-4b9f-9a17-af4295964a0f)

Am I missing some setting in uosc.conf?

tomasklaen:
Related: #636

You can disable it with the new `opacity` option.

bananakid:
@tomasklaen would you kindly tell if there's a way to set semi-transparent background for Controls? I've tried `opacity=controls=0.5` and it didn't work while `opacity=timeline=0.5` worked fine for Timeline.

tomasklaen:
Controls don't have a background. But it wouldn't be a problem to add it now that we have the `opacity` option. I'll add it to a to-do list.

bananakid:
@tomasklaen, thank you for support!

bananakid:
@tomasklaen, by the way it appears to me that Volume doesn't respect `volume_border=0` in 5.0.0, I can make the border thicker but it retains thin black line even if zero is specified in `uosc.conf`.

christoph-heinrich:
> the opacity of the progress bar when `timeline_style=bar` doesn't respect `opacity=timeline=0.3` value specified in `uosc.conf`.

`opacity=position=0.3`

bananakid:
@christoph-heinrich, thanks, I've edited the initial comment since I figured how some of the things work (`position` was among them). Would you kindly tell if it's possible to set the color and opacity of the indicator that appears when you hover mouse cursor on the Timeline? It looks like a thin vertical line in full height of timeline right where the cursor is pointing.

christoph-heinrich:
> Another small issue is when milliseconds are enabled in Timeline the time indicator on the right keeps moving

[...]

PR #752, opened 2023-10-28
fix: shifting of digits due to non monospace font

christoph-heinrich:
Why draw the whole string at once when you can draw each symbol individually two times!

Not sure about this but it does solve the shifting digits mentioned in the comment.

~~Would be nice to also use this in the time tooltip, but I don't see an easy way of doing that.~~

Ref. https://github.com/tomasklaen/uosc/issues/722#issuecomment-1783866244

### DIFF
--- a/dist/scripts/uosc/elements/Timeline.lua
+++ b/dist/scripts/uosc/elements/Timeline.lua
@@ -351,6 +351,34 @@ function Timeline:render()
 		ass:txt(x, y, align, text, opts)
 	end
 
+	local function draw_semi_monospace_time(x, y, align, time_human, opts)
+		local widths, width_total = {}, 0
+		zero_rep = timestamp_zero_rep(time_human)
+		for i = 1, #zero_rep do
+			local width = text_width(zero_rep:sub(i, i), opts)
+			widths[i] = width
+			width_total = width_total + width
+		end
+		if align == 6 then
+			x = x - width_total
+		elseif align == 5 then
+			x = x - width_total / 2
+		end
+		local opacity = opts.opacity
+		opts.opacity = {main = opacity, primary = 0}
+		for i, width in ipairs(widths) do
+			draw_timeline_text(x + width / 2, y, 5, time_human:sub(i, i), opts)
+			x = x + width
+		end
+		x = x - width_total
+		opts.opacity.main, opts.opacity.primary = 0, opacity
+		for i, width in ipairs(widths) do
+			draw_timeline_text(x + width / 2, y, 5, time_human:sub(i, i), opts)
+			x = x + width
+		end
+		opts.opacity = opacity
+	end
+
 	-- Time values
 	if text_opacity > 0 then
 		local time_opts = {size = self.font_size, opacity = text_opacity, border = 2}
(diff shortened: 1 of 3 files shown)


po5:
libass supports some generic font families that ["map to sensible platform-specific font families"](https://github.com/libass/libass/blob/e64390522d9501a2c7e6a51c31a5b2a1f3a35cd1/libass/ass_fontselect.h#L115-L131), if we don't care about respecting the user's osd-font for timestamps: `sans-serif`, `serif`, `monospace`

christoph-heinrich:
Yes but as I pointed out [here](https://github.com/tomasklaen/uosc/issues/722#issuecomment-1783880134), I don't like how it looks.

monospace
image
this hack
image

dyphire:
LGTM

christoph-heinrich:
I thought of a way to use it in the tooltip, but the code is kinda ugly...

tomasklaen:
I don't know about this one. I'm concerned about the performance. That's a lot more ass tags this is pushing. Could you do a benchmark and report the difference between lets say drawing a hundred timestamps with `:txt()` and the same with `:timestamp()`?

I know there's a limit of ass tags after which libass can't keep up and UI starts stuttering, and the amount of ass tags this adds just doesn't feel like it's worth it.

christoph-heinrich:
I was also worried about the performance impact at first, but it works fine on a 10 year old machine locked to it's lowest frequency, so it's probably fine???

100 `:timestamp()` takes 10.3 ms
100 `:txt()` takes 2.1 ms

So yeah pretty big difference, but we're not actually drawing that many of them.

When hovering the timeline so that three timestamps are drawn the times are:
`:timestamp()` takes 810 us
`:txt()` takes 290 us

That's a big difference relatively speaking, but for some more context, leaving the timeline so that the controls are shown drives the `:txt()` time up to 570 us, and that is with not that many of them.
image

So it should be fine, as long as we're not drawing the timestamps of chapters in the chapter menu with them.

tomasklaen:
This makes me feel like adding a toilet paper roll asset made of 1M polys into a game. To remedy a problem that affects <1% people.

I would hate knowing that timestamps are this inefficient for 99% that absolutely don't need it, but I wouldn't mind merging this if we only used `:timestampt()` when `time_precision > 0`.

christoph-heinrich:
It is also a problem with `time_precision=0` but less noticeable. I already noticed it months ago without increased precision, but I suspected it might not be a problem with other fonts (didn't test) and it's not that noticeable.

image
image

As you can see the `11` takes up a little bit less space, thus shifting the rest to the right (opening the images in tabs and switching between them makes it easier to see).

But I'm also somewhat bothered by the additional performance impact, so there you go :)

christoph-heinrich:
I've tested other fonts now (should have done that much sooner) and the only ones that have that problem for me are the Liberation variants (except mono), so I guess we could say "don't use a shitty font" and be done with it.

FichteFoll:
If performance is still a concern, you could probably speed it up a little by caching the results of [this block](https://github.com/tomasklaen/uosc/commit/6cbf0731f3ce24df215e44e67da6c4d16d64eed5#diff-f313b2b2f9768dd7521bbd9b5474aba8c8205f5481b13dae735f6cd9b7a29c55R122-R128) (for the same font, which I'm not sure whether it is configurable dynamically).

christoph-heinrich:
Lua performance isn't the problem here :wink: it's about the speed of libass.

FichteFoll:
I was thinking of the `text_width` calls primarily, but it seems that is already cached, so you can disregard my comment.

PR #756, opened 2023-10-29
feat: added menu for downloading subtitles from opensubtitles.com

### LINKED ISSUE #490: Online subtitle search
Maybe it's a bit too much for minimal ui. I was thinking is there any way to search subtitle online?

tomasklaen:
subs

New menu command `download-subtitles` which can also be opened by clicking on the **Download** option in the `subtitles` menu.

It uses a new `ziggy` binary which needs to be build with `tools/build ziggy` command (haven't tested outside windows btw). If you don't want to set up go, you can [download binaries from here](https://github.com/tomasklaen/uosc/files/13198264/bin.zip), and extract them into `dist/scripts/uosc/bin` (you'll probably have to mark linux/mac binaries as executable).

ziggy will also hash the file and send the hash to Open Subtitles, so you can search even with empty query and if your file is known, you'll get results exactly for it.

The subtitle file is downloaded into the same directory as currently opened file, or `~~/subtitles` if URL is being played.

The uosc's Open Subtitles API key is currently set to development mode, so we have 100 global downloads per day. Try not to blow through it when testing :) But it seems like downloading same file over and over doesn't decrement the remaining downloads by more than one, so do that.

closes #490


Zabooby:
Not sure which method is better, but would it be possible to add [this](https://github.com/davidde/mpv-autosub) script into uosc? 

If not it would be nice to see an option to automatically download subtitles if none are detected.

tomasklaen:
> Not sure which method is better, but would it be possible to add [this](https://github.com/davidde/mpv-autosub) script into uosc?

Why does it need to be added to uosc? If you want mpv-autosub functionality, than just install and use mpv-autosub... uosc is a UI, and this PR extends the UI with searching, browsing, and downloading subtitles on demand.

Zabooby:
> Why does it need to be added to uosc? If you want mpv-autosub functionality, than just install and use mpv-autosub... uosc is a UI, and this PR extends the UI with searching, browsing, and downloading subtitles on demand.

I was asking if integrating mpv-autosub would be easier than to start from scratch.

po5:
Why is "Subtitles loaded & applied" and other info in the menu? Would make more sense as a toast or osd message.

tomasklaen:
> Why is "Subtitles loaded & applied" and other info in the menu? Would make more sense as a toast or osd message.

Since the open subtitles limits are quite low (5 downloads per day by default) I wanted to give people a chance to read them and close at their leisure. But I guess it makes sense to cancel the input field on the last window.

dyphire:
I think adding some new script APIs and using them as third-party scripts is much better than directly integrating subtitle downloads, but this may be more challenging.

I really don't like the practice of placing binary files in the script directory. I suggest adding a new option of `ziggy_path` that allows users to specify paths. The `ziggy_path=ziggy` will finding executable files next to the mpv.exe program and in environment variables on Windows, search it in the environment variables at Linux or Mac.

tomasklaen:
I didn't make it a 3rd party script because I realized I'd have to write a standalone UI so it works outside of uosc, and I really didn't feel like doing that. And since it'll only work with uosc, it might as well be part of it.

I also don't want to make ziggy optional. I intend it to become a multitool that'll allow us to do stuff we can't do from within mpv's lua environment.

> I really don't like the practice of placing binary files in the script directory. I suggest...

So make everything more complicated, and add additional installation steps? And what would be the benefit of that? It just adds unnecessary installation steps, undoing my primary goals for making ziggy in go:
- no dependencies on external runtimes
- **no additional installation steps or setups**
- everything _just works_

People already put python files in their mpv scrips. This is pretty much the same thing it just doesn't depend on external runtime.

dyphire:
> > I really don't like the practice of placing binary files in the script directory. I suggest...
> 
> So make everything more complicated, and add additional installation steps? And what would be the benefit of that? It just adds unnecessary installation steps, undoing my primary goals for making ziggy in go:
> 
> * no dependencies on external runtimes
> * **no additional installation steps or setups**
> * everything _just works_
>

The default value of `ziggy_path=~~/scripts/uosc/bin` will do the same thing, but do not hardcode the path. This way, users who need it can manually change the options to avoid binary files being placed in the script folder.
This is just an improvement without adding complex or additional operations, as there are no changes by default.

dyphire:
> I didn't make it a 3rd party script because I realized I'd have to write a standalone UI so it works outside of uosc, and I really didn't feel like doing that. And since it'll only work with uosc, it might as well be part of it.

There is already a third-party script that is only applicable to uosc: [recent-menu](https://github.com/natural-harmonia-gropius/recent-menu), and I don't think it's bad.
The reason for raising this point is that I believe subtitle downloading is an optional feature, and the necessity of integrating it is not enough

dyphire:
> > This way, users who need it can manually change the options to avoid binary files being placed in the script folder.
> 
> Are you going to change `ziggy_path` and copy & paste the `bin` folder to a different place after every uosc update? 😬 ... why?

As I said, the behavior of placing binary files in the script folder troubles me. You can understand it as a form of obsessive-compulsive disorder.

gunir:
Great feature, honestly not being able to get subtitle easily was always the downside of MPV vs VLC which has `VLCSub.lua` plugin which download subtitles from OpenSubtitles, thanks for intergrating it into uosc!

This is so great and if it's easy to extend, maybe we should allow people to add new subs sources.

AndydeCleyre:
I'm excited for this!

The error messages could be more informative, at least if reading terminal output. My naive testing of the new feature always reports "process exited with code -3  error" in the mpv window when I submit a search, but I don't know what that signifies.

Will it be possible to use our own opensubtitles login to increase query limits?

po5:
> Will it be possible to use our own opensubtitles login to increase query limits?

Already possible with the `open_subtitles_api_key` script-opt, which isn't documented for whatever reason.  
For tomas: If it's trying to obscure the builtin key from the user, make an empty string default to that key. That way it's not in the config file and can be silently updated from your side if the need arises.

tomasklaen:
> Will it be possible to use our own opensubtitles login to increase query limits?

As far as I understand from reading the open subtitles docs, authenticating users bumps the limit from 5 to a WHOLE 10, so I didn't feel like it was worth spending the time implementing it. You can increase it further by having some VIP accounts or something, but how many people will do that...

And 5 downloads per day isn't _that_ bad, since if you need more than 5, you should probably just deal with it in the browser beforehand so you don't have to bother every time a new episode starts. But I do agree that it's insanely small for what this is. Even the search result payloads are often bigger than the downloaded subtitles, and those are unlimited and need database queries to put together.

> Already possible with the open_subtitles_api_key script-opt, which isn't documented for whatever reason.
> For tomas: If it's trying to obscure the builtin key from the user, make an empty string default to that key. That way it's not in the config file and can be silently updated from your side if the need arises.

The `open_subtitles_api_key` variable is internal config that is not exposed to user options. The way open subtitles work is the API keys are per app, and users should be authenticated with their `username:password` pairs. This is annoying to implement (especially since iirc the session lasts only 24 hours), and bumps the limit only to 10, so I didn't bother.

The user _could_ theoretically make their own API key on open subtitles, set it to development mode (which gives you 100 downloads per day global, shared with all IPs using it), and use it only for themselves, but that feels like being a bad consumer, and I don't want open subtitles to start blocking our user agent.

So in other words, I don't want to do stuff that doesn't adhere to open subtitles rules and guidelines.

> maybe we should allow people to add new subs sources.

Afaik there are no other subtitle services with an API endpoints, and I'm not maintaining html serializers and trying to work around rate limiting, captchas, and stuff like that.

tomasklaen:
> The error messages could be more informative, at least if reading terminal output. My naive testing of the new feature always reports "process exited with code -3 error" in the mpv window when I submit a search, but I don't know what that signifies.

When error happens, try looking in console (backtick key), it should have more info (whole stdout + stderr), since I didn't want to put full error messages into the one line menu item title. I'll add some "see console for details" appendix to the UI message.

AndydeCleyre:
Thanks! When the error occurs, a red `[uosc]` line is added, but nothing else on it. Oh yes, I didn't actively build any go code, so I'll do it now: no, I couldn't use the build script because the `elif` on line 17 has no corresponding `if`.

tomasklaen:
> because the elif on line 17 has no corresponding if

This was fixed yesterday. Pull latest commit from main.

AndydeCleyre:
Sorry, thanks it works great!

PR #758, opened 2023-11-01
fix: script syntax

christoph-heinrich:
Without this I get errors like:
```
tools/build: line 17: syntax error near unexpected token `elif'
tools/build: line 17: `elif [ $1 == "intl" ]; then'
```
and
```
tools/build: line 17: [: ==: unary operator expected
```

I didn't get around to testing this sooner, and neither did anybody else apparently :upside_down_face:


tomasklaen:
I should've mentioned somewhere that I didn't test the building scripts outside windows 😶.

PR #762, opened 2023-11-02
feat!: reworked colors and opacity configuration

tomasklaen:
Don't test this as it doesn't work yet, it's just a proof of concept of what I've been thinking of doing:

---

Options `color` and `opacity` have been removed, and replaced with a single option `colors` which now accepts RGB(A) values for pretty much all colors throughout uosc. Example:

```
colors=timeline=000000e6,timeline_border=000e,timeline_text=fff,timeline_text_border=000,...
```

This simplifies the config (one less option, and opacities are now right next to colors they control) as well as provides more customization than before.

Since this is now just a single option, we could even have a _Themes_ wiki page for people to share their `colors=` along with screenshots.

The only problem is that there's now a lot of these values, and some are quite verbose, so any ideas on how to simplify them?

```lua
local color_defaults = {
	timeline = '000000e6',
	timeline_border = '000e',
	timeline_text = 'fff',
	timeline_text_border = '000',
	timeline_cursor = 'fff3',
	timeline_position = 'fff',
	timeline_position_text = '000',
	timeline_position_text_border = '0000',
	timeline_position_cursor = '0003',
	button = '0000',
	button_icon = 'fff',
	button_icon_border = '000',
	button_hover = 'fff1',
	button_active = 'fff',
	button_active_icon = '000',
	button_badge = 'fff',
	button_badge_text = '000',
	pause_indicator = 'fff',
	pause_indicator_fade = '0004',
	volume = '000000e6',
	volume_text = 'fff',
	volume_range = 'fff',
	volume_range_text = '000',
	speed = '0009',
	speed_text = 'fff',
	speed_text_border = '000',
	speed_notch = 'fff',
	speed_notch_border = '000',
	curtain = '1111',
	menu = '000',
	menu_text = 'fff',
	menu_placeholder = 'fff6',
	menu_item_hover = 'fff1',
	menu_item_active = 'fff',
	menu_item_active_text = '000',
	title = '000000e6',
	title_text = 'fff',
	title_text_border = '000',
	playlist_position = 'fff',
	playlist_position_text = '000',
	playlist_position_text_border = '0000',
	success = 'a5e075',
	warning = 'ffc461',
	error = 'ff616e',
	ads = 'c54e4e80',
	openings = '30abf964',
	endings = '30abf964',
}
```

### DIFF
--- a/src/uosc.conf
+++ b/src/uosc.conf
@@ -148,16 +148,9 @@ font_scale=1
 text_border=1.2
 # Border radius of buttons, menus, and all other rectangles
 border_radius=4
-# A comma delimited list of color overrides in RGB HEX format.
-# Defaults: foreground=ffffff,foreground_text=000000,background=000000,background_text=ffffff,curtain=111111,
-#           success=a5e075,error=ff616e
-color=
-# A comma delimited list of opacity overrides for various UI element backgrounds and shapes.
-# This does not affect any text, which is always rendered fully opaque.
-# Defaults: timeline=0.9,position=1,chapters=0.8,slider=0.9,slider_gauge=1,speed=0.6,menu=1,submenu=0.4,
-#           border=1,title=1,tooltip=1,thumbnail=1,curtain=0.8,idle_indicator=0.8,audio_indicator=0.5,
-#           buffering_indicator=0.3
-opacity=
+# A comma delimited list of color overrides in RGB(A) HEX format (same as CSS). Defaults:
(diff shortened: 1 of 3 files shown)


tomasklaen:
That's a really small learning curve. You don't think it's worth it? I thought people wanted this level of customization.

christoph-heinrich:
I predict that everyone will continue use configurations that are already possible, but ok let's see what happens.

po5:
This makes config a lot more unapproachable than it already is.  
Unless you want to provide a graphical editor for the UI within uosc :^)

dyphire:
Yes, the new configuration is too complex.

tomasklaen:
All right :)

Mark-Joy:
For me, this makes so much sense. How is this "too complex"? It's just a bunch of color items.
Let users decide whether they need it or not. I mean it's fine if their old config file does not have this color item, then just use the default colors.
One beautiful day, they just want to change some colors because the old UI was so familiar, and they just want to tweak it a little bit, some color changes, a little bit here, a little bit there..
For me, this feature makes so much sense!!

Solarunit:
yea, I would like to change color for some elements too.
What about accent color option instead?

PR #771, opened 2023-11-04
fix: only autohide UI when the cursor autohides

christoph-heinrich:
The cursor isn't allowed to autohide while hovering elements, so the UI shouldn't autohide either. Otherwise this leads to the situation where the UI autohides while e.g. hovering the timeline, and then the cursor autohides afterwards because it's not hovering anything anymore.

The `self.autohide_timer:kill()` in function `cursor:autohide()` doesn't change anything functionality wise, but without it the timer keeps periodically moving the mouse to `(inf, inf)` which is unnecessary and annoying when debugging cursor position related stuff.

The `self:queue_autohide()` in `cursor:decide_keybinds()` is necessary because when e.g. the cursor moves 1px off of an element then during the position update there are still the bindings for that element, so UI autohide is prevented, but then on the next render there are no bindings anymore and the UI should autohide just like the cursor.

### DIFF
--- a/src/uosc/lib/cursor.lua
+++ b/src/uosc/lib/cursor.lua
@@ -35,11 +35,11 @@ local cursor = {
 	autohide_fs_only = nil,
 }
 
-cursor.autohide_timer = (function()
-	local timer = mp.add_timeout(mp.get_property_native('cursor-autohide') / 1000, function() cursor:autohide() end)
-	timer:kill()
-	return timer
-end)()
+cursor.autohide_timer = mp.add_timeout(1, function() cursor:autohide() end)
+cursor.autohide_timer:kill()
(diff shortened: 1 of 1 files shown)


tomasklaen:
Should we add [`cursor-autohide-fs-only`](https://mpv.io/manual/master/#options-cursor-autohide-fs-only) check to `:allow_autohide()`?

And `allow_autohide` sounds like it's enabling autohide instead of just checking. Rename it to something like `is_autohide_allowed`.

christoph-heinrich:
`cursor-autohide-fs-only` now gets observed and stored in cursor instead of state, because I thought it's not relevant to anything else.

`cursor-autohide` also gets observed now and updates the autohide time.

PR #795, opened 2023-11-15
fix: make maximize workaround windows specific

christoph-heinrich:
ref. https://github.com/tomasklaen/uosc/issues/203#issuecomment-1811429746


christoph-heinrich:
Turns out there is a race condition here on wayland. Happens on both kwin and weston.
The compositor ignores the unmaximize request if mpv is still considered fullscreen.

I guess the simplest way to deal with that would be to add a small delay between setting `fullscreen` and setting `window-maximized` and hope it's enough, or a hook that waits for `fullscreen` to change before setting `window-maximized`.

No idea if other platforms might have the same problem.

PR #797, opened 2023-11-18
feat: add romanization support for search characters

dyphire:


hooke007:
Wrong indents here too.

dyphire:
Fix formatting.

src/uosc/lib/char_conv.lua
```diff
@@ -0,0 +1,94 @@
...
+		local byteCount = 1
+		if curByte > 239 then
+			byteCount = 4
+		elseif curByte > 223 then
+			byteCount = 3
+		elseif curByte > 128 then
+			byteCount = 2
+		else
+			byteCount = 1
+		end
```
christoph-heinrich, line 45: `utf8_char_bytes()` in text.lua

src/uosc/lib/char_conv.lua
```diff
@@ -0,0 +1,94 @@
...
+	for n, char in utf8_iter(chars) do
+		if ligature then
+			if string.len(char) == 1 then
+				char_conv[#char_conv + 1] = char
+			else
+				char_conv[#char_conv + 1] = pyTable[char] or char
+			end
+		else
+			if string.len(char) <= 2 then
+				if (char ~= ' ' and n ~= chars_length) then
```
christoph-heinrich, line 72: `n` here isn't the nth character, but the index in the string at which the character starts. So for an ASCII string that would be the same, but if the string contains multi-byte utf-8 characters it's not the same anymore.

christoph-heinrich (changes requested):
I haven't looked too closely yet and also didn't test it, but that conversion function looks quite different to how I expected it to be.
Can you please explain what's up with that ligature stuff? I never expected that to be a factor, and it's not obvious to me how that's beneficial (I probably just don't understand).

dyphire:
> I haven't looked too closely yet and also didn't test it, but that conversion function looks quite different to how I expected it to be. 
Can you please explain what's up with that ligature stuff? I never expected that to be a factor, and it's not obvious to me how that's beneficial (I probably just don't understand).

Here is a linguistic knowledge: Chinese and Japanese are [Logogram](https://en.wikipedia.org/wiki/Logogram) text without spaces separating them.

When romanizing, we need to deal with two situations: when searching for full text, there is no need for special processing of logogram text (i.e. ligatures); When searching for the initials, it is necessary to manually insert spaces for separating logogram text (i.e. separator).
For example, the romanization of `芙蓉` is `furong`, and when preparing for the initials, it is `fu rong`, so it's initials is `fr`.

src/uosc/lib/char_conv.lua
dyphire: Yes, what I want here is the nth character, but it seems that I made a mistake. Is there any way to fix it?

christoph-heinrich:
> Here is a linguistic knowledge: Chinese and Japanese are [Logogram](https://en.wikipedia.org/wiki/Logogram) text without spaces separating them.
> 
> When romanizing, we need to deal with two situations: when searching for full text, there is no need for special processing of logogram text (i.e. ligatures); When searching for the initials, it is necessary to manually insert spaces for separating logogram text (i.e. separator). For example, the romanization of `芙蓉` is `furong`, and when preparing for the initials, it is `fu rong`, so it's initials is `fr`.

Ah that makes sense, now I remember that weeks ago I already wondered how the initials thing should work with such languages but I forgot to look into it. Deciding that based on unicode ranges is probably more robust then simple byte count, but if that works well enough in practice then checking unicode ranges would be wasting performance.

Speaking of performance, have you ever tried benchmarking it? Two things that _might_ make a difference are:
1. clearing the cache with `itable_clear()` from std.lua instead of creating new ones.
2. `table.concat` could be stored in a local variable to make accessing it faster.

And a minor nit: We usually use #var instead of string.len(var), but I don't think this makes much of a difference. Doing the local variable trick might actually make it faster then using #, but I've never actually tested that.

dyphire:
> Ah that makes sense, now I remember that weeks ago I already wondered how the initials thing should work with such languages but I forgot to look into it. Deciding that based on unicode ranges is probably more robust then simple byte count, but if that works well enough in practice then checking unicode ranges would be wasting performance.

Yes, it would be better to handle it based on the Unicode range. It was only when I answered your previous question that I noticed that Korean with byte numbers greater than 2 also has space separation, which means that using byte count processing will result in incorrect romanization of Korean in the future. But I'm not sure how much impact this will have on performance.

> 
> Speaking of performance, have you ever tried benchmarking it? Two things that _might_ make a difference are:
> 
> 1. clearing the cache with `itable_clear()` from std.lua instead of creating new ones.
> 2. `table.concat` could be stored in a local variable to make accessing it faster.

It's too late. I'm ready to go to bed. When I wake up, I'll think about it.

tomasklaen:
Dunno when I'll look into this in detail, but my only requirement is **0** performance and memory hit for people that don't need this. So if uosc `languages` option doesn't contain a langauge that needs it, there should be no json files read/parsed, and search adjustment in this case should effectively just be:

```lua
if romanization_not_needed then return true end
```

No additional loops through strings, trying to replace characters from empty tables, and no additional string comparisons.

dyphire:
> Dunno when I'll look into this in detail, but my only requirement is **0** performance and memory hit for people that don't need this. So if uosc `languages` option doesn't contain a langauge that needs it, there should be no json files read/parsed, and search adjustment in this case should effectively just be:
> 
> ```lua
> if romanization_not_needed then return true end
> ```
> 
> No additional loops through strings, trying to replace characters from empty tables, and no additional string comparisons.

Done.

dyphire:
> 
> Speaking of performance, have you ever tried benchmarking it? Two things that _might_ make a difference are:
> 
> 1. clearing the cache with `itable_clear()` from std.lua instead of creating new ones.
> 2. `table.concat` could be stored in a local variable to make accessing it faster.

I have studied your proposal and I did not understand the use of `itable_clear()`.
As for using `table.concat` as a local variable, I guess it refers to the processing of search characters in Menu.lua?

christoph-heinrich:
> I have studied your proposal and I did not understand the use of `itable_clear()`.

Instead of creating a new table `cache = {}` you can do `itable_clear(cache)` to avoid some allocations.

> As for using `table.concat` as a local variable, I guess it refers to the processing of search characters in Menu.lua?

What I mean is that this
```lua
for _, char in utf8_iter(chars) do
    ...
    sp[#sp + 1] = table.concat(cache)
    ...
end
```
turns into this
```lua
local concat = table.concat
for _, char in utf8_iter(chars) do
    ...
    sp[#sp + 1] = concat(cache)
    ...
end
```

You'd have to measure if it actually makes a difference in this case, but it can make a difference in hot loops.

dyphire:
> > I have studied your proposal and I did not understand the use of `itable_clear()`.
> 
> Instead of creating a new table `cache = {}` you can do `itable_clear(cache)` to avoid some allocations.
> 
> > As for using `table.concat` as a local variable, I guess it refers to the processing of search characters in Menu.lua?
> 
> 
> You'd have to measure if it actually makes a difference in this case, but it can make a difference in hot loops.

Thank you for your suggestion. It seems very reasonable. I push the relevant changes.

dyphire:
Using the following test code for benchmark performance testing, the result is `total time:134.00ms`
```lua
str = "第13集-争自尊大嘴充文豪 解危难湘玉设大局"
local begin = os.clock()
for i = 1, 50000 do
    char_conv(str)
end
print(string.format('total time:%.2fms\n', ((os.clock() - begin) * 1000)))
```

src/uosc/lib/text.lua
```diff
@@ -479,6 +479,7 @@ end
 do
 	local word_separators = {
 		' ', '　', '\t', '-', '–', '_', ',', '.', '+', '&', '(', ')', '[', ']', '{', '}', '<', '>', '/', '\\',
+		'（', '）', '【', '】', '；', '：', '《', '》', '“', '”', '‘', '’', '？', '！',
```
tomasklaen, line 482: I think we should make this a set at this point:

```lua
local word_separators = create_set({' ', ...})
```

And use like:

```lua
if word_separators[char] then
```

src/uosc/main.lua
```diff
@@ -130,6 +130,9 @@ if options.autoload then mp.commandv('set', 'keep-open-pause', 'no') end
 local intl = require('lib/intl')
 t = intl.t
 
+--[[ CHARACTER CONVERSION ]]
+local char_conv = require('lib/char_conv')
```
tomasklaen, line 134: This shouldn't be `local`. It's accessible in `Menu.lua` because you're declaring `char_conv()` globally inside `char_conv.lua`, so there's no point for the assignment here.

Just do this:

```lua
--[[ INTERNATIONALIZATION ]]
local intl = require('lib/intl')
t = intl.t
require('lib/char_conv')
```

And inside `char_conv.lua`, add `local` in front of every variable and function that doesn't need to be global.

src/uosc/elements/Menu.lua
```diff
@@ -770,6 +770,7 @@ end
 ---@return MenuStackItem[]
 function search_items(items, query, recursive, prefix)
 	local result = {}
+	local concat = table.concat
```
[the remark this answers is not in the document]
po5: As pointed out by christoph, accessing local variables is faster.

src/uosc/lib/char_conv.lua
```diff
@@ -0,0 +1,83 @@
...
+	end
+end
+getPyTable()
+
+function need_romanization()
+	if next(pyTable) ~= nil then
+		return true
+	end
+	return false
+end
```
tomasklaen, line 32: This could just be:

```lua
function need_romanization()
	return next(pyTable) ~= nil
end
```

src/uosc/elements/Menu.lua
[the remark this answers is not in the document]
tomasklaen: Oh, missed that :) No issue if there's a reason behind it. I just thought it was a debugging leftover or something.

dyphire:
I have pushed some changes, please let me know your thoughts.

dyphire:
The upstream commit https://github.com/mpv-player/mpv/commit/33e922eabe4c5227afaf93022aefada63bd21b78 of mpv has implemented switching IME for input. But as @christoph-heinrich  mentioned earlier, switching to non English IME will break the key binding. This means that romanized input processing is still very meaningful, no require switching keyboard layout so as not to break key binding.

PR #814, opened 2023-12-04
main.lua: Support overriding `ziggy_path` via environment variable

nbraud:
This is useful when running on a platform builds of `ziggy` aren't provided for... or when using a distro-provided build.

The idea for this patch originated in NixOS/nixpkgs#270962, as I needed a way to point `uosc` at the nix-managed build of `ziggy`.


tomasklaen:
`MPV_UOSC_ZIGGY` is fine. And ziggy is a name of the multitool from Quantum Leap :)

ziggy

PR #816, opened 2023-12-08
feat!: `features` option to enable performance impacting non-essential features

tomasklaen:
Replaces option `text_width_estimation`, which is now one of its available flags:

```
features=accurate_text_width
```

Currently available flags:

```
accurate_text_width
accurate_sorting
```

Are there better names for these? Maybe just dropping the `accurate_` prefixes, since the "this an improved version of X" is kind of implied already.


christoph-heinrich:
`features` doesn't seem right for that, but I can't think of anything better.
It's not enabling additional features, but instead provides the option for alternative better but slower implementations.

`improved_variant=text_width` or something like that maybe?

hooke007:
`hr`.

po5:
`refine=text_width,sorting`

dyphire:
`improve_experimental=text_width,sorting`
Using experimental descriptions implies that this is not a flat out improvement.

PR #822, opened 2023-12-26
feat: add items to playlist from files menu when holding ctrl

### LINKED ISSUE #821: Menu to append files to playlist
how to create Open file in queue/playlist from Context Menu?
https://github.com/tomasklaen/uosc/blob/b36cefed888c1cd2d4e0b667cc177d781c2af987/README.md?plain=1#L428

christoph-heinrich:
Closes #821

### DIFF
--- a/src/uosc/lib/menus.lua
+++ b/src/uosc/lib/menus.lua
@@ -190,7 +190,7 @@ end
 
 -- Opens a file navigation menu with items inside `directory_path`.
 ---@param directory_path string
----@param handle_select fun(path: string): nil
+---@param handle_select fun(path: string, mods: Modifiers): nil
 ---@param opts NavigationMenuOptions
 function open_file_navigation_menu(directory_path, handle_select, opts)
 	directory = serialize_path(normalize_path(directory_path))
@@ -273,7 +273,7 @@ function open_file_navigation_menu(directory_path, handle_select, opts)
 			return
 		end
 
-		if info.is_dir and not meta.modifiers.alt then
+		if info.is_dir and not meta.modifiers.alt and not meta.modifiers.ctrl then
 			--  Preselect directory we are coming from
 			if is_to_parent then
 				inheritable_options.selected_path = directory.path
@@ -281,7 +281,7 @@ function open_file_navigation_menu(directory_path, handle_select, opts)
 
 			open_file_navigation_menu(path, handle_select, inheritable_options)
 		else
-			handle_select(path)
+			handle_select(path, meta.modifiers)
 		end
 	end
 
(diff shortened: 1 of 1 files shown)


hooke007:
> You can use ctrl+enter or ctrl+click to load the whole directory in mpv instead of navigating its contents.

Ctrl has been used. Could the dir be append in playlist in uosc? I never tried that.

christoph-heinrich:
Looks like the readme wasn't updated in https://github.com/tomasklaen/uosc/commit/6dfef3d3f8e7d17b107d64eb4e442ad2fef83860

tomasklaen:
I think shift would be a better modifier here. And other menus could probably use it for more appropriate things as well. It was a mistake to hardcode shift handling into Menu :/

So, we need to remove the click/enter shift handling from Menu, and leaving it for individual menu openers (they can set their menu as keep_open, and close manually when shift is not present), which on our side would be only the main menu.

hooke007:
> So, we need to remove the click/enter shift handling from Menu, and leaving it for individual menu openers (they can set their menu as keep_open, and close manually when shift is not present), which on our side would be only the main menu.

It is complex. IMO, `Shift` is useful with track menu, especial when I use mpv as music a player.

tomasklaen:
True, track menus could use shift too. Still, it shouldn't be hardcoded into Menu.

The biggest issue I see at the moment is that 3rd party scripts have no access to the modifiers yet, so they can't implement manual closing like this. Are there any scripts that heavily rely on current shift behavior?

christoph-heinrich:
I actually think shift as it behaves right now might be useful in addition to the ctrl introduced here.

tomasklaen:
Oh wait, I assumed the file menu already won't close when adding to playlist. I wouldn't want adding to playlist shortcut to close the menu.

christoph-heinrich:
Sorry for the delay on this.

The menu now stays open when appending to the playlist.
There is no feedback to the user that anything happened, which makes it not so great to use. Maybe we should flash the selected item when the menu stays open upon selection, but that is a more general problem and doesn't have to be part of this PR.

tomasklaen:
The more I think about it, the more I want this to be on the shift modifier. It makes way more sense there, and current shift behavior is pointless here. In main menu, shift activates without closing, and this would add to playlist without closing. Fits way better imo.

But I realize that will be a bit more work as we need to move shift handling from menu into individual menu creators.

Or I can just merge this as is, and move it to shift some time later. Up to you.

christoph-heinrich:
I expect to be pretty busy over the next month or so, so I think merging as is and adjusting later is the way to go.

tomasklaen:
All right. But swap the alt<->ctrl modifiers so that directory activation remains on ctrl.

christoph-heinrich:
You want me to undo https://github.com/tomasklaen/uosc/commit/6dfef3d3f8e7d17b107d64eb4e442ad2fef83860?

tomasklaen:
Oh, forgot about that :)

PR #831, opened 2024-01-11
fix: click threshold not working when any clickable element visible

tomasklaen:
This moves click threshold handling to cursor zones, and reworks them a bit. Each zone can now set `window_dragging = false` in their hitbox (defaults to `true` when `nil`), which will disable window dragging only when cursor is on top of that zone.

Resolves #820

---

There's a weird issue with this. If I click the close window button with this PR, the window closes, but the process doesn't actually exit. I have to Ctrl+C it in the console 😶. This doesn't happen on main. Anyone any ideas? It's not click threshold timer, as it happens even with it disabled.

### DIFF
--- a/src/uosc/lib/cursor.lua
+++ b/src/uosc/lib/cursor.lua
@@ -25,6 +25,7 @@ local cursor = {
 		mbtn_right = 0,
 		wheel = 0,
 	},
+	is_dragging_prevented = false,
 	event_forward_map = {
 		primary_down = 'MBTN_LEFT',
 		primary_up = 'MBTN_LEFT',
@@ -43,6 +44,7 @@ local cursor = {
 		wheel_down = 'wheel',
 		wheel_up = 'wheel',
 	},
+	window_dragging_blockers = create_set({'primary_click', 'primary_down'}),
 	event_propagation_blockers = {
 		primary_down = 'primary_click',
 		primary_click = 'primary_down',
@@ -98,6 +100,7 @@ end
 -- - One event type per zone: only the last bound zone per event type gets triggered.
 -- - In current implementation, you can only use `_click` or `_down`. If you bind both, only the last bound one will fire.
 -- - `primary_up` and `primary_click` prevent autohide when bound.
+-- - Primary `_down` and `_click` automatically disable dragging. Define `window_drag = true` on hitbox to re-enable.
 -- - `move` even zones are not triggered due to it being a high frequency event that is currently not needed as a zone.
 ---@param event string
 ---@param hitbox Hitbox
@@ -205,6 +208,7 @@ end
 -- Enables or disables keybinding groups based on what event listeners are bound.
 function cursor:decide_keybinds()
 	local new_levels = {mbtn_left = 0, mbtn_right = 0, wheel = 0}
+	self.is_dragging_prevented = false
 
 	-- Check global events.
 	for name, handlers in ipairs(self.handlers) do
@@ -218,7 +222,12 @@ function cursor:decide_keybinds()
 	for _, zone in ipairs(self.zones) do
 		local binding = self.event_binding_map[zone.event]
 		if binding and cursor:collides_with(zone.hitbox) then
-			new_levels[binding] = math.max(new_levels[binding], zone.hitbox.window_drag == false and 2 or 1)
+			local new_level = (self.window_dragging_blockers[zone.event] and zone.hitbox.window_drag ~= true) and 2
+				or math.max(new_levels[binding], zone.hitbox.window_drag == false and 2 or 1)
+			new_levels[binding] = new_level
+			if new_level > 1 then
+				self.is_dragging_prevented = true
+			end
 		end
 	end
 
@@ -326,7 +335,9 @@ end
 function cursor:leave() self:move(math.huge, math.huge) end
 
 function cursor:is_autohide_allowed()
-	return options.autohide and (not self.autohide_fs_only or state.fullscreen) and not Menu:is_open()
+	return options.autohide and (not self.autohide_fs_only or state.fullscreen)
+		and not self.is_dragging_prevented
+		and not Menu:is_open()
 end
 mp.observe_property('cursor-autohide-fs-only', 'bool', function(_, val) cursor.autohide_fs_only = val end)
 
--- a/src/uosc/lib/cursor.lua
+++ b/src/uosc/lib/cursor.lua
@@ -3,19 +3,9 @@ local cursor = {
 	y = math.huge,
 	hidden = true,
 	hover_raw = false,
-	-- Event handlers that are only fired on cursor, bound during render loop. Guidelines:
-	-- - element activations (clicks) go to `primary_down` handler
-	-- - `primary_up` is only for clearing dragging/swiping, and prevents autohide when bound
-	---@type {[string]: {hitbox: Hitbox; handler: fun(...)}[]}
-	zones = {
-		primary_down = {},
-		primary_up = {},
-		secondary_down = {},
-		secondary_up = {},
-		wheel_down = {},
-		wheel_up = {},
-		move = {},
-	},
+	-- Event handlers that are only fired on zones defined during render loop.
+	---@type {event: string, hitbox: Hitbox; handler: fun(...)}[]
+	zones = {},
 	handlers = {
 		primary_down = {},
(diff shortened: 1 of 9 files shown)


christoph-heinrich:
I'll have a look on sunday, but for me the process still exits when I press the close button.

src/uosc/lib/cursor.lua
```diff
@@ -27,12 +27,22 @@ local cursor = {
...
+		mbtn_right = 0,
+		wheel = 0,
+	},
+	event_forward_map = {
+		primary_down = 'MBTN_LEFT',
+		primary_up = 'MBTN_LEFT',
+		secondary_down = 'MBTN_RIGHT',
+		secondary_up = 'MBTN_RIGHT',
+		wheel_down = 'WHEEL',
+		wheel_up = 'WHEEL',
```
christoph-heinrich, line 44: Shouldn't those be `WHEEL_DOWN`/`WHEEL_UP`?

christoph-heinrich:
Not too keen on having `window_drag` directly in Rect/Circle, but otherwise the hitbox would have to be a table around the shape which also ain't great, so I guess it should stay as is.

I still have no explanation why the process doesn't quit for you with the exit button. Seems to me it should behave identically towards mpv as before except when click_threshold is used, but even then I don't see how that could happen.

Would be great if other people could test this too, maybe we can find a pattern for when this happens.

Sneakpeakcss:
> There's a weird issue with this. If I click the close window button with this PR, the window closes, but the process doesn't actually exit. I have to Ctrl+C it in the console 😶. This doesn't happen on main. Anyone any ideas? It's not click threshold timer, as it happens even with it disabled.

Even worse when `mpv` isn't launched through console and the processes just stack in the void without any notice.

> Would be great if other people could test this too, maybe we can find a pattern for when this happens.

Delaying this: https://github.com/tomasklaen/uosc/blob/0f970b5d8e4b819b9ed22cfab09e20cb49b6f9ee/src/uosc/elements/TopBar.lua#L20

``` diff
- mp.command(type(self.command) == 'function' and self.command() or self.command)
+ mp.add_timeout(.1, function() mp.command(type(self.command) == 'function' and self.command() or self.command) end)
```

seems to 'fix' it on my end (WIN10 22H2). However, the required value is a bit random. Initially `.05` was enough, but then it stopped working, so… _Just mpv things ?_

justmpvthings

___

Also as a btw, in current state this PR breaks changing volume with scroll:

<sub>1)</sub>
```
Wheel_Up no-osd add volume  5
Wheel_Down no-osd add volume -5
```
With this, the volume only changes when no uosc elements are shown, so potentially with something like `timeline_persistency=paused` it stops working at all. 

Then there's this scenario:

<sub>2)</sub>
```
Wheel_Up no-osd add volume  5; script-binding uosc/flash-volume
Wheel_Down no-osd add volume -5; script-binding uosc/flash-volume
```
Where it only changes volume one step at a time until the element dissapears again.

Changing the mentioned `WHEEL_DOWN`/`WHEEL_UP` only partially fixes this. In case of `1` it now works when elements are shown, but only for `WHEEL_DOWN`, and in case of `2` it's similar, but makes `WHEEL_UP` even worse.

Splitting this:

```diff
- wheel = math.max(self:get_binding_level('wheel_down'), self:get_binding_level('wheel_up')),
+ wheel_down = math.max(self:get_binding_level('wheel_down')),
+ wheel_up = math.max(self:get_binding_level('wheel_up')),
```

seems to make it work properly, but i have no idea if this is a proper change. I’m just passing by because this PR looked like something that would allow seeking by holding `LMB` over the progress bar without moving the window, the same as when using `window-dragging=no`. So, this whole thing is way beyond me.

tomasklaen:
The mpv not exiting is something with flags passed to `mp.enable/disable()` methods. If I don't pass any flags to any binding group, it works fine. But in the time I've had, I wasn't able to figure out what exactly is the cause or how to work around it :(

---

I can confirm that wheel event forwarding is borked even when adding `_DOWN/_UP` to forward names. Haven't looked more into it yet though.

christoph-heinrich:
> The mpv not exiting is something with flags passed to `mp.enable/disable()` methods. If I don't pass any flags to any binding group, it works fine. But in the time I've had, I wasn't able to figure out what exactly is the cause or how to work around it :(

Might be a windows specific thing, since you both are on windows and `[vo/gpu/win32] uninit` doesn't happen. Maybe a deadlock in platform specific code? You should open an issue with mpv about that. We'd still have to find a workaround to avoid bumping the requirement to the next mpv release, but otherwise it'll never get fixed.

tomasklaen:
Reported here: https://github.com/mpv-player/mpv/issues/13327

By isolating the issue I realized that this only happens when we close the app inside a `down` event. So the obvious workaround is to move button click callbacks to `up` events, which is something I wanted to do sometime anyway.

---

We still have the wheel events forwarding not working issue. @christoph-heinrich, you added the forwarding code, I just naively adapted it for other events and it doesn't seem to work right for wheel, could you glance at it when you have time if there's some obvious fix?

christoph-heinrich:
Here is a fix for the wheel forwarding.
Also a fix for ignoring double clicks only when on top of elements, otherwise fullscreening while some element is visible doesn't work.
input.conf now triggers commands on button up for mouse buttons like is the case by default (it's wrong on main, oops).

```diff
diff --git a/src/uosc/lib/cursor.lua b/src/uosc/lib/cursor.lua
index 66b9ce1..7beb60c 100644
--- a/src/uosc/lib/cursor.lua
+++ b/src/uosc/lib/cursor.lua
@@ -40,8 +40,8 @@ local cursor = {
 		primary_up = 'MBTN_LEFT',
 		secondary_down = 'MBTN_RIGHT',
 		secondary_up = 'MBTN_RIGHT',
-		wheel_down = 'WHEEL',
-		wheel_up = 'WHEEL',
+		wheel_down = 'WHEEL_DOWN',
+		wheel_up = 'WHEEL_UP',
 	}
 }
 
@@ -128,16 +128,17 @@ function cursor:trigger(event, ...)
 			-- Forward events if there was no handler.
 			local active = find_active_keybindings(forward_name)
 			if active then
+				local is_wheel = event:find('wheel', 1, true)
 				local is_up = event:sub(-3) == '_up'
 				if active.owner then
 					-- Binding belongs to other script, so make it look like regular key event.
 					-- Mouse bindings are simple, other keys would require repeat and pressed handling,
 					-- which can't be done with mp.set_key_bindings(), but is possible with mp.add_key_binding().
-					local state = is_up and 'um' or 'dm'
+					local state = is_wheel and 'pm' or is_up and 'um' or 'dm'
 					local name = active.cmd:sub(active.cmd:find('/') + 1, -1)
 					mp.commandv('script-message-to', active.owner, 'key-binding', name, state, forward_name)
-				elseif not is_up then
-					-- input.conf binding
+				elseif is_wheel or is_up then
+					-- input.conf binding, react to button release for mouse buttons
 					mp.command(active.cmd)
 				end
 			end
@@ -165,7 +166,8 @@ function cursor:decide_keybinds()
 	local mbtn_left_level = math.max(self:get_binding_level('primary_down'), self:get_binding_level('primary_up'))
 	local new_levels = {
 		mbtn_left = mbtn_left_level,
-		mbtn_left_dbl = mbtn_left_level,
+		-- window dragging only gets prevented when on top of an element, which is when double clicks should be ignored
+		mbtn_left_dbl = mbtn_left_level == 2 and 2 or 0,
 		mbtn_right = math.max(self:get_binding_level('secondary_down'), self:get_binding_level('secondary_up')),
 		wheel = math.max(self:get_binding_level('wheel_down'), self:get_binding_level('wheel_up')),
 	}
```

tomasklaen:
So this turned into a slight even handling refactor that adds `primary_click` and `secondary_click` events simulation.

All I could think of to try seems to work fine now, but it'd be nice if someone could test their use cases too.

christoph-heinrich:
You want elements to autohide despite being hovered over? I'd prefer if they don't do that.

christoph-heinrich:
Window dragging doesn't get disabled for buttons.

tomasklaen:
Both should be fixed now.

christoph-heinrich:
Looks good so far. I'll keep using it to see if I notice anything.

christoph-heinrich (approved):

PR #871, opened 2024-03-09
fix: don't trim too much when speed < 1 with playtime-remaining

christoph-heinrich:
Speed < 1 can require more parts of the formatted time then when speed >= 1.
E.g. when a video is almost 1min long and the speed is reduced to <1 then the minutes part of the remaining time would still get trimmed, even though it's not `00:`.

Alternatively the max seconds could also be
```lua
local max_seconds = math.max(state.duration, (state.duration - state.time) / speed)
```
so that the time only gets longer when the remaining time is large enough to require less trimming.

Really not sure what the best way of handling this is...

### DIFF
--- a/src/uosc/main.lua
+++ b/src/uosc/main.lua
@@ -435,7 +435,8 @@ function update_human_times()
 		if state.duration then
 			local speed = state.speed or 1
 			if options.destination_time == 'playtime-remaining' then
-				state.destination_time_human = format_time((state.time - state.duration) / speed, state.duration)
+				local max_seconds = speed >= 1 and state.duration or state.duration / speed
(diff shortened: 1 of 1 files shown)


tomasklaen:
I'd be fine with this. But shouldn't the elapsed time be handled the same way? So that its length format always matches the destination.

hooke007:
> But shouldn't the elapsed time be handled the same way

Where we display this data? theoretically, it's the `duration we had played`. So it shouldn't be scaled with speed.

christoph-heinrich:
> Where we display this data? theoretically, it's the `duration we had played`. So it shouldn't be scaled with speed.

It's not about scaling the elapsed time with speed, but about showing the same number of digits for elapsed time and destination/remaining time.

hooke007:
Got it.

PR #894, opened 2024-04-30
feat: support aligning window controls to the left

### LINKED ISSUE #892: Follow theme's preference for the window controls location (left vs right)
Hi all,

I started using `uosc` and absolutely love it so far but I always get tripped-up by windows controls. I use Linux and all my window control buttons (minimize/max/close) are on the top left side of the window. `uosc` has them on the top right at all times. Default `mpv` OSD detects the preference automatically and places it on the left or right. I have `border=no` set.

Any way to move them to the left? 

Thanks so much!

christoph-heinrich:
I don't like the way left alignment looks...

The calculation of `self.ax` isn't necessary anymore with the way clicks are handled now and afaik there is no benefit in accurately adjusting the boundary box, so for the sake of simplicity I've removed that instead of making it work with left alignment.

Closes #892

### DIFF
--- a/src/uosc.conf
+++ b/src/uosc.conf
@@ -109,7 +109,8 @@ menu_type_to_search=yes
 # Can be: never, no-border, always
 top_bar=no-border
 top_bar_size=40
-top_bar_controls=yes
+# Can be: `no` (hide), left or right
+top_bar_controls=right
 # Can be: `no` (hide), `yes` (inherit title from mpv.conf), or a custom template string
 top_bar_title=yes
 # Template string to enable alternative top bar title. If alt title matches main title,
--- a/src/uosc/elements/TopBar.lua
+++ b/src/uosc/elements/TopBar.lua
@@ -62,24 +62,22 @@ function TopBar:init()
 		return state.fullormaxed and 'set fullscreen no;set window-maximized no' or 'set window-maximized yes'
 	end
 
-	-- Order aligns from right to left
-	self.buttons = {
-		TopBarButton:new('tb_close', {
-			icon = 'close', background = '2311e8', command = 'quit', render_order = self.render_order,
-		}),
-		TopBarButton:new('tb_max', {
-			icon = 'crop_square',
-			background = '222222',
-			command = get_maximized_command,
-			render_order = self.render_order,
-		}),
-		TopBarButton:new('tb_min', {
-			icon = 'minimize',
-			background = '222222',
-			command = 'cycle window-minimized',
-			render_order = self.render_order,
-		}),
-	}
+	local close = TopBarButton:new('tb_close', {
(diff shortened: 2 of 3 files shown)


tomasklaen:
This looks bad. If we want left alignment here, we can't just slap the controls to the left. Everything else needs to be realigned to the other side now. So playlist position needs to be in the right corner, and then main/alt/chapter titles. And I don't know if chapter `[title][time]` should be swapped to `[time][title]`. I guess I prefer it to stay `[title][time]` even when aligned to the right.

Also the option is now not backwards compatible, and everyone that has `yes` currently (pretty much everyone) will have the controls disabled. We need to at the very least alias `'yes'` to `'right'`.

tomasklaen:
Actually, the `└` characters will need to be flipped to the other side too. And then the time to the right would look out of place, so it needs to be on the left after all. So annoying. Not sure if it's worth it tbh :)

And we should decide on the convention for keyword that disables stuff, as currently we sometimes use `never`, sometimes `none`, and this adds `no`.

christoph-heinrich:
> Actually, the `└` characters will need to be flipped to the other side too. And then the time to the right would look out of place, so it needs to be on the left after all. So annoying. Not sure if it's worth it tbh :)

Yeah I'm also not so sure about the added complexity. I'm currently playing around to see if it can be done in a not too complex way. No idea if that will result in a push today or not.

> And we should decide on the convention for keyword that disables stuff, as currently we sometimes use `never`, sometimes `none`, and this adds `no`.

Using `no` in place of `none` works (`volume=no|left|right`), but the other way around is weird (`top_bar_title=none|yes|custom_template`)

christoph-heinrich:
image

tomasklaen:
The remaining time needs to be on the left :( And when we align to the left, the text alignment of all titles (except playlist position) needs to use ass align 6, not 4.

christoph-heinrich:
> And when we align to the left, the text alignment of all titles (except playlist position) needs to use ass align 6, not 4.

You think so? I left it at 4 because I thought the start of the title is more important then the end. For right-to-left languages it's the other way around, so in a way we should actually have text alignment and window controls alignment be two separate options, but then that option should also control alignment in the menu, at which point this thing is getting way out of hand for something I'll never use...

tomasklaen:
No, no additional alignment options. I just wanted for texts to align vertically on the side they are attached to again. And we _could_ check if the text overflows the rectangle and flip its alignment back to 4, but that's getting too annoying. All right, then leave it as is, and just swap the chapter title<>remaining time when on right and lets be done with this.

christoph-heinrich:
I gave the align 6 thing a shot. It doesn't line up all that well either way.

without align 6:
image
with align 6:
image

PR #902, opened 2024-05-05
menus.lua: tweaks and fixes the display of the track menu

dyphire:
mpv upstream commit https://github.com/mpv-player/mpv/commit/ab3b174 has introduced support for BCP 47 language tags,
 and now track lang has uppercase and lowercase content.

Avoid showing the selection status of secondary subtitle in the subtitle menu at the same time.

current behavior of mian:
image
apply changes:
image

### DIFF
--- a/src/uosc/lib/menus.lua
+++ b/src/uosc/lib/menus.lua
@@ -109,6 +109,7 @@ function create_select_tracklist_type_menu_opener(menu_title, track_type, track_
 			items[#items].separator = true
 		end
 
+		local track_prop_index = tonumber(mp.get_property(track_prop))
 		local first_item_index = #items + 1
 		local active_index = nil
 		local disabled_item = nil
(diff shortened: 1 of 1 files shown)


christoph-heinrich (changes requested):
Seems reasonable

src/uosc/lib/menus.lua
```diff
@@ -148,10 +148,10 @@ function create_select_tracklist_type_menu_opener(menu_title, track_type, track_
 					title = (track.title and track.title or t('Track %s', track.id)),
 					hint = table.concat(hint_values, ', '),
 					value = track.id,
-					active = track.selected,
+					active = track.selected and track.id == tonumber(mp.get_property(track_prop)),
```
christoph-heinrich, line 151: I'm somewhat bothered by `tonumber(mp.get_property(track_prop))` being done twice on each iteration of the loop.

Youtube videos can add a ton of subtitle tracks if not restricted to specific languages via the ytdl options, and I don't know if the difference would be noticeable (probably not tbh), but since it's easy to avoid we might as well :smile:

dyphire: Change to use local variables, but only do it in a loop anyway.

christoph-heinrich: It's still doing `tonumber(mp.get_property(track_prop))` on each loop iteration though.

dyphire: I see what you mean, done.

christoph-heinrich (approved):

PR #927, opened 2024-07-01
feat: show playlist files in the file navigation menu

christoph-heinrich:
So far the file navigation menu would only show media files and directories, but opening playlist files via that menu is a valid use case as well.

From a quick look at `demux_playlist.c` and a bit of testing, those seem to be the only playlist formats mpv supports.
mpv can also open `.txt` files as playlists, but they need to be provided via the `--playlist=<path>` argument and thus don't work here.

ref. https://github.com/tomasklaen/uosc/discussions/926


hooke007:
Playlist should be load by `loadlist` command instead. And the ext name is not necessary actually.

christoph-heinrich:
> Playlist should be load by loadlist command instead.

Yes that would be better.

> And the ext name is not necessary actually.

I'm not sure what you mean with that. The extensions are used for filtering which files show up in the navigation menu, so of course they are necessary.

hooke007:
ext name is not necessary for mpv.

PR #929, opened 2024-07-08
feat: use loadlist when opening playlists

christoph-heinrich:
Opening a playlist with loadfile can lead to problems like autoload.lua populating the playlist with other files that were in the same directory as the playlist file.

To avoid that the command loadlist can be used to avoid having mpv treat the file like any other, just to notice during demuxing that it is a playlist.

ref. https://github.com/tomasklaen/uosc/pull/927#issuecomment-2210135586

@hooke007 You won't like that this is based on the file extension, but I don't know how else we would go about doing it. We could read the beginning of the file and deciding based on that, but I'd rather not do that unless it's _very_ important to do that. I've checked if we could do something via mime-type (e.g. `file --mime-type` on linux), but m3u8 playlists simply show up as plain text, so we'd have to treat all text files as playlists.

### DIFF
--- a/src/uosc/lib/menus.lua
+++ b/src/uosc/lib/menus.lua
@@ -591,10 +591,11 @@ function open_open_file_menu()
 	menu = open_file_navigation_menu(
 		directory,
 		function(path, mods)
+			local command = has_any_extension(path, config.types.playlist) and 'loadlist' or 'loadfile'
 			if mods.ctrl then
-				mp.commandv('loadfile', path, 'append')
+				mp.commandv(command, path, 'append')
 			else
-				mp.commandv('loadfile', path)
+				mp.commandv(command, path)
 				Menu:close()
 			end
 		end,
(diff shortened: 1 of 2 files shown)


tomasklaen:
Too many lines :) Just do:

```lua
local command = has_any_extension(path, config.types.playlist) and 'loadlist' or 'loadfile'
if mods.ctrl then
	mp.commandv(command, path, 'append')
else
	mp.commandv(command, path)
	Menu:close()
end
```

christoph-heinrich:
That was way too obvious... :disappointed:

Want me to add you as co-author? would be kinda hilarious for such a simple change.

mrfragger:
uosc 5.4.0 When clicking on .m3u8 playlist and opening with mpv it shows just the titles
image

If using uosc to navigate to .m3u8 and open it it shows the stream info also. 
image

I hardly ever use iptv but if I do I'd just click on the m3u8 to open it so it's mostly a non-issue.

tomasklaen:
We display playlist item titles if available, otherwise we fallback to the item's path/url. We do not set or manage titles, that is done by mpv or whatever populated the playlist entries.

PR #947, opened 2024-08-12
feat: add char_conv option

bennyyip:
Currently char_conv doesn't work if the first language in `languages` is `en`. This PR adds an option to enable it.

It is disabled by default and **BREAKS** the old behavior. The old users of this feature need to add following line to their `usoc.conf`:
```
char_conv=yes
```

### DIFF
--- a/src/uosc.conf
+++ b/src/uosc.conf
@@ -234,3 +234,7 @@ languages=slang,en
 #   window_border, top_bar, timeline, controls, volume,
 #   idle_indicator, audio_indicator, buffering_indicator, pause_indicator
 disable_elements=
+
+# Convert non-latin characters to latin for searching
+# Character get converted only if it is from one of `languages`
+char_conv=no
--- a/src/uosc/lib/char_conv.lua
+++ b/src/uosc/lib/char_conv.lua
@@ -3,12 +3,9 @@ require('lib/text')
 local char_dir = mp.get_script_directory() .. '/char-conv/'
 local data = {}
 
-local languages = get_languages()
-for i = #languages, 1, -1 do
-	lang = languages[i]
-	if (lang == 'en') then
-		data = {}
-	else
+if options.char_conv then
+	local languages = get_languages()
+	for _, lang in ipairs(languages) do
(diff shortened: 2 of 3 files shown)


dyphire:
The char_conv is designed to be used only for localized language romanization, and when the languages option's language priority is headed by en it means that the user is an English-speaking user who doesn't need localized language processing. Not enforcing char_conv is expected, and there is no reason to add an option to force it and break the usefulness of the languages option.

bennyyip:
> it means that the user is an English-speaking user who doesn't need localized language processing

No. I use English for UI but I still prefer to search chapter/playlist without switching IME. It doesn't make sense to skip romanization when the user is using English.

> all you need is to apply a patch like the one below

Adding a new option is better than applying a extra patch.

> instead of adding an option to break the existing behavior.

Yes, it's bad to break existing behavior. Maybe we can detect the `languages` options to decide whether it's enable to keep backward compatibility. But it's more complicated than my current implementation. 

I doubt only few users know this feature because it's not documented. It should not be big problem to break an undocumented behavior.

dyphire:
I see what you need now, but this can be accomplished by simply applying the following patch without adding the option, and it won't cause any damage.
This was a consideration I was missing when I added the Romanization feature, thanks for pointing this out.
```diff
---
 src/uosc/lib/char_conv.lua | 6 +-----
 1 files changed, 1 insertion(+), 5 deletions(-)

diff --git a/src/uosc/lib/char_conv.lua b/src/uosc/lib/char_conv.lua
index 8e5183c..101a629 100644
--- a/src/uosc/lib/char_conv.lua
+++ b/src/uosc/lib/char_conv.lua
@@ -6,11 +6,7 @@ local data = {}
 local languages = get_languages()
 for i = #languages, 1, -1 do
 	lang = languages[i]
-	if (lang == 'en') then
-		data = {}
-	else
-		table_assign(data, get_locale_from_json(char_dir .. lang:lower() .. '.json'))
-	end
+	table_assign(data, get_locale_from_json(char_dir .. lang:lower() .. '.json'))
 end
 
 local romanization = {}
-- 
```

dyphire:
> Could you make a new PR with this patch?

#948

PR #948, opened 2024-08-12
fix: always reads the romanization file for the corresponding language in the `language` option

dyphire:
When I originally added this feature, I overlooked that this was actually different from UI localization. It has different usage scenarios: UI main language and romanization files can be separated.

See-also #947


src/uosc/lib/char_conv.lua
bennyyip, line 1: There is no need to iterate the array backwards. 
```lua
for _, lang in ipairs(languages) do
```

dyphire: sure

PR #956, opened 2024-08-30
fix: implement localization correctly and update simplified chinese translation

dyphire:


tomasklaen:
Thx for the fixes, but the line 87 (`t('%s to delete', 'del, ctrl+del')`) is actually correct, so revert that one.

dyphire:
> Thx for the fixes, but the line 87 (`t('%s to delete', 'del, ctrl+del')`) is actually correct, so revert that one.

This is incorrect, it has multiple quotes making localization impossible.

dyphire:
> Thx for the fixes, but the line 87 (`t('%s to delete', 'del, ctrl+del')`) is actually correct, so revert that one.

Done.

PR #957, opened 2024-08-30
Fluent Icons

natural-harmonia-gropius:
As icons are used more and more, maintenance becomes difficult, and we need to re-examine how to change icons.


natural-harmonia-gropius:
Sorry, that I want to make a pr to my own fork.

PR #966, opened 2024-09-01
feat: add more actions to track menu and updated simplified chinese translation

dyphire:


tomasklaen:
This will need some more tweaking, and refactoring of the code above since I'd want the shortcuts to be f5 and del, not the current modifiers of enter, and current code doesn't allow that. I also see a bug in my code. Gimme a sec.

tomasklaen:
All right. Remove the modifier code from `handle_activate` function for these two actions.

Split the remove and reload code into `remove()` and `reload()` functions.

Create a new:

```lua
---@param event MenuEventKey
---@param close fun()
local function handle_key(event, close)
	if event.id == 'f5' then
		reload()
	elseif event.id == 'del' then
		remove()
	end
end
```

and add it to `on_key` prop below.

dyphire:
Since `on_key` doesn't pass the `event.value`, I can't get it to work properly.

tomasklaen:
Eh, there's lots more things to refactor here, but I won't torment you with these back and forths further. I'll take care of it. Thx for PR!

Issue #964, opened 2024-09-01
Option to hide Activate as secondary subs button


Solarunit:
Option to hide Activate as secondary subs button in Subtitles menu.
I do not plan to use secondary subs but I am forced to see this button every time now.

tomasklaen:
There are valid criticisms with current side buttons, like up & down in playlist are probably a bit too much, and actions shouldn't dangle outside menus when it's not necessary (hints/icon not that important to not be covered, like playlist), and I'm about to address that.

But your issues with footnote and secondary sub track button are way too nitpicky. These are both valuable features that don't affect the usability at all.

> forced to see this button every time now

That's a bit melodramatic, isn't it? You can't expect every gui to have an option for each of its buttons to remove it when you don't plan to use it.

dyphire:
> like up & down in playlist are probably a bit too much

I think up & down in playlist are useful for pure mouse and touch operations and should not be removed.

tomasklaen:
They are only useful if you want to move one item something like 1-2 positions away. Any other case is incredibly annoying to deal with, and would be more suited for drag and drop UX, which is something I'm not looking forward to implementing here, if ever...

I might re-add them back though if more people would want that.

Sneakpeakcss:
> They are only useful if you want to move one item something like 1-2 positions away. Any other case is incredibly annoying to deal with, and would be more suited for drag and drop UX, which is something I'm not looking forward to implementing here, if ever...
> 
> I might re-add them back though if more people would want that.

I don't see myself using this, but there's also the "select the button with right arrow and hold enter" use-case. A bit more convenient for moving entries without using both hands for modifier keys, especially when all the scrolling keys are in the same keyboard area.

However, it's a bit wonky because any mouse movement hides the buttons, and also resets the selection to the beginning of the playlist even when cursor is outside of the menu area.

Solarunit:
>That's a bit melodramatic, isn't it? You can't expect every gui to have an option for each of its buttons to remove it when you don't plan to use it.

Well, I think secondary sub track button looks cool but at the same time it makes menu looks less minimal like before, plus that hints/icon. 

IMO secondary subs is a very niche feature and could be disabled by default.

tomasklaen:
This kind of configuration granularity is a net negative because it introduces too much configuration fatigue for everyone trying to configure their players. Having to go through hundreds of options because there's one guy for every button who doesn't want to see it is not reasonable.

So sorry, but will have to be a wontfix.

[...]

PR #975, opened 2024-09-03
feat: optimize track title and update Chinese translation

dyphire:
ref #964

### DIFF
--- a/src/uosc/intl/zh-hans.json
+++ b/src/uosc/intl/zh-hans.json
@@ -41,6 +41,8 @@
 	"Loop file": "单个循环",
 	"Loop playlist": "列表循环",
 	"Menu": "菜单",
+	"Move down": "下移",
+	"Move up": "上移",
 	"Navigation": "导航",
 	"Next": "下一个",
 	"Next page": "下一页",
--- a/src/uosc/lib/menus.lua
+++ b/src/uosc/lib/menus.lua
@@ -215,6 +215,8 @@ function create_select_tracklist_type_menu_opener(opts)
 		end
 
 		local track_prop_index, snd_prop_index = get_props()
+		local filename = mp.get_property_native('filename/no-ext')
+		local escaped_filename = string.gsub(filename --[[@as string]], '[%(%)%.%+%-%*%?%[%]%^%$%%]', '%%%1')
 		local first_item_index = #items + 1
 		local active_index = nil
 		local disabled_item = nil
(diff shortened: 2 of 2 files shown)


po5:
You should reuse the existing Lua pattern escape code:
https://github.com/tomasklaen/uosc/blob/3621e0b4c6033f6ca2cc0f845cc9fc8209c574f3/src/uosc/elements/TopBar.lua#L68

dyphire:
> You should reuse the existing Lua pattern escape code:
> 
> https://github.com/tomasklaen/uosc/blob/3621e0b4c6033f6ca2cc0f845cc9fc8209c574f3/src/uosc/elements/TopBar.lua#L68

Missed it.

tomasklaen:
Better do a `nil` check on `track.title`.

But I don't know. Is this a good idea? I don't feel good about seeing items like this:

image

I could do a poor man's solution to the width calculation instead: ensure there's always space for at least 3 external buttons outside. That should cover 99% use cases.

dyphire:
> Better do a `nil` check on `track.title`.

I did at first but when it was an external track the track title was the filename this could not be empty.

> 
> But I don't know. Is this a good idea? I don't feel good about seeing items like this:

I improved it and it looks much better now.

> I could do a poor man's solution to the width calculation instead: ensure there's always space for at least 3 external buttons outside. That should cover 99% use cases.

I would say that's fine too, even so this PR is still a nice improvement.

tomasklaen:
```lua
if track.lang and track.title:lower() == track.lang:lower() then
	track.title = nil
end
```

What is this for? Track title in here will always contain extension, so it will never match lang, no?

dyphire:
> ```lua
> if track.lang and track.title:lower() == track.lang:lower() then
> 	track.title = nil
> end
> ```
> 
> What is this for? Track title in here will always contain extension, so it will never match lang, no?

No, the `track.title = track.title:gsub(escaped_filename .. '%.?', ''):gsub('%.([^%.]+)$', '')` has stripped the extension.

dyphire:
There was a small error that has been fixed.

Sneakpeakcss:
There's a crash when subtitle menu is opened when nothing is playing:

```
[uosc] 
[uosc] stack traceback:
[uosc] 	...mpvtest/scripts/uosc/lib/menus.lua:226: in function 'serializer'
[uosc] 	...mpvtest/scripts/uosc/lib/menus.lua:91: in function 'callback'
[uosc] 	...mpv-xmpvtest/scripts/uosc/main.lua:870: in function 'fn'
[uosc] 	mp.defaults:233: in function 'fn'
[uosc] 	mp.defaults:66: in function 'handler'
[uosc] 	mp.defaults:385: in function 'handler'
[uosc] 	mp.defaults:515: in function 'call_event_handlers'
[uosc] 	mp.defaults:557: in function 'dispatch_events'
[uosc] 	mp.defaults:508: in function <mp.defaults:507>
[uosc] 	[C]: at 0x7ff6d29fc020
[uosc] 	[C]: at 0x7ff6d29fae80
[uosc] Lua error: ...mpvtest/scripts/uosc/lib/menus.lua:226: bad argument #1 to 'gsub' (string expected, got nil)
```

tomasklaen:
Fixed. Thx for catching that.

PR #980, opened 2024-09-04
fix: updated Simplified Chinese translation

dyphire:


tomasklaen:
I'll keep this open until next version so that we don't have 10 of these per version in history :).

PR #1034, opened 2024-12-14
fix: replaces the deprecated wmic command

dyphire:
The wmic disabled by default since Win11 24H2
image


tomasklaen:
On Win10, the `fsutil fsinfo drives` command produces this output:

```
Drives: C:\ D:\ E:\ F:\ G:\
```

And with your serialization, this makes the first drive in the list be `s:`.

dyphire:
> On Win10, the `fsutil fsinfo drives` command produces this output:
> 
> ```
> Drives: C:\ D:\ E:\ F:\ G:\
> ```
> 
> And with your serialization, this makes the first drive in the list be `s:`.

I was using a non-English system and missed it. Fixed.

PR #1059, opened 2025-02-11
feat: adapting ime features for mpv

dyphire:
See-Also: https://github.com/mpv-player/mpv/pull/15832


po5:
`state.ime_active == false`?

dyphire:
> `state.ime_active == false`?

Makes sense, done.

PR #1094, opened 2025-04-30
feat: working text cursor when editing search inputs

### LINKED ISSUE #766: Supports moving the cursor when inputing in search menus
When using search menus, it should supports moving the cursor during input, achieving more flexible input operations. This is very useful for subtitle downloads.

Refer to the code already implemented in the console script.

tomasklaen:
This ads a working cursor when editing menu search input fields. I'd be surprised if there weren't some edge case issues I missed with all the different menu & search types we have, but as far as I can tell, at the moment there's only one small issue:

When there's a space in front of the cursor, our `text_width()` measuring misbehaves and overshoots the measurement of the width of the query tail (the part after the cursor), so the cursor position and the whole query jumps around, see below.

cursor_text_width

I've tried giving it a quick look but couldn't find an obvious fix for this. maybe @christoph-heinrich could look at it? Either way, it's still better than nothing and I wouldn't mind merging it as is.

closes #766


dyphire:
Overall this could work.
But besides the space issue you mentioned, it didn't handle CJK characters well in my tests.

https://github.com/user-attachments/assets/b31e80d1-5574-4f56-a3cc-31ae9fcbd4db

christoph-heinrich:
I wouldn't mind giving this a look and trying to fix that positioning and multi-byte character issue, but I have a more pressing issue to attend to currently. If you don't mind waiting a few days (up to ~1 week I guess), then I'll see what I can do :wink:

tomasklaen:
No worries :) I'll merge this as is and you can PR if you find some fixes when convenient.

PR #1102, opened 2025-05-15
fix: cursor in search inputs is not moved correctly by utf-8 characters

dyphire:
By the way, the `search_query_delete` function doesn't work properly.


tomasklaen:
Move the `utf8_next` and `utf8_prev` functions to `text.lua`.

PR #1109, opened 2025-05-19
fix: menu search cursor position when in front of a space

christoph-heinrich:
ass_escape() replaces leading spaces with '\h' (two characters), which led to the visible width of tail not lining up with the estimated one.

ref. https://github.com/tomasklaen/uosc/pull/1094


tomasklaen:
Thanks! :)

Issue #1090, opened 2025-04-21
Hide unused UI elements for pictures


Timberfang:
MPV with uosc is my favorite media viewer; I use it for video, audio, and images. When viewing still images, would it be possible to hide elements like subtitles, speed control, and volume? As far as I can tell, these elements don't do anything with still images. Of course, they're still useful for audio & video, though, so I wouldn't want to disable them entirely.

tomasklaen:
I made the volume slider hide on images.

But the controls were already configurable to do this with the disposition syntax (documented in the default `uosc.conf` file).

For example, changing `subtitles` to `<!image>subtitles` will only show the subtitles button if currently open file is not a single image. Same for speed: `<!image>speed`.

I added these dispositions to default controls config option. If you've edited yours, you'll need to tweak it accordingly.

Timberfang:
Sorry about that, I should have checked more thoroughly. Thank you for your help all the same!

PR #1113, opened 2025-05-20
fix: Broken speed UI

### LINKED ISSUE #1112: 5.9.0 made the speed UI disappear for me
There's nothing seemingly logged to console, and I used their respective default `.conf` files.
5.8.0

5.9.0

Removing `<video,audio>` from `uosc.conf` fixes the issue (https://github.com/tomasklaen/uosc/commit/d286bd02130c66ace1d5ea897c7f2d458a91614f). However, despite this being there, video, and audio files don't show me the speed UI. Also you can change the speed for animated image formats, Imo it's just better to keep the speed UI regardless. If hiding it is really wanted, a duration check (with maybe an exclusion for pipes, hls, and dash) would be much more apt.

brian6932:
fixes #1112

### DIFF
--- a/src/uosc.conf
+++ b/src/uosc.conf
@@ -80,7 +80,7 @@ progress_line_width=20
 #   fullscreen = cycle:crop_free:fullscreen:no/yes=fullscreen_exit!?Fullscreen
 #   loop-playlist = cycle:repeat:loop-playlist:no/inf!?Loop playlist
 #   toggle:{icon}:{prop} = cycle:{icon}:{prop}:no/yes!
-controls=menu,gap,<!image>subtitles,<has_many_audio>audio,<has_many_video>video,<has_many_edition>editions,<stream>stream-quality,gap,space,<video,audio>speed,space,shuffle,loop-playlist,loop-file,gap,prev,items,next,gap,fullscreen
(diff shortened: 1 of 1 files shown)


hooke007:
It's not a fix just a revert.
https://github.com/tomasklaen/uosc/issues/1090

brian6932:
@hooke007 it literally doesn't work.
Sure it's a revert, but it's currently broken.

hooke007:
>IMO, the debate over "preference" will never be exhausted.

It was talking about the issue 1090, not yours.

brian6932:
My bad then, wasn't really clear.

hooke007:
https://github.com/tomasklaen/uosc/blob/609dd978763c8045e98220c1c2aa47ccb7512f77/src/uosc/main.lua#L32

It looks like a typo in the uosc conf because scrpt use the different vaule.

brian6932:
Ah you're correct, I didn't know this.

tomasklaen:
Thx for PR, but the issue here was that comma in dispositions was acting as AND instead of OR. I also moved the subtitles to `<video,audio>`, as having subtitle button on an idle window makes no sense.

PR #1139, opened 2025-07-09
feat: use native clipboard API if available

layercak3:
This works better on wayland display servers when the user does not give permission to use data-control or it is unsupported by the display server, due to keyboard focus requirements. It also removes the need to build or install ziggy when not using its other functions.

If the property does not exist or is unavailable then ziggy is used.


layercak3:
It depends on implementation details of the clipboard backend. The wayland and wayland VO backends always return an empty string, but the win32 backend will return `nil` and `MPV_ERROR_PROPERTY_ERROR`.

If there is no clipboard backend enabled, or the VO backend is used and the VO is unavailable or the VO doesn't implement the `VOCTRL_GET_CLIPBOARD` control, then the error is `MPV_ERROR_PROPERTY_UNAVAILABLE`.

I've updated it to fallback to ziggy only if the error is `MPV_ERROR_PROPERTY_NOT_FOUND` or `MPV_ERROR_PROPERTY_UNAVAILABLE`.

tomasklaen:
Thx!

PR #1140, opened 2025-07-11
feat: custom disposition

ghost:
Use case: I have a script that shows Matroska file attachments with uosc menu, which tells me if the fonts used by embedded subtitles are embedded in the MKV file. I want the button that triggers the menu to only show up when the current playing file is an MKV file.

By the way, shouldn't this be `element:observe_mp_property` so that the handlers can be removed in `Controls:on_options`?

https://github.com/tomasklaen/uosc/blob/cfe1b2eea34668fcfc706ae467960a9f580e0902/src/uosc/elements/Controls.lua#L271

### DIFF
--- a/src/uosc.conf
+++ b/src/uosc.conf
@@ -63,7 +63,7 @@ progress_line_width=20
 #     - `has_chapter` - true for files with chapter list
 #     - `stream` - true if current file is read from a stream
 #     - `has_playlist` - true if current playlist has 2 or more items in it
-#     - `{mpv_prop}` - any mpv property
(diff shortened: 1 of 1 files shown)


tomasklaen:
Can you give me an example of an mp property that could be used like a disposition? Because most of the props are not booleans, but something like arrays, and need to be converted into a usable disposition with `#prop > 0`, so the usefulness of this is very limited.

I think a better solution would be for your script to use `set-button`, and for me to add support for managed buttons to define `hide: boolean` prop which would hide it from the controls bar.

Or allow scripts to send us custom dispositions.

---

And yeah, that should be `element:observe_mp_property()`, thanks for catching that :)

tomasklaen:
Thanks!

PR #1148, opened 2025-07-21
feat: fuzzy search based on fzy algorithm

dyphire:
The fzy.lua is token from https://github.com/mpv-player/mpv/blob/master/player/lua/fzy.lua.
Using the same fuzzy matching algorithm as the mpv internal script is a good choice.
Prioritize the results of exact matches. At the same time, the matching content has been highlighted.

Close #645, #1135, #1142

### DIFF
--- a/src/uosc.conf
+++ b/src/uosc.conf
@@ -152,7 +152,7 @@ text_border=1.2
 # Border radius of buttons, menus, and all other rectangles
 border_radius=4
 # A comma delimited list of color overrides in RGB HEX format. Defaults:
-# foreground=ffffff,foreground_text=000000,background=000000,background_text=ffffff,curtain=111111,success=a5e075,error=ff616e
+# foreground=ffffff,foreground_text=000000,background=000000,background_text=ffffff,curtain=111111,success=a5e075,error=ff616e,match=69c5ff
 color=
 # A comma delimited list of opacity overrides for various UI element backgrounds and shapes.
 # This does not affect any text, which is always rendered fully opaque. Defaults:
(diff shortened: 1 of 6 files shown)


tomasklaen:
Don't have too much time atm to examine this thoroughly (like are 3 loops in `search_items()` necessary?), but what I caught thus far:

- `highlight_match()` 2nd param types are wrong.
- Move stuff that is not strictly menu related like `highlight_match()` and `get_roman_match_positions()` to `text.lua`.
- You're assigning `ass_safe_title`, but you're not `ass_escape()`ing the text anywhere. I think it should happen in `highlight_match()`
- Set the default color to `69c5ff` and make the highlighted part bold.

dyphire:
> Don't have too much time atm to examine this thoroughly (like are 3 loops in `search_items()` necessary?), but what I caught thus far:
> 
> * `highlight_match()` 2nd param types are wrong.
> * Move stuff that is not strictly menu related like `highlight_match()` and `get_roman_match_positions()` to `text.lua`.
> * You're assigning `ass_safe_title`, but you're not `ass_escape()`ing the text anywhere. I think it should happen in `highlight_match()`
> * Set the default color to `69c5ff` and make the highlighted part bold.

Done.

Regarding the 3 loops, this is mainly for the exact match and the acronym matching results to be highlighted correctly. Putting the romanized match and hint match into the first loop will invalidate the highlight. No better idea about this for the time being.

tomasklaen:
I ran out of time today to examine this thoroughly, but it seems to work fine and I don't want to make you wait more so I'm just gonna merge it and hope for the best :)

Issue #886, opened 2024-04-22
Codec and resolution labels on the controls bar


Solarunit:
I think it would be nice to have codec and resolution labels that you can add on the controls bar.
Something like this:

![mpv-codec-info](https://github.com/tomasklaen/uosc/assets/20891064/24cf278e-907e-448c-9e32-515f11fe39c5)

tomasklaen:
Maybe we can have a new control element that displays a template while subscribing to its properties like we do with the title.

Solarunit:
I see It could be a problem with converting mpv resolution property like `3840x1920`  to just `4K`  or `1920x1080` to `1080 HD` to make it look more compact on controls bar.

tomasklaen:
The resolution shorthands could be resolved by providing a custom `{resolution}` template macro which we format behind the scenes.

This would also need some basic templating features like built in function for formatting stuff, e.g. `{human_time(duration)}`.

I expect this to take a lot of fiddling and time to implement. Can't promise I ever get to it.

mrfragger:
I realize this is for control bar and not titlebar but in the meantime this is what I use for titlebar to display video resolution. Some reason can't upload screenshot but looks like below.

64% 00:20:11 00:31:45 00:11:34 https://www.youtube.com/watch?=blahblah 441 kbps 47 kbps 360p

`top_bar_alt_title=${percent-pos}% ${playback-time} ${duration} ${playtime-remaining} ${?demuxer-via-network==yes:${path}}${?demuxer-via-network==no:${media-title}} ${?file-size:${file-size}} ${?video-bitrate:${video-bitrate}} ${?audio-bitrate:${audio-bitrate}} ${?video-params/h:${video-params/h}}p`

toggle it on/off with shortcut input.conf
`C             script-message-to uosc toggle-elements top_bar       #! [Playback] > On/off uosc top bar progress %`

Solarunit:
My idea is inspired by Kodi video info in the right bottom corner. So I want similar labels in controls bar :slightly_smiling_face:

Ammar-A7med:
> My idea is inspired by Kodi video info in the right bottom corner. So I want similar labels in controls bar 🙂

i see it good feature better than info shit

PR #1149, opened 2025-07-26
Pseudo labels

Deulfi:
**Improves button behavior based on icon presence.**  
- Disables background hover when the button lacks an icon.  
- Centers all iconless buttons—this is acceptable because we only want iconless buttons in two cases:  
  1. **Invisible buttons**, where centering is irrelevant (useful for hiding without layout changes).  
  2. **Pseudo labels**, where centering improves visual alignment.  

**Note:** Iconless buttons can only be created via the **button API**. Buttons generated through **uosc.conf** cannot be iconless (no changes for normal user).

Normal Buttons:
without_button_adjustment
With adjustments and hidden background:
with_button_adjustment

I tested it a few months and had no problems, even if the spacing must be done somewhat manually. But is not really a big issue as long as only short information is put into the "label".

### DIFF
--- a/src/uosc/elements/Button.lua
+++ b/src/uosc/elements/Button.lua
@@ -48,7 +48,7 @@ function Button:render()
 	if is_hover and is_clickable and background_opacity < 0.3 then background_opacity = 0.3 end
 
 	-- Background
-	if background_opacity > 0 then
+	if background_opacity > 0 and (self.icon ~= '') then
 		ass:rect(self.ax, self.ay, self.bx, self.by, {
 			color = (self.active or not is_hover) and background or foreground,
 			radius = state.radius,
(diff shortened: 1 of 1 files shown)


tomasklaen:
This co-opts button's label to be something it's not supposed to be and makes buttons into a weird chimeras with hidden behavior that makes them not-a-buttons when there's no icon. I really don't like this.

I do want to eventually add a support for string/meta elements in controls bar (#886) that will display stuff based on provided mpv template, but this is not a way to do it.

Deulfi:
Yeah, the slapdash nature was one reason I hesitated to open a pull request.

I assume invisible button is also unwanted (I thought iconless buttons are unintended anyway)? And yeah, I know, #886 was the reason I tried it in the first place.

Have a nice day =)

PR #1156, opened 2025-08-14
feat: add autohide_ignore_dragging option

7up4:
### Summary
This adds the autohide_ignore_dragging option, allowing the OSC to follow mpv’s cursor-autohide timing even if a UI element has disabled window dragging (e.g. when the cursor is over clickable controls).

### Motivation
On multi-monitor setups, especially when the player is displayed on a secondary screen, the system cursor can unexpectedly “jump” or stay positioned over the OSC — often directly on a button or in the corner.
In such cases, is_dragging_prevented is true, and the current logic prevents autohide, leaving the OSC permanently visible until the cursor is moved manually.

This option provides a way to override that safety check so OSC hides automatically regardless of dragging prevention.

### DIFF
--- a/src/uosc.conf
+++ b/src/uosc.conf
@@ -181,6 +181,12 @@ time_precision=0
 buffered_time_threshold=60
 # Hide UI when mpv autohides the cursor. Timing is controlled by `cursor-autohide` in `mpv.conf` (in milliseconds).
 autohide=no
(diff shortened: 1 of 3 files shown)


7up4:
Agreed. No problem. Just thought that someone might find this option useful. Thanks!

PR #1159, opened 2025-08-23
feat: display youtube heatmap on timeline

### LINKED ISSUE #832: Show youtube's replay heatmap data
For youtube videos display the "most replayed" heatmap:
replay

yt-dlp makes this available with:
`yt-dlp --print '%(heatmap)j' 'https://www.youtube.com/watch?v=BIO8elqCBOs'`

Sneakpeakcss:
All the interactions with other options that i found are handled, so at this point this is as much done as i’m capable of.

I know the goal is to keep the config minimal, but in the end i've chosen to include the mentioned positioning for users who use a low `timeline_size` value but still want to see the heatmap. In that case, the heatmap uses the default youtube height, which coincidentally matches uosc's default timeline `40` size, so that's nice.

The `timeline_heatmap` option could still be removed, but turning it off through opacity options felt a bit convoluted.

closes #832

### DIFF
--- a/src/uosc.conf
+++ b/src/uosc.conf
@@ -14,6 +14,8 @@ timeline_border=1
 timeline_step=5
 # Render cache indicators for streaming content
 timeline_cache=yes
+# Render heatmap for youtube videos. Available: overlay, above, no
+timeline_heatmap=overlay
 
 # When to display an always visible progress bar (minimized timeline). Can be: windowed, fullscreen, always, never
 # Can also be toggled on demand with `toggle-progress` command.
@@ -152,11 +154,11 @@ text_border=1.2
 # Border radius of buttons, menus, and all other rectangles
 border_radius=4
 # A comma delimited list of color overrides in RGB HEX format. Defaults:
-# foreground=ffffff,foreground_text=000000,background=000000,background_text=ffffff,curtain=111111,success=a5e075,error=ff616e,match=69c5ff
+# foreground=ffffff,foreground_text=000000,background=000000,background_text=ffffff,curtain=111111,success=a5e075,error=ff616e,match=69c5ff,heatmap=00adee
 color=
 # A comma delimited list of opacity overrides for various UI element backgrounds and shapes.
 # This does not affect any text, which is always rendered fully opaque. Defaults:
-# timeline=0.9,position=1,chapters=0.8,slider=0.9,slider_gauge=1,controls=0,speed=0.6,menu=1,submenu=0.4,border=1,title=1,tooltip=1,thumbnail=1,curtain=0.8,idle_indicator=0.8,audio_indicator=0.5,buffering_indicator=0.3,playlist_position=0.8
+# timeline=0.9,position=1,chapters=0.8,slider=0.9,slider_gauge=1,controls=0,speed=0.6,menu=1,submenu=0.4,border=1,title=1,tooltip=1,thumbnail=1,curtain=0.8,idle_indicator=0.8,audio_indicator=0.5,buffering_indicator=0.3,playlist_position=0.8,heatmap=0.4
 opacity=
 # A comma delimited list of features to refine at a cost of some performance impact.
 # text_width - Use a more accurate text width measurement that measures each text string individually
--- a/src/uosc/elements/Timeline.lua
+++ b/src/uosc/elements/Timeline.lua
@@ -18,12 +18,20 @@ function Timeline:init()
 	self.progress_line_width = 0
 	self.is_hovered = false
 	self.has_thumbnail = false
+	self.heatmap = nil
 
 	self:decide_progress_size()
 	self:update_dimensions()
 
-	-- Release any dragging when file gets unloaded
-	self:register_mp_event('end-file', function() self.pressed = false end)
+	-- Load Youtube heatmap data if available
+	self:register_mp_event('file-loaded', function()
+		self.heatmap = load_youtube_heatmap()
+	end)
+	-- Release any dragging and clear heatmap when file gets unloaded
+	self:register_mp_event('end-file', function()
+		self.pressed = false
+		self.heatmap = nil
+	end)
 end
 
 function Timeline:get_visibility()
@@ -257,7 +265,30 @@ function Timeline:render()
 	ass:draw_stop()
 
 	-- Progress
-	ass:rect(fax, fay, fbx, fby, {opacity = config.opacity.position})
+	local function draw_progress()
+		ass:rect(fax, fay, fbx, fby, {opacity = config.opacity.position})
+	end
+
+	-- Youtube heatmap
+	local function draw_heatmap()
+		if options.timeline_heatmap ~= 'no' and self.heatmap and config.opacity.heatmap > 0 and visibility > 0 then
+			local is_above = options.timeline_heatmap == 'above'
+			local height = math.min(40, size / self.size * 40)
+			local ax, ay = bax, is_above and (bay - height) or (bay + self.top_border)
+			local bx, by = bbx, is_above and bay or bby
+			local opts = {color = config.color.heatmap, opacity = config.opacity.heatmap * visibility}
+			ass:smooth_curve(ax, ay, bx, by, self.heatmap, opts)
+		end
+	end
+
+	-- Change draw order based on 'timeline_style' to keep the heatmap visible
+	if is_line then
+		draw_heatmap()
+		draw_progress()
+	else
+		draw_progress()
+		draw_heatmap()
+	end
 
 	-- Uncached ranges
 	if state.uncached_ranges then
--- a/src/uosc/lib/ass.lua
+++ b/src/uosc/lib/ass.lua
@@ -266,3 +266,47 @@ function ass_mt:spinner(x, y, size, opts)
 	self:icon(x, y, size, 'autorenew', opts)
 	request_render()
 end
+
+-- Renders a smooth curve from Bezier segments.
+---@param ax number
+---@param ay number
+---@param bx number
+---@param by number
+---@param points number[] Flat table of normalized points (0–1): start point followed by segment entries cp1x, cp1y, cp2x, cp2y, px, py, ...
+---@param opts? {color?: string; border?: number; border_color?: string; opacity?: number|{primary?: number; border?: number, shadow?: number, main?: number}; clip?: string}
+function ass_mt:smooth_curve(ax, ay, bx, by, points, opts)
+	if not points or #points < 8 then return end
+	opts = opts or {}
+	local border_size = opts.border or 0
+	local tags = '\\pos(0,0)\\rDefault\\an7\\blur0'
+	-- border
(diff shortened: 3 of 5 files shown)


tomasklaen:
Good work! :) A couple changes:

Config should just be a single:

```
# Render heatmap for youtube videos. Avilable: overlay, above, no
timeline_heatmap=overlay
```

And in code every value that is not `overlay`, or `above` is treated as heatmap disabled.

---

In `timeline.lua`, store all heatmap related cache in one `self._heatmap` object. So:

```lua
self._heatmap.last_x
self._heatmap.cached_midpoints
-- ... etc
```

---

I'm a bit worried about performance. In your testing, what is the max number of points you've seen?

Sneakpeakcss:
Done.

> I'm a bit worried about performance. In your testing, what is the max number of points you've seen?

From what i've seen, it's always `100` points regardless of video length. 

Here's few examples that i've used:

1 min: https://www.youtube.com/watch?v=7PJi9KSDVbQ
15 min: https://www.youtube.com/watch?v=wk9L1N9bRRE
10h: https://www.youtube.com/watch?v=SBuf4vzfr-M

In normal use, i didn't notice any performance issues, unless you treat the `mpv` window as a canvas and constantly move the mouse rapidly back and forth while timeline is visible. In that case (if we believe performance tab from `stats.lua`) there's a jump of a few percent in CPU usage (though this is probably CPU dependant).

Waving the cursor forces constant redraws, so while midpoints are cached, the control points in `ass:smooth_curve` aren't. And from what i've seen, nothing really gets cached(?), so waving around isn't ever a good idea. I guess drawing rectangles isn't that intensive in comparison.
Previously i tried caching them similarly to midpoints, but it turned into a global mess. 

The best concise option that i've managed to come up with is this:


```diff
diff --git a/timeline.lua b/timeline.lua
index 29d7c77..cabed90 100644
--- a/timeline.lua
+++ b/timeline.lua
@@ -293,6 +293,7 @@ function Timeline:render()
 			end
 			points[#points + 1] = {bbx, bby + offset}
 			self._heatmap.cached_midpoints = points
+            self._heatmap.bezier_cache = {}
 		end
 		local opts = {
 			color = config.color.heatmap, opacity = config.opacity.heatmap * visibility,
@@ -301,7 +302,7 @@ function Timeline:render()
 			local clip_overflow = string.format('%d,%d,%d,%d', bax, bay + self.top_border, bbx, bby)
 			opts.clip = '\\clip(' .. clip_overflow .. ')'
 		end
-		ass:smooth_curve(self._heatmap.cached_midpoints, opts)
+		ass:smooth_curve(self._heatmap.cached_midpoints, opts, self._heatmap.bezier_cache)
 	end
 
 	-- Change draw order based on 'timeline_style' to keep the heatmap visible
diff --git a/ass.lua b/ass.lua
index dc7b941..8abc9dd 100644
--- a/ass.lua
+++ b/ass.lua
@@ -270,9 +270,10 @@ end
 -- Draws a smooth curve through points using Catmull-Rom to Bezier conversion.
 ---@param points number[][] List of {x, y} coordinate pairs
 ---@param opts? {color?: string; border?: number; border_color?: string; opacity?: number|{primary?: number; border?: number, shadow?: number, main?: number}; clip?: string}
-function ass_mt:smooth_curve(points, opts)
+function ass_mt:smooth_curve(points, opts, cache)
 	if not points or #points < 2 then return end
 	opts = opts or {}
+    cache = cache or {}
 	local border_size = opts.border or 0
 	local tags = '\\pos(0,0)\\rDefault\\an7\\blur0'
 	-- border
@@ -291,21 +292,29 @@ function ass_mt:smooth_curve(points, opts)
 	self:draw_start()
 
 	self:move_to(points[1][1], points[1][2])
-	local function catmullrom_to_bezier(p0, p1, p2, p3)
-		local cp1x = p1[1] + (p2[1] - p0[1]) / 6
-		local cp1y = p1[2] + (p2[2] - p0[2]) / 6
-		local cp2x = p2[1] - (p3[1] - p1[1]) / 6
-		local cp2y = p2[2] - (p3[2] - p1[2]) / 6
-		return {cp1x, cp1y}, {cp2x, cp2y}
+	if points ~= cache.last_points then
+		local curve = {}
+		local function catmullrom_to_bezier(p0, p1, p2, p3)
+			local cp1x = p1[1] + (p2[1] - p0[1]) / 6
+			local cp1y = p1[2] + (p2[2] - p0[2]) / 6
+			local cp2x = p2[1] - (p3[1] - p1[1]) / 6
+			local cp2y = p2[2] - (p3[2] - p1[2]) / 6
+			return {cp1x, cp1y}, {cp2x, cp2y}
+		end
+		local n = #points
+		for i = 1, n - 1 do
+			local p0 = points[math.max(i-1,1)]
+			local p1 = points[i]
+			local p2 = points[i+1]
+			local p3 = points[math.min(i+2,n)]
+			local cp1, cp2 = catmullrom_to_bezier(p0, p1, p2, p3)
+            curve[i] = {cp1 = cp1, cp2 = cp2, p = p2}
+		end
+		cache.segments = curve
+		cache.last_points = points
 	end
-	local n = #points
-	for i = 1, n - 1 do
-		local p0 = points[math.max(i-1,1)]
-		local p1 = points[i]
-		local p2 = points[i+1]
-		local p3 = points[math.min(i+2,n)]
-		local cp1, cp2 = catmullrom_to_bezier(p0, p1, p2, p3)
-		self:bezier_curve(cp1[1], cp1[2], cp2[1], cp2[2], p2[1], p2[2])
+	for _, c in ipairs(cache.segments) do
+		self:bezier_curve(c.cp1[1], c.cp1[2], c.cp2[1], c.cp2[2], c.p[1], c.p[2])
 	end
 	self:draw_stop()
 end
```

This seems to shave a bit of that additional CPU usage while timeline is visible. Should i push this ?

Last time i was trying to make this return the whole ASS formated string and cache it with the rest of `self._heatmap`, and all i managed was to almost get an aneurysm, because all it did was render the data as plaintext... or worse, required more loops to directly compare all the data inside the table, which defeated the purpose a bit.

tomasklaen:
Separate out the function that converts points to bezier points, and make `ass:smooth_curve()` just accept these. Then just cache the bezier points directly with:

```lua
self._heatmap.cached_midpoints = points_to_bezier(points)
```

Sneakpeakcss:
I wanted to avoid gutting the `smooth_curve()` like that in case something else would use it in the future. So with that in mind, should i seperate it inside of `ass.lua` as:

```lua
function ass_mt:points_to_bezier(points)
    if not points or #points < 2 then return {} end
    local function catmullrom_to_bezier(p0, p1, p2, p3)
        local cp1x = p1[1] + (p2[1] - p0[1]) / 6
        local cp1y = p1[2] + (p2[2] - p0[2]) / 6
        local cp2x = p2[1] - (p3[1] - p1[1]) / 6
        local cp2y = p2[2] - (p3[2] - p1[2]) / 6
        return {cp1x, cp1y}, {cp2x, cp2y}
    end
    local curve = {}
    for i = 1, #points - 1 do
        local p0 = points[math.max(i - 1, 1)]
        local p1 = points[i]
        local p2 = points[i + 1]
        local p3 = points[math.min(i + 2, #points)]
        local cp1, cp2 = catmullrom_to_bezier(p0, p1, p2, p3)
        curve[i] = {cp1 = cp1, cp2 = cp2, p = p2}
    end
    curve.start_point = points[1]
    return curve
end
```
and cache it with the:

```lua
self._heatmap.cached_midpoints = ass:points_to_bezier(points)
```
Or did you mean to move it to `utils.lua` ? 

Moving this staright to `timeline.lua` would turn `smooth_curve()` into a tiny wrapper around `move_to()` and `bezier_curve()` that's hardly reusable by anything else.

Sneakpeakcss:
Since we're moving the the control point calculations outside of `smooth_curve()` we need to somehow pass the starting points to it. Previously it was just `self:move_to(points[1][1], points[1][2])`

Sneakpeakcss:
Okay, i think i've got it now. At least i hope so, because this whole thing has been a learning *curve* already :^)

Both midpoints and curve are now "flattened":

```
[0,1080,1.600000,1078.624466,4.800000,1072.486438,9.600000,1071.746798,14.400000, ...
```

> ```lua
> self:move_to(points[1], points[2])
> local max = math.floor((#points - 2) / 6) * 6 + 2 -- ensure we don't overflow when invalid number of points
> for i = 3, max, 6 do
> 	self:bezier_curve(points[i], points[i+1], points[i+2], points[i+3], points[i+4], points[i+5])
> end
> ```
> 
> Not tested, I'm sure the snippet above has some off by one errors :)

Well, it did work without issues, so there's that.

tomasklaen:
Ok, looking good, but I have one last optimization request :) I should've spent a bit more time reviewing this before so I'd think of it sooner. Sorry for wasting your time like this 😶.

Instead of scaling, positioning, and converting points to bezier on every frame with some timeline dimension change, we just convert the points into bezier once inside `file-loaded` event but normalize them to `0-1` coordinate space (0,0: top-left, 1,1: bottom-right), so `self.heatmap` will be an already beziered table of flattened points that range from 0 to 1.

The `ass_mt:smooth_curve(points, opts)` will then change to `ass_mt:smooth_curve(ax, ay, bx, by, points, opts)` and it'll draw the points within the requested rectangle, so we won't have to convert or cache any tables between frames.

Sneakpeakcss:
> Ok, looking good, but I have one last optimization request :) I should've spent a bit more time reviewing this before so I'd think of it sooner. Sorry for wasting your time like this 😶.

My main motivation for this was trying to understand how `uosc` positions its elements without using `/pos` and avoiding this terrible shifting and misalignment whenever the window is resized. That thing defeated me previously when i was making my *bars-module-utility-thingy:*

<details>
<summary>demo</summary>

https://github.com/user-attachments/assets/eda7af6b-5a48-4250-9a82-7974da4ecf46

<sub>Now i can tell myself that i will refactor this thing... one day, surely :^)</sub>

</details>

So anything beyond that was a nice bonus 👍

> Instead of scaling, positioning, and converting points to bezier on every frame with some timeline dimension change, we just convert the points into bezier once inside `file-loaded` event but normalize them to `0-1` coordinate space (0,0: top-left, 1,1: bottom-right), so `self.heatmap` will be an already beziered table of flattened points that range from 0 to 1.
> 
> The `ass_mt:smooth_curve(points, opts)` will then change to `ass_mt:smooth_curve(ax, ay, bx, by, points, opts)` and it'll draw the points within the requested rectangle, so we won't have to convert or cache any tables between frames.

Done. The whole thing looks a lot tidier now.

tomasklaen:
The clipping in `draw_heatmap()` is not necessary. Just position it where it needs to be.

And this little tweak:
```lua
local function scale(x, y)
	local sx = ax + x * (bx - ax)
	local sy = ay + y * (by - ay)
	return sx, sy
end
```
->
```lua
local width, height = bx - ax, by - ay
local function scale(x, y)
	return ax + x * width, ay + y * height
end
```

and we're good :)

PR #1166, opened 2025-09-03
fix: incorrect anchoring on final heatmap point

Sneakpeakcss:
Somehow i managed to miss it.

___

On top of that, this exposed an issue:

overshoot

While we normalize the points they can still overshoot while creating control points for a smooth curve:

$cp_1^y = 0.900000 + \frac{0.900000 - 0.012372}{6} \approx 1.047938$  
$cp_2^y = 0.900000 - \frac{0.897786 - 0.900000}{6} \approx 0.900369$ 

$cp_1^y = 0.070464 + \frac{0.036057 - 0.588875}{6} \approx -0.021672$  
$cp_2^y = 0.036057 - \frac{0.551626 - 0.070464}{6} \approx -0.044137$  

I've checked youtube just to make sure if i didn't mess something up, and we basically have a match with what's inside of youtube's SVG:

<sub>uosc:</sub>
uosc
<sub>youtube site display:</sub>
youtube
<sub>youtube full SVG:</sub>
youtubefullsvg

From what i can tell, youtube draws the heatmap by filling rectangles with color and using an SVG `clip-path` to mask the visible area, which slices the peaks that extend beyond their 40px tall clipping region, so this is expected behaviour.

Seems like we're going to need that `\clip` to keep everything tidy without distorting the heatmap.

Should i open a seperate PR for this, or should i simply add it to this commit to avoid polluting the history too much ?

<sub> Also fixed that stray space indentation while i'm at it :^) </sub>

### DIFF
--- a/src/uosc/lib/utils.lua
+++ b/src/uosc/lib/utils.lua
@@ -136,8 +136,8 @@ end
 function get_point_to_rectangle_proximity(point, rect)
 	local dx = math.max(rect.ax - point.x, point.x - rect.bx)
 	local dy = math.max(rect.ay - point.y, point.y - rect.by)
-    local distance = math.sqrt(math.max(0, dx)^2 + math.max(0, dy)^2)
-    return distance + math.min(0, math.max(dx, dy))
+	local distance = math.sqrt(math.max(0, dx)^2 + math.max(0, dy)^2)
+	return distance + math.min(0, math.max(dx, dy))
 end
 
 ---@param point_a Point
@@ -999,7 +999,8 @@ function load_youtube_heatmap()
 			norm[#norm + 1], norm[#norm + 2] = norm_x, norm_y
 		end
 		-- Add final anchor
-		norm[#norm + 1], norm[#norm + 2] = 1, max_norm_y
+		local last_y = math.min(max_norm_y, 1 - (data[#data].value / max_val))
+		norm[#norm + 1], norm[#norm + 2] = 1, last_y
 		norm[#norm + 1], norm[#norm + 2] = 1, 1
 		return points_to_bezier(norm)
 	end


Sneakpeakcss:
> Add to this one and name it something like "heatmap tweaks and fixes".

This should be fine, i hope.

> And add clipping only if timeline is overlayed. When it's above it's fine if it overflows on top.

I don't know why youtube cuts it like that when they draw it floating above their timeline, but it sure looks ugly, so keeping the peak in `above` mode is certainly the plan.

However, the `above` placement still requires clipping, because overshooting values over `1` create this weird shadow line over the border:

fullopaqueShadow

The example peak in the PR’s opening post already extended 5px above the 40px rectangle that we use for `above`, so i've just extended the upper bound by 10px to be safe in case another edge case draws an even higher peak.

And we end up with:

<sub>above</sub>
above

<sub>overlay</sub>
overlay

Sneakpeakcss:
> > However, the above placement still requires clipping, because overshooting values over 1 create this weird shadow line over the border:
> 
> Can you elaborate here? I don't think I'm getting/seeing what you mean here.

When the control point overshoots above `1`, the bottom line of the heatmap that should smoothly close with a straight line from `{1, 1}` to `{0, 1}` suddenly has an additional pixel under it. Even with heatmap opacity set to `1`, it still displays as a shadow that blends with whatever is under it, in this case uosc's 1px timeline border.

I'm not exactly sure how to properly show it here, since even with high-contrast colours it's faint, but it's certainly there:

1px `timeline_border` with blue progress line and green background:

image

2px `timeline_border`:

image

and with clip it disappears:

image

 

Two more examples with 2px `timeline_border`:

<sub>without clip:</sub>

image

<sub>with clip:</sub>

image

_It takes a lunatic with a 500% magnifier, who checks the alignment of everything because he noticed it and spent way too much time wondering if he somehow aligned the whole thing wrongly, so i'm not surprised by the confusion :^)_

PR #1194, opened 2025-12-09
fix: unset cache_underrun when demuxer-cache-state becomes unavailable

layercak3:
If --idle and --force-window is used and playback of a buffering file is aborted, the buffering indicator will remain enabled until a new file is loaded because of this not being unset.


tomasklaen:
Thx!

PR #1207, opened 2026-02-07
Add ability to customize window border color

SelfishPig:
Adds the ability to separately customize the window border color.


tomasklaen:
Rename it to `window_border`.

PR #1226, opened 2026-05-22
fix: inconsistent sorting logic in non-windows platform, fixed #1225

### LINKED ISSUE #1225: Inconsistent sorting logic
There is a discrepancy in the file/string sorting logic between Windows and non-Windows environments within uosc/lib/utils.

## Describe the bug

On Windows: The native StrCmpLogicalW function treats numbers within strings as numerical values. It correctly recognizes decimals/versions in a natural sorting order: 1.1 < 1.2 < 1.11.

On Non-Windows (Linux/macOS): sort_lua handles the string differently. It treats the parts after the decimal point strictly as individual numbers or regular characters, which causes a "pure numeric natural sort" fallback. As a result, it evaluates 1.11 as smaller than 1.2 because 1 comes before 2 at that character position. This leads to an inconsistent order: 1.1 < 1.11 < 1.2.

## To reproduce

Sort a list of strings containing decimal-like numbers or version numbers (e.g., ["1.1", "1.11", "1.2"]):

- Expected behavior (Natural/Windows Sort): 1.1, 1.2, 1.11
- Actual behavior on non-Windows: 1.1, 1.11, 1.2

## Why this happeds

StrCmpLogicalW on Windows splits strings by digits but considers punctuation boundaries differently, or specifically handles dot-separated numbers natively for file systems. The fallback sort_lua implementation doesn't perfectly match this behavior, leading to unexpected UI ordering differences depending on the user's OS.

SuniRein:
#1225

### DIFF
--- a/src/uosc/lib/utils.lua
+++ b/src/uosc/lib/utils.lua
@@ -54,15 +54,14 @@ do
 
 	-- alphanum sorting for humans in Lua
 	-- http://notebook.kulchenko.com/algorithms/alphanumeric-natural-sorting-for-humans-in-lua
-	local function padnum(n, d)
-		return #d > 0 and ('%03d%s%.12f'):format(#n, n, tonumber(d) / (10 ^ #d))
-			or ('%03d%s'):format(#n, n)
+	local function padnum(n)
+		return ("%03d%s"):format(#n, n)
 	end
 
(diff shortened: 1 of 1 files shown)


tomasklaen:
Thanks!

PR #1232, opened 2026-07-19
added a no ai policy

ghost:


po5:
I'm subscribed to uosc development and get email notifications, I haven't seen any evidence of problematic AI PRs and I'm sure the same is true of all projects you're submitting drive-by low effort PRs to, with markdown that isn't even formatted correctly and won't render. I am preemptively blocking you from ever interacting from any of my own repos.

You are causing many more problems rather than adding value:

image

PR #1233, opened 2026-07-22
fix: cursor_up could activate selected item somewhere else

### LINKED ISSUE #1001: Clicking non-selectable item clicks last hovered item
I don't think it should do this.

https://github.com/user-attachments/assets/db63a6c5-eb35-4174-a9eb-a2f6a895c6f0

ewt45:
fixes: #1001 

As mentioned in the issue, if mouse is currently hovering on an unselectable item, clicking will activate a selectable item somewhere else.

this pr adds a new property `Menu.mouse_hovered_index`. it's like `selected_index` but is updated even when the item is not selectable.  
on `handle_cursor_up`, `mouse_hoverd_index == selected_index` is checked before activating the selected item.


tomasklaen:
Thanks!

PR #1235, opened 2026-07-30
feat: center speed indicator by improving layout algorithm

ukjent7:
Before：
屏幕截图 2026-07-30 085056

now:
屏幕截图 2026-07-30 084758

屏幕截图 2026-07-30 085423

No visual regressions observed when shrinking the window.

屏幕截图 2026-07-30 090106


tomasklaen:
Haven't examined every line but at a glance it looks fine and seems to work with everything I've thrown at it. Thanks!

PR #1237, opened 2026-08-02
MTBN_right to toggle time precision

escapezn:
It let MTBN_right to toggle time precision like stock osc.lua


tomasklaen:
This was too opinionated and non-configurable, but the feature commit above adds a new `timeline_mbtn_right` option that allows the exact same thing.

PR #1251, opened 2026-08-30
Adjust the conditions for changes in the 'has_many_editions' attribute

088uiop:
Fix: Ensure `has_many_edition` state updates correctly when switching to files without editions

Previously, the `editions` property observer had a conditional check:
`if editions then ...`. When switching from a file with editions to one without,
the `editions` property becomes `nil`. Because of the `if` condition, the
`set_state('has_many_edition', editions > 1)` line never executed, leaving the
state stuck at `true`.

After removing the condition, correctly returns `false` when
`editions` is `nil`, properly hiding the editions button.


tomasklaen:
Thanks!

